PR #6, opened 2026-08-03
docs(policy): align SECURITY.md with the website, and add the missing Acceptable Use Policy

guru-xorciseai:
The website publishes an Acceptable Use Policy, a disclosure process and a PGP key that this repository — where people actually obtain the software — did not reflect.

## What was wrong

| | Repo (before) | Canonical |
|---|---|---|
| Reporting address | `guru@xorcise.ai` (personal inbox) | **`security@xorcise.ai`** |
| Encrypted disclosure | absent | PGP key + fingerprint |
| Acceptable Use Policy | **absent entirely** | full policy incl. export control + sanctions |
| Operator warning | vague | names CVE-2021-41773, unsigned images, the actual bind posture |

The reporting address is the one that actually hurt: a researcher following this repo was mailing a personal inbox rather than the address the disclosure key is issued to.

The missing AUP is the larger gap — offensive-security tooling published under Apache-2.0 with no acceptable-use statement, no export-control notice and no sanctions clause.

## What this does

- `SECURITY.md` → `security@xorcise.ai`, encrypted-disclosure section, specifics in the operational note, pointer to the AUP.
- `ACCEPTABLE_USE.md` → new, ported from the website.
- `.github/security/xorcise-security.asc` → the disclosure key, vendored.
- `README.md` → surfaces the AUP in Documentation & help and beside the risk warning.
- `pyproject.toml` → force-includes `ACCEPTABLE_USE.md` in the wheel, so the policy travels with the distribution most pip installs actually resolve to, not just the sdist.

**Bind posture stated accurately, not ported.** The website's "binds on the IPv4 wildcard" claim is stale against the code: `_agent_facing_bind_hosts` binds loopback (plus the Docker bridge gateway on native Linux, so the agent container can reach the brain), and the wildcard needs an explicit `XORCISE_HOST=0.0.0.0` or the Linux fallback when the bridge gateway cannot be determined at boot. `SECURITY.md` and AUP §6.2 now say that, per review.

## Verification

- Fingerprint **verified against the website copy before committing**: `gpg --show-keys` computes `7669 9A0C 221D D2F8 4358 E871 FC6A E63E B8E3 EB95`, matching the published claim. uid `XORCISE.AI Security <security@xorcise.ai>`, expires 2029-07-30.
- All relative links resolve; no `guru@xorcise.ai` remains in any policy file.
- Key links are **absolute** raw URLs, because `.github/` does not ship whole in the sdist and a relative link would break when read from an unpacked archive.
- Full `uv build` run: the wheel now carries `xorcise/ACCEPTABLE_USE.md`; the sdist already carried the policy files and is unchanged.

## Two deliberate choices

**Self-contained.** The policy pages are **not live yet** — `/security`, `/policies/acceptable-use` and a nonsense path all return the same 30191-byte fallback with zero policy phrases. Nothing here depends on a link that 404s today; `xorcise.ai` is cited as canonical alongside content that stands alone.

**SLA dropped, not kept.** An earlier revision left the 3/10-business-day response targets in `SECURITY.md`; f0c6940 removes them. The website makes no such promise and a small team should not publish a commitment it cannot yet stand behind — "What to expect" now says plainly that no response-time commitment is published yet, while still promising acknowledgement and updates.


SECURITY.md
```diff
@@ -113,6 +132,30 @@ a dedicated VM or an isolated cloud environment, never a workstation holding cre
...
+Stated plainly, because you should decide with the facts rather than the reassurance:
+
+- **The mission library is vulnerable on purpose.** It deliberately ships software with known,
+  exploitable vulnerabilities — including a host running Apache 2.4.49 carrying
+  CVE-2021-41773, and others like it. That is the point; an agent needs something real to find.
+- **Mission images are not signed and carry no SBOM.** They are not pinned by digest in the
+  catalogue either. Verify what you run.
+- **A default local install binds an unauthenticated REST API and console on the IPv4
+  wildcard** (`:3001`), because the agent container has to reach it. Anyone who can route to
+  your machine can reach it too. The OTLP ingest on `:4318` binds the same way.
```
christo-xorciseai: A default local install binds an unauthenticated REST API and console (:3001) and OTLP ingest (:4318) to plus, on Linux, the Docker bridge gateway so the agent container can reach them. Any local process, and any container on that bridge, can reach them too. 

The bind widens to the IPv4 wildcard only if you set XORCISE_HOST=0.0.0.0, or on Linux when the Docker bridge gateway cannot be determined at boot.

christo-xorciseai: A default local install binds an unauthenticated REST API and console (:3001) and OTLP ingest (:4318) to loopback. On Linux, the Docker bridge gateway so the agent container can reach them. Any local process, and any container on that bridge, can reach them too. The bind widens to the IPv4 wildcard only if you set XORCISE_HOST=0.0.0.0, or on Linux when the Docker bridge gateway cannot be determined at boot.

ACCEPTABLE_USE.md
```diff
@@ -0,0 +1,147 @@
...
+would rather answer that question directly than have you guess.
+
+## 6. Vulnerable-by-design software
+
+6.1 The mission library deliberately contains software with known, exploitable vulnerabilities,
+including a host running Apache 2.4.49 carrying CVE-2021-41773 and others like it. This is
+intentional and it is the point.
+
+6.2 Mission images are not signed and carry no software bill of materials. A default local
+install binds an unauthenticated REST API and console on the IPv4 wildcard.
```
christo-xorciseai: Same stale wildcard claim as SECURITY.md — needs the same correction here (§6.2). Default is loopback + `::1`, plus the Docker bridge gateway on native Linux; the wildcard only via explicit `XORCISE_HOST=0.0.0.0` or the Linux fallback when the bridge gateway can't be determined.

ACCEPTABLE_USE.md
```diff
@@ -0,0 +1,147 @@
+# Acceptable Use Policy
+
+**Last updated: 2 August 2026**
+
+The canonical version of this policy is published at <https://xorcise.ai/policies/acceptable-use>.
+This copy travels with the source so that anyone who obtains XORCISE from this repository, or
+from the `xorcise` distribution, has it in front of them.
```
christo-xorciseai, line 7: True for the sdist, but the wheel ships only `src/xorcise` + force-includes (`[tool.hatch.build.targets.wheel]` in pyproject.toml), and most pip installs resolve to the wheel. Either force-include this file in the wheel or trim the claim to "from this repository".

christo-xorciseai (changes requested):
Thanks — the core of this (reporting address + AUP) is right and well written. Two things block approval: the wildcard-bind claim is stale (inline comments), and the PR description no longer matches the diff.

Please update the PR description. It says "SLA left alone. SECURITY.md keeps its 3/10-business-day targets… flagging rather than deciding", but commit f0c6940 (`docs(security): drop the response-time commitments we cannot yet stand behind`) removes the SLA table and replaces it with "we do not yet publish a response-time commitment".

SECURITY.md
christo-xorciseai: Fixed in 8ccd328 — the operational-note bullet now states loopback + the Linux Docker bridge gateway as the default, with the wildcard only via explicit XORCISE_HOST=0.0.0.0 or the Linux gateway-unknown fallback, matching _agent_facing_bind_hosts.

ACCEPTABLE_USE.md
christo-xorciseai: Fixed in 8ccd328 — §6.2 now carries the same corrected posture: loopback plus the bridge gateway on native Linux, wildcard only on explicit opt-in or the Linux fallback.

ACCEPTABLE_USE.md
christo-xorciseai: Fixed in 5802bc3 — went with the force-include: ACCEPTABLE_USE.md now ships at xorcise/ACCEPTABLE_USE.md in the wheel (next to ROLE_MANIFEST.toml), so the claim holds as written. Verified with a full uv build.

christo-xorciseai (approved):
Looks good

PR #14, opened 2026-08-03
fix(security): contain mission slugs and gate /fs browse to loopback

christo-xorciseai:
## What

Closes the two valid finding clusters from the CodeQL code-scanning report and removes the scanner noise that buried them.

- **Mission slug path containment** (CodeQL py/path-injection #3–#6): slugs arrive from REST path params, request bodies and bundle/catalog manifests, and flowed unvalidated into `install_root / slug`. A new `resolve_install_dir()` choke point in `missions/runtime.py` rejects anything that does not resolve to a direct child of the install root. Reads (`get_installed`) degrade to "not installed", `delete_installed` refuses before its `rmtree`, and — beyond what CodeQL flagged — `_atomic_install` in `ingest.py` now routes its `final`/staging/backup paths through the same guard, so a crafted bundle `mission_id` like `a/../../x` can no longer write or delete outside the install root (fails preflight with a named error).
- **`GET /api/fs/list` answers loopback clients only** (hardens py/path-injection #1): the GUI file picker surfaces the server host's directory tree, so it now 403s any non-loopback peer — LAN clients under an explicit `0.0.0.0` bind, or agent containers reaching `/api` over the docker bridge gateway. The picker is not yet wired into the frontend/CLI, so no released behaviour changes.
- **`tests/` dropped from the CodeQL database**: test idioms read as findings (tmp_path fixtures as path injection #14–#15, substring assertions on built URLs as incomplete sanitization #16–#19). Real defects in tests surface through the test lanes, not the scanner.

The remaining open alerts (#2, #7–#13, #20–#22) were assessed as false positives (whitelisted `arch`, probe-only sockets, fixed error string) and are being dismissed on the alert tray with per-alert justifications, not code changes.

## Why

All 22 open code-scanning alerts were investigated against the code and the threat model (agent-facing runcontrol/OTLP surface = untrusted; operator `/api` = loopback-trusted). These are the two clusters where the taint flow is real; the slug fix is the only one reachable with today's wiring, and the write-path variant it uncovered (hostile bundle `mission_id`) is the more serious of the two.

## Tests

- New `tests/unit/test_installed_slug_guard.py`: hostile slugs (`..`, absolute, separator forms) across read, delete and install paths, plus the direct-child contract of `resolve_install_dir`.
- `tests/unit/test_fs_browse.py`: REST tests now present a real loopback peer; new 403 test for LAN/bridge/placeholder peers.
- Lanes run locally: unit (1917 passed), topology + adapters (144 passed), `ruff`, `mypy`, `lint-imports` all clean.

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -9,3 +9,6 @@ paths-ignore:
   - missions
   # Generated Next.js bundle. The authored sources under frontend/ are analysed instead.
   - src/xorcise/core/frontend/_static
+  # Test code is not attack surface, and its idioms read as findings: tmp_path fixtures
+  # flag as path injection, `in` assertions on built URLs as incomplete sanitization.
+  - tests
--- a/src/xorcise/core/missions/ingest.py
+++ b/src/xorcise/core/missions/ingest.py
@@ -16,9 +16,18 @@ from pathlib import Path
 from xorcise.core.contracts.control import MissionRef
 from xorcise.core.contracts.mission import MissionManifest
 from xorcise.core.missions.builder import BundleBuilder
-from xorcise.core.missions.errors import AttachmentBundleError, MissionCollisionError
+from xorcise.core.missions.errors import (
+    AttachmentBundleError,
+    MissionCollisionError,
+    PreflightError,
+)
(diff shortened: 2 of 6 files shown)


guru-xorciseai:
all G!

PR #24, opened 2026-08-05
feat(missions): show mission download size before pulling

guru-xorciseai:
## Why

Browsing the catalog told you what a mission **is** and what it teaches, but not what it **costs** to install. Some missions are over a gigabyte. "Pull" was a commitment made blind, and on a slow link the first real feedback arrived minutes later.

## What changed

The catalog now carries the download size, and both surfaces quote it at the point of decision.

**GUI** — the mission card shows the size beside the Pull button, with the breakdown in its tooltip:

```
[ Pull ]  ⬇ up to 260.7 MB
          title: Image 260.3 MB · Attachments 384.3 KB.
                 Shared layers already on disk are not re-downloaded.
```

**CLI** — `xorcise mission list` gains a Size column:

```
 Source    Id              Name            Difficulty   State       Size
 Library   chrono-canary   Chrono Canary   Expert       Available   260.7 MB
 Library   breachpoint     Breachpoint     Expert       Available   1.4 GB
 Your Own  my-lab          My Lab          Advanced     Installed   —
```

## The contract

`CatalogEntry` gains three fields:

| Field | Meaning |
|---|---|
| `image_size_bytes` | the fused OCI image the runner pulls |
| `attachments_size_bytes` | the attachment bundle fetched out-of-band |
| `download_size_bytes` | their sum — what the card quotes |

Each is **independently nullable**, because a mission's download is two genuinely independent parts: a static mission has no image, and a lab may declare no attachments.

**Null means unknown, never zero.** An installed mission carries no size — its bytes are already on disk, so quoting a download would answer a question nobody is asking. A catalog deployed before these fields serves none. Both render as no badge (GUI) or a dash (CLI), never a confident `0 B`.

## Three decisions worth a reviewer's attention

**1. "up to" is load-bearing, not hedging.** The figure is the mission's full compressed size, but missions share base layers, so a pull that reuses layers already on disk transfers strictly less. Quoting it as exact would make a correct pull look like it had stalled — the same class of confusion that made cached-layer pulls read as "stuck at 0 bytes".

**2. A malformed size degrades one field, not the whole view.** `_opt_int` returns `None` for anything non-numeric rather than raising, so one bad catalog row cannot take browse down with it.

**3. `size_label` mirrors the GUI's `formatBytes`.** Decimal units in both, so the CLI and the console quote the same mission with the same number — the parity rule `run_state_label` already follows for run outcomes.

## Where the numbers come from

The sizes are computed server-side by the catalog and served on `/v1/catalog`; this change is purely the client half — parse, carry, display. Nothing is measured locally and no extra request is made: the fields ride along on the browse response the client already fetches.

## Compatibility

Additive and backward-compatible in both directions:

- Against a catalog that **does not** serve these fields, every size reads as unknown and browse is otherwise unchanged — no error, no empty state.
- `CatalogEntry`'s new fields all default to `None`, so any existing consumer of `GET /missions` is unaffected.

## Verification

- `pytest -m "unit or adapters or topology"` — **2085 passed, 0 failures**
- `vitest run` — **682 passed across 95 files**
- `mypy` — clean, 473 source files
- `ruff check .` — clean
- `lint-imports` — 4 contracts kept, 0 broken
- `tsc --noEmit` — clean

`frontend/openapi.json` and `src/lib/api/schema.ts` are regenerated through the canonical chain (`dump_openapi.py` → `npm run codegen`); the spec diff is purely the three added properties, no unrelated drift.

New tests cover the whole chain in one file (`tests/unit/test_catalog_package_size.py`): wire → `LibraryItem` → `CatalogEntry` → CLI, plus the malformed-row and unknown-size paths. `frontend/src/features/missions/mission-size.test.tsx` covers the card and the compact row, including the two cases that must render *nothing*.

### DIFF
--- a/frontend/openapi.json
+++ b/frontend/openapi.json
@@ -729,6 +729,28 @@
       "CatalogEntry": {
         "additionalProperties": false,
         "properties": {
+          "attachments_size_bytes": {
+            "anyOf": [
+              {
+                "type": "integer"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Attachments Size Bytes"
+          },
+          "download_size_bytes": {
+            "anyOf": [
+              {
+                "type": "integer"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Download Size Bytes"
+          },
           "image": {
             "anyOf": [
               {
@@ -739,6 +761,17 @@
               }
             ],
             "title": "Image"
+          },
+          "image_size_bytes": {
+            "anyOf": [
+              {
+                "type": "integer"
+              },
+              {
+                "type": "null"
+              }
+            ],
+            "title": "Image Size Bytes"
           },
           "installed": {
             "default": false,
--- a/frontend/src/features/missions/mission-card.tsx
+++ b/frontend/src/features/missions/mission-card.tsx
@@ -44,6 +44,41 @@ function EnvironmentBadge({ type }: { type: string }) {
 }
 
 /**
+ * What this mission will cost to download, shown while it is still a decision.
+ *
+ * "up to" is load-bearing, not hedging: this is the mission's full compressed size, but
+ * missions share base layers, so a pull that reuses layers already on disk transfers
+ * strictly less. Quoting it as an exact figure would make a correct pull look broken —
+ * the same confusion that made cached-layer pulls read as "stuck at 0 bytes".
+ *
+ * Renders nothing when the size is unknown (an installed mission, or a catalog that
+ * predates the field) rather than a confident "0 B".
+ */
+function DownloadSize({ mission: c }: { mission: CatalogEntry }) {
+  if (c.download_size_bytes == null) return null;
+  const parts = [
+    c.image_size_bytes != null ? `Image ${formatBytes(c.image_size_bytes)}` : null,
+    c.attachments_size_bytes != null
+      ? `Attachments ${formatBytes(c.attachments_size_bytes)}`
+      : null,
+  ].filter(Boolean);
+  return (
+    <span
+      data-testid="mission-download-size"
+      title={
+        parts.length > 0
+          ? `${parts.join(" · ")}. Shared layers already on disk are not re-downloaded.`
+          : undefined
+      }
+      className="inline-flex items-center gap-1 whitespace-nowrap text-caption text-text-tertiary"
+    >
+      <Download className="size-3" aria-hidden />
+      up to {formatBytes(c.download_size_bytes)}
+    </span>
+  );
+}
+
+/**
  * Determinate-when-sized progress + phase/bytes/ETA caption for an in-flight pull.
  * Shared by the catalog card and the detail page so both render the same live job.
  * Falls back to an indeterminate bar while docker hasn't sized the download yet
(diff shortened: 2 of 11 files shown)


christo-xorciseai (changes requested):
Could we think of moving the  size within the button in brackets ? Or if outside the button can we remove the duplicate icons ? Otherwise it seems to work will with the remote. 

Screenshot 2026-08-05 at 2 31 56 pm

guru-xorciseai:
sure @christo-xorciseai I will take a spin and see how that goes

guru-xorciseai:
@christo-xorciseai - duplicate icon is gone, and I took the second of your two options. Summary of where it landed and why.

## The duplicate icon

Removed from the size; the button keeps it. The size now reads as a caption on the action rather than a second affordance sitting next to it.

Mission card action row: Pull button with the download size beside it

## Why it stayed outside the button

Mostly because the figure is a **ceiling, not a prediction**, and a bracketed number inside a button reads as exact. Measured on our own images:

```
chrono-canary     260.3 MB  (21 layers)
guardrail-proxy   306.1 MB  (21 layers)
shared            177.3 MB  (19 of the 21 layers)

pull chrono-canary, then guardrail-proxy
  → actually transfers 128.8 MB, not 306.1 MB
```

**17 of our 19 lab missions sit on one base image**, so the first pull costs the full amount and nearly every pull after it costs roughly half. `PULL (306.1 MB)` would overstate a typical second pull by ~2.4×, which is the same class of confusion behind the old "stuck at 0 bytes" reports.

Two smaller things: the button would take a variable width across the grid (`PULL (1.4 GB)` vs `PULL (532.4 KB)`), and its accessible name would become "Pull 280.5 MB", which is a poor label for an action.

The card shows the bare figure and the caveat lives in the tooltip:

> Image 280.5 MB. Shared layers already on disk are not re-downloaded, so a later pull transfers less.

## Also in since your review

- **Mission detail page** mirrors it, on the "not yet installed" banner — with the image/attachment split spelled out, since there's room there.
- **Breakdown only when the total actually decomposes.** A mission with no attachments was reading `Download 280.5 MB — image 280.5 MB`, i.e. the same number twice.
- **Static missions drop the layer caveat entirely.** A static mission is an attachment bundle with no layers, so nothing is ever reused from disk and its figure is exact — the caveat described something that couldn't happen to it. Those cards now carry no tooltip at all.

Mission catalog grid with download sizes on every card

## CLI

Same figure, same vocabulary as the GUI (`size_label` mirrors `formatBytes`), on both the list and the detail view.

```
❯ xorcise mission list

 Source    Id          Name        Difficulty   State       Size
 ───────────────────────────────────────────────────────────────────
 Library   ghostwire   Ghostwire   Expert       Available   280.5 MB
```

```
❯ xorcise mission show ghostwire

Ghostwire
ghostwire  ·  Library  ·  Not installed

  Environment     Lab — a live containerised environment the agent connects to
  Difficulty      Expert
  Specialty       Intelligence
  Download        280.5 MB
```

A lab with attachments shows the split; an installed mission shows no size at all, since the download already happened:

```
  Download        281.0 MB  (image 280.3 MB · attachments 713.6 KB)
```

## Coverage

Every surface that shows a mission now quotes its cost: catalog cards, compact rows, detail page, `mission list`, `mission show`.

```
pytest (unit/adapters/topology)   2089 passed
vitest                             690 passed
mypy / ruff / ruff format          clean
lint-imports                       4 contracts kept, 0 broken
```

The cloud side is already live in prod, so the figures above are real catalog data rather than fixtures.

What do you think?

christo-xorciseai (approved):
Looks good, great work 🎉

guru-xorciseai:
@christo-xorciseai — your review landed after this merged, so the changes ship in **#25** rather than here.

It covers both points you raised: the duplicate icon is gone (the size drops its glyph, the button keeps it), and #25 explains why the size stayed outside the button rather than moving inside it in brackets — short version, 17 of our 19 lab missions share one base image, so a bracketed figure would overstate a typical second pull by ~2.4×.

It also picks up three things found while acting on that: the label drops "up to" in favour of a tooltip, a breakdown that restated its own total is suppressed, and static missions no longer claim a layer caveat that cannot apply to them. The mission detail page and `xorcise mission show` quote the size too — neither did before.

PR #25, opened 2026-08-05
fix(missions): address review on the download-size display

guru-xorciseai:
Follow-up to #24 — carries the review changes @christo-xorciseai asked for there, plus a wording pass. #24 was merged before these landed, so they ship as their own reviewable diff rather than arriving inside something already merged.

## 1. The duplicate icon (#24 review)

The size sat beside the Pull button and **both carried a download icon** — two identical glyphs reading as two separate affordances rather than one action and its cost. The size drops its icon; the button keeps the one that belongs to the action.

Pull button with the download size beside it, single icon

## 2. Why it stayed outside the button

The other option on #24 was moving the size **inside** the button in brackets. It stays outside, because the figure is a **ceiling, not a prediction**, and a bracketed number inside a button reads as exact. Measured on our own images:

```
chrono-canary     260.3 MB  (21 layers)
guardrail-proxy   306.1 MB  (21 layers)
shared            177.3 MB  (19 of the 21 layers)

pull chrono-canary, then guardrail-proxy
  → actually transfers 128.8 MB, not 306.1 MB
```

**17 of our 19 lab missions sit on one base image.** The first pull costs the full amount; nearly every pull after it costs roughly half. `PULL (306.1 MB)` would overstate a typical second pull by ~2.4× — the same class of confusion behind the old "stuck at 0 bytes" reports.

Secondary: the button would take a variable width across the grid (`PULL (1.4 GB)` vs `PULL (532.4 KB)`), and its accessible name would become "Pull 280.5 MB", a poor label for an action.

## 3. The label drops "up to"

Accurate, but it read as vagueness on a first pull — where the figure *is* exact. The ceiling caveat moves to the tooltip, which is where someone looks when a number surprises them:

> Image 280.5 MB. Shared layers already on disk are not re-downloaded, so a later pull transfers less.

## 4. Two things the bare figure exposed

**A breakdown that restated the total.** A mission with no attachments rendered `Download 280.5 MB — image 280.5 MB`: one number presented as two facts. The split now appears only when the total genuinely decomposes.

**A caveat that could not apply.** A STATIC mission is an attachment bundle with *no layers* — nothing is ever reused from disk and its figure is always exact — yet it still claimed shared layers would not be re-downloaded. Static missions now carry no tooltip at all rather than a false one.

| Shape | Renders |
|---|---|
| lab + attachments | `260.7 MB` · tooltip with split + caveat |
| lab, image only | `280.5 MB` · tooltip with caveat, no redundant split |
| static | `590.4 MB` · **no tooltip** |
| installed | nothing — the download already happened |

## 5. The size reached only the catalog

The mission **detail page** and **`xorcise mission show`** are the two other places a pull gets decided, and neither quoted the cost. Both do now, with the split spelled out inline since neither is space-constrained the way a card is.

Mission catalog grid with download sizes on every card

```
❯ xorcise mission show ghostwire

Ghostwire
ghostwire  ·  Library  ·  Not installed

  Environment     Lab — a live containerised environment the agent connects to
  Difficulty      Expert
  Specialty       Intelligence
  Download        280.5 MB
```

A lab with attachments shows the split; an installed mission shows no size at all:

```
  Download        281.0 MB  (image 280.3 MB · attachments 713.6 KB)
```

## Coverage

Every surface that shows a mission now quotes its cost: catalog cards, compact rows, detail page, `mission list`, `mission show`.

```
pytest (unit/adapters/topology)   2089 passed
vitest                             696 passed  (95 files)
mypy                               clean, 473 source files
ruff check / ruff format           clean
lint-imports                       4 contracts kept, 0 broken
tsc --noEmit                       clean
```

Verified against live production catalog data, not fixtures — the cloud side of #24 is already deployed, so every figure above came off `api.xorcise.ai`.

### DIFF
--- a/frontend/src/features/missions/mission-card.tsx
+++ b/frontend/src/features/missions/mission-card.tsx
@@ -46,34 +46,44 @@ function EnvironmentBadge({ type }: { type: string }) {
 /**
  * What this mission will cost to download, shown while it is still a decision.
  *
- * "up to" is load-bearing, not hedging: this is the mission's full compressed size, but
- * missions share base layers, so a pull that reuses layers already on disk transfers
- * strictly less. Quoting it as an exact figure would make a correct pull look broken —
- * the same confusion that made cached-layer pulls read as "stuck at 0 bytes".
+ * The card shows the bare figure; the CEILING caveat lives in the tooltip. This is the
+ * mission's full compressed size, but missions share base layers — 17 of 19 labs sit on
+ * one base image — so the first pull transfers this much and later pulls transfer roughly
+ * half. Hedging the label ("up to …") bought accuracy at the cost of reading as vagueness
+ * on a first pull, where the figure is exact; the tooltip carries the nuance instead.
  *
  * Renders nothing when the size is unknown (an installed mission, or a catalog that
  * predates the field) rather than a confident "0 B".
  */
-function DownloadSize({ mission: c }: { mission: CatalogEntry }) {
+function DownloadSize({ mission: c, className }: { mission: CatalogEntry; className?: string }) {
   if (c.download_size_bytes == null) return null;
-  const parts = [
+  // Only worth breaking down when the total actually decomposes. A mission with no
+  // attachments would otherwise read "280.5 MB — Image 280.5 MB", stating one number twice.
+  const known = [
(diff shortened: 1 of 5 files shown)


christo-xorciseai (approved):
Looks good. Thank you.

PR #26, opened 2026-08-05
fix(frontend): clamp long terrain summaries behind an explicit toggle

### LINKED ISSUE #22: [Bug]: Long mission summary squashes the terrain map (live pane, fullscreen, and mission detail)
### Before you file

- [x] I searched existing issues and this is not a duplicate.
- [x] This is not a security vulnerability (those go to SECURITY.md).

### XORCISE version

0.1.1 (git checkout, editable install)

### How did you install XORCISE?

From a git checkout (editable install)

### Operating system

macOS 15 (Sequoia)

### Python version

3.12.13

### Docker version

n/a — frontend layout bug; reproducible on the mission detail page without a running environment.

### Steps to reproduce

1. Use any mission whose terrain `summary` runs long — e.g. a static RE mission whose summary is a full paragraph (~5–6 wrapped lines at typical pane widths).
2. Open the live run view's **Terrain** pane (split pane or the Terrain tab below 1280px), the fullscreen map overlay, or the mission detail page's **Terrain** section.
3. Compare how much vertical space the summary strip takes versus the map viewport.

### Expected behaviour

The summary should not consume the map's vertical space without bound. The map is the point of the pane; a long summary should be clamped (e.g. to ~2 lines) with an explicit control to expand it, so the graph keeps its height.

### What actually happened

The summary renders in full, always. Every wrapped line comes straight out of the map
viewport (`flex-1` with a `min-h-[360px]` floor —
`frontend/src/features/live-trace/terrain-map.tsx:817` on `main`), so on a mission with a
paragraph-length summary:

- the split/tab Terrain pane squashes the map toward its 360px floor and the pane scrolls
  to reach the zoom controls;
- the fullscreen overlay (#21) spends the top ~6 lines of the screen on prose before the
  graph starts;
- the mission detail Terrain section (fixed-height container, #20) loses the same slice.

Image

### Anything else

History: this behaviour is a documented, deliberate decision that shipped in the repo's initial public commit (`1d902259`) — the summary used to be one `truncate`d line with the rest revealed on hover, and the comment at  `frontend/src/features/live-trace/terrain-map.tsx:803-812` explains why that was removed  ("a sentence you have to hover to finish is a sentence nobody reads"). The rationale assumed summaries "run to two or three lines", where the cost is "one line of map". Shipped missions have outgrown that assumption: at paragraph length the trade inverts, and the map — the pane's actual job — pays for it.

The fix should honour the original objection to hover-gated text: clamp to ~2 lines with a persistent, explicit expand/collapse affordance (a click target, not hover-reveal). The unit test pinning the current behaviour (`terrain-map.test.tsx` — "renders the mission summary in full, with no truncation") would need updating as part of that deliberate reversal.

Observed on `main` (`9c918418`) and on the PR #21 branch; the summary block itself is untouched by #20/#21 — this dates back to the initial commit.

christo-xorciseai:
> Part of a stack based on #21 (`feat/live-terrain-fullscreen`); #31 sits on top of this PR, and #27 sits on #31.
> Merge order: #21 → this → #31 → #27.

## What changed

The terrain map's summary strip now clamps to two lines with a persistent
**Show more / Show less** click target (announcing state via `aria-expanded`). The toggle
renders only when the clamped box actually overflows — measured (`scrollHeight` vs
`clientHeight`, re-measured on window resize, element resize, and the fullscreen portal
toggle) rather than guessed from character count, since wrap width varies pane to pane. A
new run/mission resets the clamp.

Expansion is **layout-inert**: the in-flow strip stays clamped, and the full text draws as
an absolutely-positioned overlay over the top of the map, padding-matched so its first two
lines land exactly over their clamped copies (it reads as the strip growing over the graph,
not a popover appearing). The graph and its zoom/Fit cluster never move — an in-flow
expansion pushed the viewport down and shoved the controls out of a fixed-height card (the
mission page's 480px box). The overlay text is `aria-hidden`; the strip's visually-clamped
copy is the single accessible one.

## Why

Closes #22. A paragraph-length summary rendered in full, every wrapped line taken straight
out of the map viewport below — squashing the graph to its `min-h` floor in the live pane,
the fullscreen overlay, and the mission detail section (see the issue's `derelict-manifest`
repro).

This deliberately reverses the always-full-wrap decision from the initial commit while
keeping its stated principle of **no hover-gated text**: the toggle is an explicit click
target, not hover-reveal (which also doesn't exist on touch). `line-clamp` hides visually
only, so the full text stays in the DOM for assistive tech and find-in-page — and the
graph never yields height at all.

## Testing

```
$ cd frontend && npm run typecheck && npm test
tsc --noEmit: clean
Tests  683 passed (683)   # full suite on the stack head
```

The pinning test was updated to the new contract: clamped by default with the full text in
the DOM; no dead toggle while the text fits; the toggle appears on overflow; expanding
keeps the strip clamped and shows the absolutely-positioned overlay (aria-hidden text, one
toggle at a time); collapsing removes it.

**Test lanes affected**:

- [ ] `unit` — no Docker, fast
- [ ] `adapters` — port/adapter contracts
- [ ] `topology` — manifest / compose / import / extras parity
- [ ] `integration` — needs Docker or a live server process
- [x] `frontend` — Next.js typecheck / vitest

## User-facing impact

- [ ] No user-visible change (internal refactor, tests, docs only)
- [ ] **Breaking change** — describe the break and the migration path below

Visual behaviour change on the terrain map's summary strip; no contract changes.

## Documentation

- [ ] Documentation updated in this PR
- [ ] Documentation needed and tracked in a follow-up issue (link it)

## Dependencies

- [x] No dependency changes
- [ ] Dependencies changed **and** the lockfile is committed

## Checklist

- [x] **No secrets, credentials, tokens, keys, internal hostnames or customer data** appear in
      the diff — including in test fixtures, logs and comments
- [x] No references to internal-only systems (private trackers, internal URLs, internal
      process names) in code or documentation

---

- [x] I have signed the [Contributor License Agreement](../CLA.md), or will when the CLA
      assistant comments below. If I am contributing in the course of employment, my employer
      has signed the Corporate CLA.

### DIFF
--- a/frontend/src/app/globals.css
+++ b/frontend/src/app/globals.css
@@ -395,18 +395,31 @@ body {
   -webkit-mask-image: none;
   mask-image: none;
 }
-/* Expanded terrain summary, drawn OVER the map (never through the layout, so the graph and its
-   zoom controls can't be pushed). Only max-height is animated — the sheet's own height is set
-   inline by the component, from the collapsed strip's height to its content height — so opening
-   and closing read as one continuous growth out of the two-line strip it covers. */
-.tm-summary-sheet {
-  transition: max-height 0.2s ease;
+/* Terrain summary: the copy uses the full width and the amber action occupies only the end of the
+   final visible line. A short card-coloured fade protects the label without reserving a wide
+   right-hand column on both lines; the persistent panel still grows over the map. */
+.tm-summary-slot {
+  height: calc(2lh + 1rem);
+}
+.tm-summary-panel {
+  max-height: calc(2lh + 1rem);
+  transition: max-height 0.2s cubic-bezier(0.2, 0, 0, 1);
+}
+.tm-summary-copy {
+  max-height: 2lh;
+  transition: max-height 0.2s cubic-bezier(0.2, 0, 0, 1);
+}
+.tm-summary-toggle {
+  background: linear-gradient(to right, transparent, var(--color-card) 1.75rem);
+  transition: color 0.15s ease;
 }
 
 @media (prefers-reduced-motion: reduce) {
   .tm-edge-note { transition: none; }
-  /* Calm (guide §11/§12): the sheet still opens and closes, it just arrives at its size. */
-  .tm-summary-sheet { transition: none; }
+  /* Calm (guide §11/§12): the summary still opens and closes, it just arrives at its size. */
+  .tm-summary-panel,
+  .tm-summary-copy,
+  .tm-summary-toggle svg { transition: none; }
   /* Guide §11/§12: no glow/flicker loops in the workspace under reduced motion
      (Calm). Freeze the terrain pulses/glow/flow to their resting state. */
   .tm-ring-amber,
--- a/frontend/src/app/globals.css
+++ b/frontend/src/app/globals.css
@@ -395,8 +395,18 @@ body {
   -webkit-mask-image: none;
   mask-image: none;
 }
+/* Expanded terrain summary, drawn OVER the map (never through the layout, so the graph and its
+   zoom controls can't be pushed). Only max-height is animated — the sheet's own height is set
+   inline by the component, from the collapsed strip's height to its content height — so opening
+   and closing read as one continuous growth out of the two-line strip it covers. */
+.tm-summary-sheet {
+  transition: max-height 0.2s ease;
+}
+
 @media (prefers-reduced-motion: reduce) {
   .tm-edge-note { transition: none; }
+  /* Calm (guide §11/§12): the sheet still opens and closes, it just arrives at its size. */
+  .tm-summary-sheet { transition: none; }
   /* Guide §11/§12: no glow/flicker loops in the workspace under reduced motion
      (Calm). Freeze the terrain pulses/glow/flow to their resting state. */
   .tm-ring-amber,
--- a/frontend/src/features/live-trace/terrain-map.test.tsx
+++ b/frontend/src/features/live-trace/terrain-map.test.tsx
@@ -65,14 +65,12 @@ describe("TerrainMap", () => {
     ],
   } as Partial<ResolvedTerrainV2>;
 
-  test("clamps the summary to two lines behind an explicit toggle (issue #22)", async () => {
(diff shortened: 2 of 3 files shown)


christo-xorciseai:
Follow-up (`be60fd0f`) after review feedback that the toggle felt jarring — "the whole terrain re-renders". Two distinct causes, both fixed:

**1. It really was re-rendering the whole graph.** The entire SVG subtree was inline in the component body, so *any* state change reconciled every group, edge and node — plus one `foreignObject` per edge note whose fresh inline `style` object re-applied the custom properties that feed `.tm-edge-note`'s `max-height` transition. So a summary toggle re-rendered the map, and so did the legend toggle, the wheel tip, and **every pointermove of a pan**. The graph is now memoised on the inputs that actually change it (layout, fold, colours, hover, selection), so unrelated UI state can't touch it — this also makes panning markedly cheaper.

**2. The sheet snapped.** It appeared and disappeared at full size. It now *grows*: `max-height` animates from the strip's own height (so it starts exactly congruent with the two clamped lines it covers) to its measured content height, capped to the card, and reverses on close — the sheet is kept mounted for the exit via the existing `useExitTransition` seam, so the motion is continuous in both directions rather than animating open and vanishing shut. The transition lives in `.tm-summary-sheet` beside the `.tm-edge-note` precedent and is dropped under `prefers-reduced-motion`.

Verified: `npm test` 682 passing on this branch (683 on the stack head), `tsc --noEmit` clean, and `next build` compiles + exports cleanly (the measurement uses a layout effect, guarded to `useEffect` on the server since these pages are statically prerendered).

christo-xorciseai:
## Collapsed:
image


## Expanded:
image

guru-xorciseai:
Landed in #27 as `c440ecd`. This PR was closed unmerged when the stack broke (#21 was squash-merged, which rewrote the ancestry these branches were built on). The commits were rebased `--onto main` and merged via #27 — the introduced patch was verified byte-identical to the pre-rebase stack tip, so nothing from here was lost.

PR #32, opened 2026-08-05
feat(missions): make Other providers a designed coming-soon state

guru-xorciseai:
## What

The **Missions → Other providers** tab rendered a dashed box with a centred icon and two sentences. That box was structurally the same component as `NoResults`, twelve lines away in the same file — so two unrelated states wore the same costume:

- *"Your filters matched nothing"* — a dead end the user clears.
- *"Other providers"* — a capability XORCISE has committed to and not shipped.

The house treatment for the second kind already existed — `ComingSoonPanel`, the amber-wash panel behind the **Ingest a bundle** dialog and the Settings **Distributed deployment** card. The catalog was the only surface not using it.

This makes the tab a **field** rather than a box: a provider constellation fills the canvas — the catalog at the centre, two lit spokes for the providers that exist today, six dashed slots for the ones that do not — with the message centred over it as a lens.

## Before / after

| | Before | After |
|---|---|---|
| **Desktop** 1920×1080 |  |  |
| **Laptop** 1440×900 |  |  |
| **Tablet** 820×1180 |  |  |
| **Mobile** 390×844 |  |  |

> Screenshots live on the standalone `pr-assets/other-providers-canvas` branch so no binaries land in `main`. **Delete it after merge:**
> `git push origin --delete pr-assets/other-providers-canvas`

## Why it is drawn this way

The field borrows the **terrain map's** existing visual vocabulary rather than inventing one: amber for a live node, `--color-muted-foreground` for unknown, a flowing dash on an active edge, an expanding halo on a live node. No new colour, font, or system primitive.

Reusing its two motion classes (`tm-flow`, `tm-ring-amber`) also inherits their reduced-motion rule from `globals.css`, where guide §11/§12 already bans glow and flicker loops in the workspace. A hand-rolled keyframe would have needed its own media query and would eventually have drifted from that rule.

The blur is depth of field, not glass: the field is drawn sharp everywhere and only the card's `backdrop-filter` throws it out of focus, so the spokes visibly soften as they pass beneath it.

`ComingSoonPanel` gains an opt-in `align="center"`. **The two existing hosts keep the default `align="start"` and render unchanged** — verified by screenshot.

## Layout fixes that fell out of building it

**Catalog scroller → flex column.** The tab body now claims the height that is left instead of estimating it with `calc(100vh - 23rem)`. That estimate did not know the pane's own chrome and overshot, putting a scrollbar on a tab with nothing below the fold. A `-mx-4` bleed did the same on the other axis — the scroller is `overflow-y-auto`, which forces `overflow-x: auto`, so 16px of bleed bought a full horizontal scrollbar. Grid tabs still scroll normally; the sticky filter bar still pins.

**Stats strip → explicit grid.** `flex-wrap` let the row decide how many cells fit and produced 3-up with **difficulty** dropped alone onto a second row — the widest cell, given the whole width, with nothing beside it. Now stated outright:

| viewport | layout |
|---|---|
| `< 480px` | 1 column |
| `480–1023px` | 2×2 |
| `≥ 1024px` (`lg`) | one row of 4, with dividers |

The `lg` threshold is derived, not picked: a difficulty row is a label + five fixed dots + a count ≈ **210px that cannot compress**, so four of them plus gaps need ~912px of card interior ≈ a 1024px viewport.

**Difficulty row.** Capped, so the count stops drifting from the meter it belongs to, and its cell rule now draws only in the single-row layout — a `border-left` separates two cells only when there is a cell to its left; in the 2×2 it landed on the cell that *starts* a row and read as a stray line.

## Verification

- `npx tsc --noEmit` — clean
- `npx vitest run` — **698 passed / 95 files**, 0 failures (re-measured on CI after rebasing onto `c440ecd`)
- `pytest tests/topology` — **18 passed** (import-linter contracts + parity)
- CLI/GUI parity lanes (`test_cli_parity`, `test_cli_json_parity`, `test_cli_catalog_ux`) — green
- **No scroll on either axis**, measured as `scrollHeight − clientHeight`, at 1920×1080, 1600×950, 1536×864, 1440×900, 1366×768, 1280×720 and 1024×768
- Reduced motion: all three animations report `animation-name: none` under `prefers-reduced-motion: reduce`
- Grid tabs still scroll (1417–2145px); sticky filter bar still pins (`stickyTop === scrollerTop` after a 900px scroll)

Mobile still scrolls, as it did before this change (225px → 513px) — the stats dashboard alone fills a 390×844 viewport, and the coming-soon card is taller than the dashed box it replaces.

## No CLI change

`CatalogEntry.source` is `Literal["your_own", "library"]`, and `mission list --source` accepts exactly `library` and `your-own`. A third-party provider is not representable in the domain yet, so there is nothing for the CLI to list or label. The tab predates this PR, so the CLI ↔ GUI gap is unchanged by it. When providers do ship, the contract's `Literal` widens first, then `_SOURCES` and `source_label` follow.

## Scope

Frontend only — 5 files, no API, contract, or CLI surface touched.

### DIFF
--- a/frontend/src/components/ui/coming-soon.tsx
+++ b/frontend/src/components/ui/coming-soon.tsx
@@ -25,16 +25,28 @@ import { cn } from "./cn";
 export function ComingSoonPanel({
   headline,
   steps,
+  align = "start",
   className,
   children,
 }: {
   headline: string;
   /** Illustrative, not interactive — the shape of the flow. */
   steps: { icon: ReactElement<{ className?: string }>; label: string }[];
+  /**
+   * `start` (default) is the reading layout the two original hosts use — a dialog and a
+   * settings card, where the panel sits in a column of other left-aligned content.
+   *
+   * `center` is for a panel that IS the surface, with nothing to align to: it centres the
+   * axis, drops the measure caps so the prose fills the box instead of stopping short of
+   * it, and moves the sparkle to the corner so it stops pulling the top row off-centre.
+   * Opt-in rather than inferred, so neither existing host shifts.
+   */
+  align?: "start" | "center";
   className?: string;
   /** Body prose. */
   children: ReactNode;
 }) {
+  const centered = align === "center";
   return (
     <div
       className={cn(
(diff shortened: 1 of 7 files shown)


christo-xorciseai:
The changes look good but I think the background pulsing orbs looks a bit blurry and distracting. Do you think we can replace that artifact with a simple background gradient ? 

image


Also, I think we should extend the coming soon banner to the locally ingested tab section as well ? Currently it looks quite misleading.

image

guru-xorciseai:
Added a second commit for the misleading empty state on **Your Own**.

It read *"No local missions yet. Ingest a bundle to add one."* — which instructs the reader to perform the one action this build cannot perform, since the **Ingest a bundle** button eight pixels above it opens a coming-soon preview rather than a file browser. Until ingestion ships that is every install, so it is the first thing a new operator reads.



It now carries the same coming-soon panel the Ingest dialog does, then points at the shelf that works today.

The two surfaces **share one component** rather than repeating the copy — they make the same promise about the same unshipped capability, and written out twice they would eventually disagree about what ingestion does.

Only reached when no filters are active: a search that matches nothing still gets `NoResults`, which is a different problem with a different way out. There's a test pinning that the two cannot collapse into one.

`Grid`'s `empty` widens from `string` to `ReactNode` and is rendered as given rather than wrapped in a `<p>`, which would have set the panel in body type.

The background field on **Other providers** is unchanged in this push — keeping it as-is for now.

`tsc` clean · **700 frontend tests** (698 + 2 new) · no scroll on either axis.

christo-xorciseai (approved):
Looks good to me 👍

PR #35, opened 2026-08-06
fix(runner): strip mission host port publishes

### LINKED ISSUE #34: [Bug]: mission-authored host port publishes are neither stripped nor preflighted (breachpoint binds host :5000)
### Before you file

- [x] I searched existing issues and this is not a duplicate.
- [x] This is not a security vulnerability (those go to SECURITY.md).

### XORCISE version

`0.1.2.dev8+gc440ecd49`

### How did you install XORCISE?

From a git checkout (editable install)

### Operating system

macOS 15.7.7 (build 24G720), Apple Silicon — Docker Desktop, host-daemon sibling mode

### Python version

3.12.13

### Docker version

Docker version 29.6.2, build dfc4efb

### Steps to reproduce

1. Enable **System Settings → General → AirDrop & Handoff → AirPlay Receiver** (macOS binds host `:5000`). This is on by default on many Macs.
2. `xorcise up`
3. `xorcise mission pull breachpoint`
4. `xorcise run create --agent <agent> --mission breachpoint`

The run fails during environment bring-up. Any host process holding `:5000` reproduces it; AirPlay Receiver is simply the common one on macOS.

### Expected behaviour

The mission deploys. Agents reach every mission service over the tailnet at its carved subnet IP on the service's **container** port, so no mission needs a host port published at all — a host-side listener on `:5000` should be irrelevant to a run.

Failing that, the conflict should be caught by preflight and reported with the offending port named, rather than surfacing as a generic `deploy_failed`.

### What actually happened

`compose up` cannot bind the host port and the run is closed out `deploy_failed`, with nothing in the output naming the port or the conflict. `xorcise doctor` reports "No problems found" throughout.

Two independent gaps combine here:

**1. Mission-authored `ports:` are never stripped.** `containers/mission-base/entrypoint.sh:39` composes the mission's own `docker-compose.yml` verbatim:

```
docker compose -p "${XORCISE_PROJECT:-mission}" \
    -f /mission/docker-compose.yml -f /mission/net-override.yml up -d
```

`build_net_override` (`src/xorcise/core/runner/netoverride.py`) pins subnets, adds the router and merges `static_ips`, but never neutralises `ports:`. In macOS host-daemon mode the mission stack is composed against the **host** daemon, so any mission-authored publish becomes a binding on the operator's Mac.

**2. Preflight does not scan mission ports.** `src/xorcise/core/cli/_preflight.py` only knows the control-plane ports:

```python
_PLANE = {3001: "rest", 4318: "otlp", 8800: "runner", 8080: "headscale"}
```

`ports_in_use()` / `conflict_message()` are never given the mission's ports, so the collision is discovered by Docker at deploy time instead of by XORCISE before it.

### Logs / output

```shell
# Catalog-wide scan: pulled all 26 library missions and extracted
# /mission/docker-compose.yml from each image.
#
#   26 missions   19 lab (compose inspected)   7 static (no image, by design)
#
# Only 2 of 19 lab missions publish host ports:

breachpoint          kusto=8081:8080 ; web=5000:5000      <-- binds host :5000
operation-tessera    eval-harness=5055:5000

# The other 17 lab missions publish nothing. ghostwire's compose states the
# intended convention explicitly:
#
#   "the platform routes to container port 80 via the tailnet IP
#    (no host ports are published)"
```

### Anything else

**Scope of the fix.** The host publishes appear to be unnecessary in every case — the agent reaches these services over the tailnet by carved IP on the container port, which is why 17 of 19 lab missions publish nothing. `breachpoint`'s `web` maps `5000:5000` and `operation-tessera`'s `eval-harness` maps `5055:5000`; in both cases the container port is what matters and the host publish is pure collision surface.

Suggested direction (happy to send a PR if the approach is agreed):

1. **Strip `ports:` in the net-override.** Compose merges `ports` additively across files, so it cannot be removed by overlay alone — it likely needs the mission compose to be rewritten before `compose up`, or a documented rule that mission bundles must not publish host ports (enforced at ingest/build).
2. **Extend preflight to the mission's ports.** `resolve_ports()` / `conflict_message()` already model exactly this; they just need to be handed the mission's declared ports before deploy so the failure names the port.

**Note on the two affected missions.** `breachpoint` is the one that actually collides with AirPlay Receiver. `operation-tessera` publishes host `5055`, so it is *not* affected by an AirPlay conflict — worth stating because these two are easy to conflate when the container port in both cases is `5000`.

This is a platform issue, not a mission-content one: the manifest schema (`environment`: `compose_file`, `entry_networks`, `static_ips`) has no way to declare a host port, so there is currently no layer at which such a publish is reviewed or validated.

christo-xorciseai:
## What changed

Before each mission deployment, the runner inspects the already-pulled fused image without starting mission code and identifies every Compose service declaring `ports:`. It adds `ports: !reset []` for those services to the existing per-run network override, alongside the carved subnet, router, and static-IP configuration.

The inspection is bounded, fail-closed, cleaned up in all cases, and cached by immutable image ID. Because existing remote images already consume `/mission/net-override.yml`, no mission or base-image rebuild is required.

## Why

Mission services are reached over their carved tailnet subnet at the service container IP and container port. Host publications are unnecessary and can collide with unrelated listeners on the operator host, notably macOS AirPlay on port 5000.

Closes #34

## Testing

```text
uv run pytest tests/unit -q
1949 passed

uv run pytest tests/adapters -q
128 passed

uv run ruff check .
All checks passed!

uv run ruff format --check .
499 files already formatted

uv run mypy
Success: no issues found in 475 source files

uv run lint-imports
Contracts: 4 kept, 0 broken
```

Remote-image validation:

- Pulled the existing remote `breachpoint` image through the patched checkout.
- Deployment-time inspection detected authored publications on `kusto` and `web`.
- Merged the generated override using the image embedded Compose v2.33.0; the effective service model contained no published ports.
- Held host port 5000 open and started a service with an authored `5000:80` publication using the dynamically generated override; it started successfully and Docker reported `published_ports={}`.

**Test lanes affected:**

- [x] `unit` — no Docker, fast
- [x] `adapters` — port/adapter contracts
- [x] `topology` — manifest / compose / import / extras parity
- [x] `integration` — needs Docker or a live server process
- [ ] `frontend` — Next.js typecheck / vitest

This runner/networking change should receive the `full-ci` label before merge.

## User-facing impact

- [x] Backwards-compatible correction: mission-authored host publications are suppressed while container and tailnet ports remain unchanged

Existing installed remote missions receive the fix on their next deployment after upgrading XORCISE. They do not need to be pulled or rebuilt again.

## Documentation

## Dependencies

- [x] No dependency changes

## Checklist

- [x] No secrets, credentials, tokens, keys, internal hostnames or customer data appear in the diff
- [x] No references to internal-only systems appear in code or documentation

## Release notes

Suggested label: `bug`

---

- [x] I have signed the Contributor License Agreement, or will when the CLA assistant comments below.

### DIFF
--- a/src/xorcise/core/runner/docker/__init__.py
+++ b/src/xorcise/core/runner/docker/__init__.py
@@ -114,6 +114,14 @@ class DockerDriver(ABC):
     def image_exists(self, image: str) -> bool:
         """Whether the image is already in the local store (local-store-first pull)."""
 
+    def published_port_services(self, image: str) -> tuple[str, ...]:
+        """Services whose authored Compose definitions publish ports on the Docker host.
+
+        Real drivers inspect the already-pulled fused image without starting it. The default is
+        empty for non-Docker drivers; their placeholder images have no mission stack to inspect.
+        """
+        return ()
+
     @abstractmethod
     def reap_managed(self) -> list[str]:
         """Force-remove every xorcise-managed per-run container; return what was reaped.
@@ -175,6 +183,7 @@ class StubDockerDriver(DockerDriver):
         # unknown, so existing stub-mode behaviour (a plain READY) is preserved.
         self.container_states: dict[str, ContainerState] = {}
         self.service_states: dict[str, tuple[ServiceState, ...]] = {}
+        self.port_services: dict[str, tuple[str, ...]] = {}
         self.compose_projects: set[str] = set()  # test-settable projects holding a network
         self.removed_projects: list[str] = []  # projects released by remove_run_resources
         self.specs: list[ContainerSpec] = []  # every run()'s spec, for assertions
@@ -198,6 +207,9 @@ class StubDockerDriver(DockerDriver):
 
     def image_exists(self, image: str) -> bool:
         return image in self.present
+
+    def published_port_services(self, image: str) -> tuple[str, ...]:
+        return self.port_services.get(image, ())
 
     def run(self, spec: ContainerSpec) -> ContainerHandle:
         self._counter += 1
--- a/src/xorcise/core/runner/docker/driver.py
+++ b/src/xorcise/core/runner/docker/driver.py
@@ -6,10 +6,16 @@ the `runner` extra; only constructing the driver requires the SDK + a daemon.
 
 from __future__ import annotations
 
+import contextlib
+import io
 import json as _json
 import platform as _platform
-from collections.abc import Callable
+import tarfile
+from collections.abc import Callable, Mapping
+from pathlib import PurePosixPath
 from typing import Any
+
+import yaml
 
 from xorcise.core.runner.docker import (
     MANAGED_LABEL,
@@ -21,6 +27,27 @@ from xorcise.core.runner.docker import (
     PullProgress,
     ServiceState,
 )
+
+_MISSION_COMPOSE_PATH = "/mission/docker-compose.yml"
+_MAX_COMPOSE_BYTES = 4 * 1024 * 1024
+
+
+def _published_port_services(content: bytes | str) -> tuple[str, ...]:
+    """Parse authored Compose and return services that declare host publications."""
+    try:
+        compose = yaml.safe_load(content)
+    except yaml.YAMLError as exc:
+        raise RuntimeError("mission docker-compose.yml is not valid YAML") from exc
+    if not isinstance(compose, Mapping):
+        raise RuntimeError("mission docker-compose.yml must contain a mapping")
+    services = compose.get("services")
+    if not isinstance(services, Mapping):
+        raise RuntimeError("mission docker-compose.yml must contain a services mapping")
+    return tuple(
+        str(name)
+        for name, spec in services.items()
+        if isinstance(spec, Mapping) and "ports" in spec
+    )
 
 
 def _parse_compose_ps(output: bytes | str) -> tuple[ServiceState, ...]:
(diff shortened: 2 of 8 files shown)


containers/mission-base/entrypoint.sh
```diff
@@ -35,9 +35,14 @@ if [ -n "${XORCISE_HEADSCALE_CA_B64:-}" ]; then
...
 
-# 4. bring up the mission + the router (compose interpolates ${XORCISE_AUTHKEY} etc. from env)
+# 4. bring up the mission + the router (compose interpolates ${XORCISE_AUTHKEY} etc. from env).
+# The builder-generated reset override removes mission-authored host publishes: agents reach
+# mission services over the carved tailnet subnet, and publishing on the operator host creates
+# needless collision surface (for example macOS AirPlay commonly owns port 5000).
 docker compose -p "${XORCISE_PROJECT:-mission}" \
-    -f /mission/docker-compose.yml -f /mission/net-override.yml up -d
+    -f /mission/docker-compose.yml \
+    -f /mission/xorcise-port-reset.yml \
```
guru-xorciseai: Is it necessary to add this? coz this will involve rebuilding the missions in XORCISE Remote.

guru-xorciseai (changes requested):

containers/mission-base/entrypoint.sh
christo-xorciseai: Yea this should not be the way should be handling the port exposition.

christo-xorciseai:
This MR has been super-seeded by #41

Issue #37, opened 2026-08-06
[Bug]: teardown leaks unlabelled, never-started mission containers in host-daemon mode (fixed container_name then wedges all future runs)


christo-xorciseai:
### Before you file

- [x] I searched existing issues and this is not a duplicate.
- [x] This is not a security vulnerability (those go to SECURITY.md).

### XORCISE version

`0.1.2.dev8+gc440ecd49` (mechanism verified against `main`; originally observed on a branch carrying the #34 port-strip fix, which is unrelated to this bug)

### How did you install XORCISE?

From a git checkout (editable install)

### Operating system

macOS 15.7.7 (build 24G720), Apple Silicon — Docker Desktop, host-daemon sibling mode

### Python version

3.12.13

### Docker version

Docker version 29.6.2, build dfc4efb

### Steps to reproduce

In macOS host-daemon mode, a mission compose service that is **(a)** not stamped with an `xorcise.*` label and **(b)** never reaches `started` survives teardown entirely.

Mission-level reproduction (`breachpoint`, whose `kusto-init` helper matches both conditions):

1. `xorcise up`
2. `xorcise run create --agent <agent> --mission breachpoint`
3. Let the deploy fail before the dependency graph fully starts (the amd64-emulated Kusto emulator is slow to reach `started`, so `kusto-init` stays in `created`).
4. `xorcise run create --agent <agent> --mission breachpoint` again → fails at readiness, permanently.

**Minimal reproduction (no mission required)** — this isolates the mechanism to four Docker commands:

```bash
docker network create xtest-net
docker create --name xtest-created --network xtest-net \
  --label com.docker.compose.project=xtestproj alpine:latest sleep 3600

docker network inspect xtest-net --format '{{json .Containers}}'   # => {}
docker network rm xtest-net                                        # => succeeds, no 403
docker inspect xtest-created --format '{{.State.Status}}'          # => created (survives)
docker create --name xtest-created alpine:latest sleep 60          # => Conflict
```

### Expected behaviour

Teardown releases every container belonging to a run's compose project, regardless of label or lifecycle state, so a failed deploy leaves nothing behind on the operator's daemon.

### What actually happened

The container is missed by all three reap paths at once and is left dangling. Because the affected mission services use fixed `container_name` values, the orphan then wedges **every future run of that mission on that host** with `Conflict. The container name "/kusto-init" is already in use`. Neither `xorcise down` nor per-run teardown clears it; only a manual `docker rm -f` does.

[...]

PR #41, opened 2026-08-11
feat(runner)!: DinD-only on docker:29.7.1, plus base-generation compatibility gating (CLI + UI)

christo-xorciseai:
## Summary

Makes the macOS runtime work, removes the topology that was papering over it, and — because removing the fallback means an incompatible base now fails hard with no recourse — turns the base image into a **versioned compatibility contract** that is checked up front and surfaced in the UI.

The mission stack now **always** runs inside the run's own container (DinD). The host-daemon "sibling" mode is gone, along with the `macos_container_runtime` setting. A host that cannot nest, or a mission built on a base generation this XORCISE cannot run, is refused with an actionable message before any expensive work — and the incompatibility shows on the catalog card and mission page, not only as a failed run.

Ten commits, grouped into three themes:

| | Commit | What |
| --- | --- | --- |
| 1 | `feat(runner)` | Nested-container capability probe + plumbing (no behaviour change) |
| 2 | `docs(runner)` | Root-cause of the macOS failure — it was never nesting |
| 3 | `docs(runner)` | Correction: the base bump is one of two fixes, not the only one |
| 4 | `feat(mission-base)` | `docker:27-dind` → `docker:29.7.1-dind`; pin the router image |
| 5 | `feat(runner)!` | **Breaking:** DinD only; fail closed when a host cannot nest |
| 6 | `fix(rest)` | Map the nesting refusal to 503 so it survives the HTTP hop |
| 7 | `feat(runner)` | **Base-generation compatibility gate** — label + tag, refuse an incompatible fuse |
| 8 | `fix(rest)` | **Harden the probe** — no stale disk cache, no false 500s, honest remediation, boot pre-warm |
| 9 | `feat(missions)` | **Surface the incompatibility in the catalog UI** (card badge + detail banner) |
| 10 | `test(cli)` | Accept the new `timeout` kwarg in a run-create post fake |

Commits 7–9 were driven by validating the runner change against the whole mission library (all 19 lab missions deployed in parallel), which surfaced the upgrade-path and probe-robustness gaps the first six commits left open.

Mirrors `fifthdomain/cloud-infrastructure/xorcise-ai!50` (the same base bump + `DOCKER_HOST` fix for the cloud-owned copy). The base-generation *versioning* contract for the cloud side is tracked separately — see Follow-ups.

## Why remove the host-daemon topology

It composed every mission's stack on the **operator's own daemon** — one shared namespace for all runs. Invisible with a single run, fatal with several:

- parallel runs collide on missions' fixed `container_name`s and on published host ports - teardown leaks un-labelled mission containers onto the operator's machine - mission bind mounts resolve against the host filesystem instead of the bundle

Issues #34 and #37 and the bind-mount failure are all the same root cause, and none of them can be fixed while the topology exists.

## Why the base bump was needed first

Rosetta cannot exec a binary reached through the `/proc/<pid>/exe` magic symlink. Minimal repro, in a plain amd64 container with **no DinD anywhere**:

``` /bin/busybox true -> ok /proc/<pid>/exe true -> rosetta error: failed to open elf at true ```

Docker <= 27 installs an OCI prestart hook doing exactly that on every container it creates (`{"path":"/proc/<dockerd-pid>/exe","args":["libnetwork-setkey","-exec-root=…"]}`), which is why the failure reads `failed to open elf at -exec-root=/var/run/docker`. Docker 28 removed the hook ([moby#47406](https://github.com/moby/moby/pull/47406)).

Measured on one unchanged host:

| wrapper arch | engine | prestart hook | nested amd64 |
| --- | --- | --- | --- |
| amd64 | 27.5.1 | present | **fails** |
| amd64 | 28.5.2 | absent | works |
| amd64 | 29.7.1 | absent | works |
| arm64 | 27.5.1 | present | works — dockerd is native, Rosetta never sees the hook |

So sibling mode never "avoided nesting" — it moved container creation off the inner Docker 27 daemon onto Docker Desktop's much newer one.

**29 over 28** because 28.x reached end of life on 2026-05-13 and 29.x is the only branch still receiving security patches — which matters for an image that runs adversarial agents against deliberately vulnerable targets. Pinned rather than floating because this hook was removed in 26.0, [restored](https://github.com/moby/moby/pull/47646), then removed again in 28.

## Nesting precondition — graceful failure

**Where: run creation, not `up`.** `up` must keep working on a host that cannot nest — static missions, the UI and past runs need nothing nested. The check runs after the static short-circuit and, for an auto-pulled mission, as the pull **precheck** — after the cheap manifest fetch, before the multi-GB image download — so an unusable host is refused without first downloading an image it could never run, let alone reserving a subnet or minting an auth key.

**What: Tier 2 is the only gate.** Actually starting a container inside a container is the thing we need, and it means the same on every platform. Tier 1 (the Rosetta binfmt registration) is deliberately **not** a gate — Linux has no such handler at all, so gating on it would refuse every Linux host. It earns its keep as the *explanation* for why Tier 2 failed, which is otherwise buried in an OCI error chain.

**How: a typed error that survives the HTTP hop.** `run create` POSTs to `/api/runs`, so the check runs server-side and never reaches the CLI's own error guard. `ContractError` is not a `RuntimeError`, so without an explicit handler it fell through as an unhandled **500 with a non-JSON body**, dropping the diagnosis the CLI reads out of `detail`. Commit 6 maps it to 503. Commit 8 also wraps the `docker.from_env()` factory failure (a `DockerException`, likewise not a `RuntimeError`) into `RuntimeError → 503`, closing the same class of unhandled-500 for a daemon that drops mid-request.

**Verdict caching — in memory, never on disk.** The Tier 2 probe (~10–40 s) is memoised **in the server process**, lock-serialised so concurrent first runs queue behind one probe instead of each launching a privileged DinD. There is deliberately **no on-disk cache**: an earlier version persisted the verdict against a host fingerprint, which turned a transient probe failure (a registry blip) into a *permanent* refusal keyed on a fingerprint that never cleared. Boot pre-warms the memo in a background task so the first real run reads it instantly rather than paying the probe under the CLI's read timeout (which is now 300 s on run-create, so a slow probe can never time the CLI out into a duplicate run). `doctor` re-probes fresh (and refreshes the memo), so fixing Rosetta and re-running `doctor` unblocks runs without a restart. The probe container carries the `xorcise.managed` label, so a crash mid-probe leaves something `xorcise down` can reap rather than a stranded privileged DinD.

**Remediation is host-aware.** The Rosetta fix is shown only on macOS (`host_is_macos()`); a non-Rosetta nesting failure on Linux gets the generic "can this host run privileged containers" fix, not "enable Rosetta in Docker Desktop" for settings that do not exist.

`nested_container_check=skip` bypasses the **check** for hosts where the probe itself cannot run (restricted CI). It can never restore the sibling topology.

## Base-generation compatibility gate

With the host-daemon fallback gone, a fused image built on an incompatible base (e.g. the engine-27 base whose prestart hook this PR exists to escape) dies at deploy with no recourse. So the base is now a versioned compatibility contract:

- `containers/mission-base/Dockerfile` declares `LABEL ai.xorcise.base.version="2"`, inherited by every `FROM xorcise/mission-base` fused image. Only the **major** gates compatibility — a minor/patch base change never invalidates an install. `build.BASE_VERSION` and the label are kept in step by `tests/topology/test_dind_base_parity.py`. - `ensure_base_image()` now **rebuilds** a present base of the wrong (or unlabelled, pre-versioning) generation, instead of the bare presence check that let an upgraded host silently keep the old base — the exact way the PR's core fix used to fail to reach existing installs. - `run create` refuses an artifact whose base major ≠ what this XORCISE runs, via a resolution ladder: the inherited label (a local `:local` fuse carries no generation in its tag), else the `-baseN` tag suffix (every published/pulled artifact), else allow-with-warning (undeterminable). The refusal is direction-aware — reinstall the mission if the artifact is older, upgrade XORCISE if it is newer — as a typed `BaseImageIncompatibleError` mapped to 409.

Mirrors the cloud's `<content_hash>-base<N>` generation tags. The full versioning *scheme* (dotted minor tags, a creator/challenge version, multi-arch) is a separate cloud contract — see Follow-ups.

## Surfaced in the catalog UI

The incompatibility used to appear only as a failed run. `CatalogEntry` now carries `base_version` / `compatible` / `compat_hint`, derived from the image ref via the **same** `build.base_compat` the run-create gate uses — so a card can never say "runnable" for something a run would refuse. Rendered with existing UI elements:

- **Mission card:** an amber `update required` badge (the existing `Badge` `warn` variant), grouped with the installed/available status, its remediation on the tooltip. - **Mission detail:** a warning banner in the `Card`/`CardContent` treatment the Preview banner already uses, naming the fix and the base it was built for, above the run button.

An undeterminable ref (a local `:local` fuse) shows no warning; the run-create gate still guards it via the image label.

## Router image is now pinned

`build.py` pulled `netoverride.ROUTER_IMAGE` (`tailscale/tailscale:stable`) directly, so re-fusing an old mission baked whatever `:stable` meant that day. It now pulls a pinned version and re-tags to the canonical name — the **deploy-time** contract the per-run override resolves, so baking under anything else strands the mission at `compose up`. Both failure modes are asserted in one test.

Commit 8 also pins the **orchestrator-router** in `headscale/provision.py`: it consumed the host's `tailscale/tailscale:stable` tag, which the fuse re-tags — so a mission build silently changed which Tailscale version the control-plane router ran.

## Deliberately kept

The compose-project teardown sweeps in `_remove_run_resources`. They are no-ops for DinD-era runs, but they are the only thing that cleans up containers left behind by runs created **before** this change — deleting them would strand those forever.

## Verification

- All 19 lab missions deploy inside the run container on the pinned base (validated in parallel); `doctor` reports `✓ nested containers: nested amd64 verified (x86_64 inside an amd64 DinD)`. - Forcing the engine-27 base produces the full nesting refusal instead of a silent degrade; flipping the required base major produces the direction-aware 409 (reinstall / upgrade) end to end, and the card badge + detail banner render for it. - New tests cover: the `base_compat` verdict and its exposure on the catalog entry, the in-memory memo (probe once, `fresh` re-probe, client closed), the macOS-vs-Linux remediation split, the Dockerfile-label parity, the (previously vacuous) doctor gate, and the card/detail UI. - CI green: unit / adapters / topology / smoke / frontend (typecheck, tests, e2e) / build-dist / CodeQL / CLA / ci-ok. `ruff` + `mypy` (481 files) + all 4 import contracts clean.

## Docker 28's bridge-firewall rework — investigated, does not hit the production path

The 28+ FORWARD-chain rework (`DOCKER-FORWARD`) blocks a container **bridging two Docker bridge networks** from forwarding between them. Reproduced with an identical test in each engine (native arch, to isolate the engine as the only variable):

| engine | dual-homed container forwards **bridge → bridge**? |
| --- | --- |
| docker:27.5.1 | **yes** — old per-bridge `-i br ! -o br -j ACCEPT` rules |
| docker:29.7.1 (shipped) | **no** — `-A FORWARD -j DOCKER-FORWARD`, whose rules are all keyed on docker bridge interfaces (`-i br-…`) |

But the production tailnet data plane does **not** use that pattern. The per-run subnet router forwards from `tailscale0` (a TUN, **not** a docker bridge) into the mission bridge — a non-bridge ingress the `DOCKER-FORWARD` rules never match. Verified on docker:29.7.1 with the router **dual-homed on two bridge networks** (the two-`entry_network` topology), using a plain non-bridge veth in a sub-netns as the `tailscale0` stand-in (identical to `tailscale0` from netfilter's view — neither is a docker bridge):

| ingress → egress through a dual-homed router (29.7.1) | result |
| --- | --- |
| non-bridge iface → mission bridge (**the production pattern**) | **forwards** |
| bridge → bridge | blocked |

So the isolation is bridge-to-bridge only, and a mission declaring two `entry_networks` stays reachable — the subnet router only ever forwards tailnet → each bridge, never bridge → bridge. (All current lab missions declare a single `entry_network` regardless.)

Residual, non-blocking: a full live `tailscaled` + Headscale run was not performed — the netfilter mechanism the rework changed is tested against the production ingress pattern, but the end-to-end stack is the gold standard. Separately, a *mission-internal* pivot that routes between two of a mission's own networks through a container would hit the bridge-to-bridge block — a mission-authoring concern, unrelated to the tailnet router, and no current mission does it.

The Docker-heavy CI lanes (`test-heavy`, real containers) are skipped on PRs — worth a `full-ci` run before merge given this is the runner rewrite.

## Follow-ups, not in this PR

- **Cloud versioning contract:** the base-generation scheme's remote half (the `ai.xorcise.base.version` label on the cloud base, dotted `-baseN.M` tags, `base_version` in the catalog, and the challenge/multi-arch axes) — a separate MR against `cloud-infrastructure/xorcise-ai`. - **Gate uninstalled missions before the pull:** currently the base gate runs post-pull for a not-yet-installed mission (the local image is what carries the generation); needs the catalog to expose `base_version` (the cloud contract above). - **Mission-image pruning (#44):** a base bump strands the previous generation's fused images with no supported way to reclaim them — the label makes staleness queryable. - **Canonical xorcise-owned router tag:** re-tagging under `tailscale/tailscale:stable` on the host is neutralised by the orchestrator-router pin, but the proper fix (a `xorcise/router` tag) forces a fleet re-fuse and belongs with the next base bump. - **Tailscale version skew:** `runs/join.py` pins `TAILSCALE_CLIENT_VERSION = "1.98.9"` for the agent client while the router pins `v1.102.2`; whether those two should track each other is a separate call. `ROUTER_PIN` now needs deliberate bumping (renovate rule / runbook note). - **Other `ContractError` subclasses** not listed in the runs router hit the same unhandled-500 path commits 6 and 8 fix; auditing the rest is separate work.


guru-xorciseai:
Approving. The thing that makes this PR is that it doesn't stop at the fix.

The `/proc/<pid>/exe` → Rosetta → moby#47406 chain is a genuinely good find, and the arch × engine × hook table is what turns it from a plausible story into a demonstrated one. Special credit for the line most people would have skipped: *"sibling mode never avoided nesting, it moved container creation onto Docker Desktop's newer daemon."* That reframes the fallback from a workaround into a bug, which is what justifies deleting it.

Three calls I want on the record because they're the ones that'll age well:

- **No on-disk verdict cache.** A persisted negative keyed on a host fingerprint turns one registry blip into a permanent refusal that never clears. In-memory + `doctor`  re-probes fresh is the right shape.
- **Tier 2 as the only gate.** Gating on the Rosetta binfmt registration would have refused every Linux host. Keeping Tier 1 as the *explanation* rather than the *condition* is a subtle, correct distinction.
- **One `build.base_compat` behind both the gate and the badge.** That makes "card says runnable, run refuses" structurally impossible instead of merely currently absent. Same instinct as `test_dind_base_parity.py` pinning the label to `BASE_VERSION`.

Also glad the `DOCKER-FORWARD` rework got chased down rather than assumed benign and that the answer landed as "doesn't touch the production path *because* `tailscale0` is a TUN," with the veth stand-in to prove it, not "seems fine."

One gate before merge: the PR notes `test-heavy` is skipped on PRs. Given this is the runner rewrite, let's get a `full-ci` run green first. Merging on that.

PR #53, opened 2026-08-19
feat(otel): export a run's raw OTLP stream as JSONL (REST + web UI + CLI)

christo-xorciseai:
## What

Adds a raw-evidence export: a run's complete OTLP stream as a JSONL file consumable by external OpenTelemetry tooling, exposed on all three surfaces.

- **REST** — `GET /api/runs/{run_id}/otlp.jsonl`: every trace batch (seq order), then every log batch, one OTLP/JSON envelope per line with **no XORCISE framing** — i.e. the Collector `otlpjson` file format, so the file feeds straight into a Collector and from there Jaeger/Tempo/etc. Payloads are re-serialized compactly so each envelope is guaranteed to occupy exactly one line. Unknown run → 404 (the stores read silently empty otherwise). `Content-Disposition: attachment` with `xorcise-run-<id8>-<mission>-otlp.jsonl`.
- **Web UI** — an "OTLP Traces (.jsonl)" button in the run result page's action row, a plain `<a download>` in the `DownloadReport` idiom. No grading gate: the evidence exists as soon as telemetry does.
- **CLI** — `xorcise run traces --export [--out PATH]`, a thin REST client of the endpoint (default `./xorcise-run-<id8>-otlp.jsonl`, `--out` implies `--export`). `--export --json` and `--export --since` are usage errors — `--json` keeps its role as the seq-framed polling envelope (the `--since` cursor lives in that framing), while `--export` is the Collector-ready snapshot.

## In-progress runs

Exports work mid-run and exit 0: the raw stores are append-only with whole-batch commits, so a partial export is always a valid prefix of the final stream. The CLI checks run state *before* downloading and annotates the success line ("partial snapshot — run still active"), so a run sealing mid-flight can only be over-labeled partial, never under-labeled complete. Once terminal, the record is sealed (ingest rejects late batches) and the export is final.

## Testing

- REST: pure-OTLP-lines shape (spans then logs, envelope round-trip, headers/filename), multiline-payload re-serialization, 404.
- CLI: default filename write, `--out` implies export, partial-snapshot annotation, both flag-conflict guards (assertions ANSI/wrap-normalized).
- Frontend: component tests for the anchor href/encoding/`download` attribute; full suite (706) + `tsc` clean.
- `mypy`, `ruff`, `lint-imports` (4/4 contracts incl. grader-never-reads-projection), topology lane all green. OpenAPI snapshot + generated schema regenerated against this base (no-drift gate).

Note: 8 pre-existing `test_cli_run.py` failures reproduce identically on clean `origin/main` in colorized terminals (ANSI codes in asserted output) — unrelated to this change.

### DIFF
--- a/frontend/openapi.json
+++ b/frontend/openapi.json
@@ -5816,6 +5816,47 @@
         ]
       }
     },
+    "/api/runs/{run_id}/otlp.jsonl": {
+      "get": {
+        "description": "The run's RAW OTLP stream as a downloadable JSONL file (OTLP/JSON lines).\n\nOne line per export batch, verbatim as the agent streamed it: trace batches\n(`resourceSpans`) in seq order, then log batches (`resourceLogs`). Deliberately NO\nXORCISE framing \u2014 no header line, no envelope \u2014 every line is a plain OTLP/JSON\nobject, so the file is directly consumable by OTel tooling that reads the\nCollector's otlpjson file format (and through a Collector, any trace backend).\nWorks mid-run as a partial export of what's been ingested so far. Unknown run \u2192 404\n(the stores yield silently empty reads). Content-Disposition is `attachment`, so a\nbrowser downloads rather than renders it.",
+        "operationId": "run_otlp_jsonl_api_runs__run_id__otlp_jsonl_get",
+        "parameters": [
+          {
+            "in": "path",
+            "name": "run_id",
+            "required": true,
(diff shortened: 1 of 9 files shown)


guru-xorciseai:
lgfm!! i tested on my local. And this feautre is nice! I will merge them to main. thanks @christo-xorciseai

PR #55, opened 2026-08-19
refactor(cli): route `run events export` through the REST server

### LINKED ISSUE #54: CLI cohesion: `run events export` bypasses the REST server
## Problem

An audit of the CLI (all modules under `src/xorcise/core/cli/`) for commands that bypass the REST server found exactly one offender that reads **run evidence** out of the local store in-process:

**`xorcise run events export`** (`src/xorcise/core/cli/commands/run.py`) imports `xorcise.core.rest.events_export.export_run_events` and runs the whole projection rebuild inside the CLI process — SQLite reads via the `events_view` seam, file write under `~/.xorcise/runs/`. It is also an awkward hybrid: it already falls back to `RestClient` for short-id prefix resolution (when the literal id yields an empty export), so it needs a running server for id resolution but bypasses it for the data itself.

This violates the standing direction that all CLI commands route through the REST server. Every other direct-access site found by the audit is justified lifecycle/diagnostic code that must work when the server is down (`up`/`down`/`serve`/`db upgrade`/`doctor`), static package-data reads (`role list`), or CLI plumbing (port discovery, update cache) — none of those can or should be converted.

## Proposal

Mirror the raw-OTLP export shape added in #53:

1. **New endpoint** `GET /runs/{run_id}/events.jsonl` — serves the normalized AgentEvent projection as a downloadable JSONL file (header line + one event per line, same content `export_run_events` writes today), `application/x-ndjson`, `Content-Disposition` attachment, 404 for an unknown run. Rendering refactored out of `events_export` so the on-seal server-side artifact write (`run_terminate.grade_and_record`) keeps working unchanged.
2. **CLI conversion** — `xorcise run events export <run_id> [--out PATH]` becomes a thin client: resolve the id through the server (proper prefix resolution, replacing the empty-export fallback dance), `GET` the JSONL, write it locally.
3. **Default output path moves to the CWD** (`./xorcise-run-<id8>-events.jsonl`), matching `run report` and `run traces --export`. Writing into the server's `~/.xorcise/runs/` tree from the client was only coherent while the CLI and server shared a process boundary; against a remote server it is wrong. The server still materializes its own copy under `~/.xorcise/runs/<run_id>/` on seal.

## Acceptance

- `run events export` performs no store/domain imports; only `RestClient`.
- Endpoint covered by unit tests (content parity with the projection view, one-line-per-event, 404).
- CLI covered by RestClient-stubbed tests; OpenAPI artifacts regenerated (no-drift gate).

christo-xorciseai:
Closes #54.

## What

Converts `xorcise run events export` — the last CLI command reading run evidence out of the local store in-process — into a thin REST client, mirroring the raw-OTLP export shape from #53.

- **New endpoint** `GET /runs/{run_id}/events.jsonl`: the normalized AgentEvent projection as a downloadable JSONL file — a `{"type": "header", ...}` line (run + adapter metadata), then one event per line. `application/x-ndjson`, `Content-Disposition: attachment`, 404 for an unknown run, works mid-run as a partial projection. Follows the plane-isolation invariant (otel display plane lazy-imported inside the handler).
- **`events_export` split**: `render_run_events()` (content) extracted from `export_run_events()` (file write), so the on-seal server-side artifact under `~/.xorcise/runs/<id>/` is produced by exactly the same code path the endpoint serves.
- **CLI conversion**: the command now resolves the id through the server (proper prefix resolution, replacing the old empty-export-then-retry fallback dance), downloads the JSONL, and writes it locally. No store or domain imports remain — only `RestClient`.

## Behavior change

The CLI's default output path moves from the server's home tree (`~/.xorcise/runs/<run_id>/agent-events.jsonl`) to the CWD (`./xorcise-run-<id8>-events.jsonl`), matching `run report` and `run traces --export`. Writing into the server's home from the client was only coherent while the two shared a machine; the server still materializes its own copy under `~/.xorcise/runs/<id>/` on run finalization, so the durable artifact is unaffected. `--out` continues to override.

## Tests

- Endpoint: content parity with the `events_view` projection, header/count integrity, download headers + filename, unknown-run 404.
- CLI: RestClient-stubbed — default-path download, `--out` override, short-id resolution through the server.
- OpenAPI artifacts regenerated (`frontend/openapi.json`, `schema.ts`); the only path delta is the new endpoint.
- Local: mypy (482 files), ruff check + format, lint-imports (4/4 contracts, incl. "Grader never reads the AgentEvent projection"), unit/topology/adapters lanes, frontend typecheck + 706 tests.

### DIFF
--- a/frontend/openapi.json
+++ b/frontend/openapi.json
@@ -5534,6 +5534,47 @@
         ]
       }
     },
+    "/api/runs/{run_id}/events.jsonl": {
+      "get": {
+        "description": "The run's NORMALIZED AgentEvent projection as a downloadable JSONL file.\n\nA `{\"type\": \"header\", ...}` line (run + adapter metadata), then one AgentEvent per\nline \u2014 clean bodies, not raw OTLP. The same content the on-seal artifact under\n`~/.xorcise/runs/<id>/` carries; what `xorcise run events export` fetches. For the\nraw OTLP stream, use GET /runs/{run_id}/otlp.jsonl. Works mid-run as a partial\nprojection of what's been ingested so far. Unknown run \u2192 404. Content-Disposition\nis `attachment`, so a browser downloads rather than renders it.",
+        "operationId": "run_events_jsonl_api_runs__run_id__events_jsonl_get",
+        "parameters": [
+          {
+            "in": "path",
+            "name": "run_id",
+            "required": true,
+            "schema": {
+              "title": "Run Id",
+              "type": "string"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "content": {
+              "application/json": {
+                "schema": {}
+              }
+            },
+            "description": "Successful Response"
+          },
+          "422": {
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/HTTPValidationError"
+                }
+              }
+            },
+            "description": "Validation Error"
+          }
+        },
+        "summary": "Run Events Jsonl",
+        "tags": [
+          "runs"
+        ]
+      }
+    },
     "/api/runs/{run_id}/events/{event_id}/raw": {
       "get": {
         "description": "The source RAW OTLP span(s) an event was derived from (drill-down).\n\n`{event_id:path}` \u2014 AgentEvent ids derive from base64 span ids, which may contain '/'.",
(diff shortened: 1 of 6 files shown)


guru-xorciseai:
okay! i like this additional feature on top of export #53

PR #57, opened 2026-08-20
feat(runner): native-first multi-arch execution (selection, fallback, verification)

christo-xorciseai:
**Stacked on #56** (`feat-mission-versioning`) — this diff shows only the multi-arch half of the contract adoption. Merge #56 first, then retarget this PR to `main`.

The client permanently pinned `linux/amd64`, so an Apple-Silicon host emulated every mission even when the catalog published a validated native arm64 image. This PR turns the pin into a per-mission decision (contract §16/AS1–AS5), made **before any byte moves**:

- **Override first**: `XORCISE_DOCKER_PLATFORM` wins unconditionally; its default becomes `""` = automatic.
- **Pre-contract catalogs are untouched**: no validated-platform list ⇒ no selection ⇒ byte-for-byte the pre-feature behaviour (pinned by test against prod-shaped responses).
- **Native-first**: the daemon's own platform when the mission validated it (three of four test-env missions publish arm64).
- **AMD64 fallback** under the host's emulation layer, with the contract's notice ("Native ARM64 image unavailable…").
- **Typed refusal**: `PlatformUnsupportedError` (a `PullError`) when no execution path exists — raised before any download, naming what the mission does support, mapped **409** on pull, update, and run-create (like the base-generation refusal). `breachpoint` (amd64-only) is the live case.

The selection is passed to Docker explicitly (AS5) and **verified after the pull** (§43-CL6): an image resolving to a different architecture than selected is refused before anything is installed. The install record and run evidence pick the verified platform up unchanged through #56's identity plumbing — no schema change in this PR.

Visibility: `xorcise doctor` reports the daemon's native platform with the PS1 wording (missions may support fewer platforms than the base), and the mission detail page lists the validated platforms.

## Verification

- New `tests/unit/test_platform_selection.py` covers the full decision table (override / no-offer / native / fallback+notice / unknown-host / typed refusal), the driver wiring, the post-pull mismatch refusal, and the pre-contract no-op.
- All lanes + `mypy` + `ruff` + `lint-imports` green (the ~40 local unit failures are the known terminal-colorization artifacts; they pass in CI). Frontend typecheck + 706 tests green; no OpenAPI delta.

---

## UI surfacing (added after the frontend gap audit)

Three follow-up commits make the contract data actionable in the web UI, not just present on the wire:

- **`feat(rest)`: the platform facts the UI cannot compute** — `CatalogEntry.platform`/`emulated` (server-computed against the daemon), installed rows borrow the catalog's `platforms` offer, `SystemInfo.host_platform`, and pull/update responses now return the recorded identity. All computed projections — **0002 remains the stack's only DB migration.**
- **`feat(frontend)`: the update action everywhere its badge appears** — a shared Update button (card, list row, and a new detail-page banner with the version delta), badge parity between card and row, validated-platform tags with PV5/PS1 tooltips, creator-SemVer chip, and an "emulated" flag on non-native installs.
- **`feat(frontend)`: platform honesty at the point of commitment** — the new-run form warns (non-blocking) before pulling/running an emulated or non-native mission, disables Start *with the reason* for what the server would refuse anyway (base mismatch, no executable platform), and stays silent when nothing is known. The results Conditions card shows the executed platform.

Frontend: 98 test files / 721 tests green, typecheck clean; backend lanes at baseline; OpenAPI regenerated with no drift.

### DIFF
--- a/frontend/src/features/missions/mission-card.tsx
+++ b/frontend/src/features/missions/mission-card.tsx
@@ -43,6 +43,11 @@ export type PlatformTag = {
  * one thing a lone "installed" badge could never tell you: WHICH arch you actually have.
  */
 export function platformTags(c: CatalogEntry): PlatformTag[] {
+  // A static mission ships an attachment and no container, so it has no architecture to claim.
+  // The catalog still sends platforms: ["linux/amd64"] for those rows, which would otherwise
+  // paint an AMD64 badge on a mission that never runs an image. Keyed on the artifact as well as
+  // the type, so a tag only ever reflects something that actually exists.
+  if (c.type === "static" || !c.image) return [];
   const all =
     c.platform && !c.platforms.includes(c.platform)
       ? [...c.platforms, c.platform]
--- a/frontend/src/features/missions/mission-card.tsx
+++ b/frontend/src/features/missions/mission-card.tsx
@@ -1,18 +1,20 @@
 "use client";
 
 import Link from "next/link";
-import { Boxes, Download, FileText, FlaskConical, FolderOpen, Loader2 } from "lucide-react";
+import { Boxes, Download, FileText, FlaskConical, FolderOpen, Loader2, RefreshCw } from "lucide-react";
 import { Badge } from "@/components/ui/badge";
 import { Button, buttonVariants } from "@/components/ui/button";
 import { Card } from "@/components/ui/card";
 import { cn } from "@/components/ui/cn";
 import { Progress } from "@/components/ui/progress";
 import type { CatalogEntry } from "@/lib/api/types";
+import { errorDetail } from "@/lib/api/client";
 import {
   formatBytes,
   formatEta,
   pullPhaseLabel,
   usePullMission,
+  useUpdateMission,
   type PullMissionState,
 } from "./queries";
 import { environmentLabel, titleCase } from "./labels";
@@ -20,12 +22,26 @@ import { DifficultyBadge } from "./difficulty-badge";
 
 const MAX_TECH = 3;
 
+/** Compact tag text for a validated platform: "linux/amd64" → "amd64" (the os is noise in a
+ *  catalog that is linux-only by contract; the full value rides the tooltip). */
+export function platformLabel(p: string): string {
+  return p.replace(/^linux\//, "");
+}
+
+const PLATFORMS_TOOLTIP =
+  "Validated platforms — verified by the remote build, never a creator claim. " +
+  "Individual missions may support fewer platforms than the mission-base.";
+
+const EMULATED_TOOLTIP =
+  "This install is not native to this host's architecture; it runs through Docker's " +
+  "emulation layer. Functional, but slower and not validated natively.";
+
 /**
  * The secondary "why" line under the single Update available status (§35/UX1): which of the
  * two versions moved — mission, base, or both. Undefined when the catalog didn't say (the
  * digest alone flagged the update) so the badge still renders, just without a tooltip.
  */
-function updateReason(c: CatalogEntry): string | undefined {
+export function updateReason(c: CatalogEntry): string | undefined {
   const parts: string[] = [];
   if (c.current_mission_version && c.current_mission_version !== c.mission_version) {
     parts.push(`Mission ${c.mission_version ?? "?"} → ${c.current_mission_version}`);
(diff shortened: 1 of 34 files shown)


src/xorcise/core/rest/docker_runtime.py
```diff
@@ -145,6 +147,86 @@ def require_nested_support(
...
+    value: str | None = None
+    try:
+        result = subprocess.run(
+            ["docker", "image", "inspect", "--format", "{{.Os}}/{{.Architecture}}", image],
+            capture_output=True,
+            timeout=5,
+            text=True,
+        )
+        raw = result.stdout.strip()
+        if result.returncode == 0 and "/" in raw:
```
guru-xorciseai: **`"/" in raw` accepts docker's empty-field output as a valid platform.**

On Docker 29 with the containerd snapshotter, `docker image inspect --format '{{.Os}}/{{.Architecture}}'` returns literally `/` — both fields empty, exit code 0 — for a foreign-arch local image. That passes this guard, so `value = "/"`.

Reproduced on an amd64 host (Docker 29.6.2, overlayfs + containerd snapshotter) against `test.api.xorcise.ai`, by tagging a real arm64 image onto a pre-contract install's ref:

```
local_image_platform = '/'   host_platform = 'linux/amd64'   _emulated = True
GET /api/missions  →  guardrail-proxy | platform= '/' | emulated= True
```

`platformLabel("/")` returns `"/"`, so the run-form warning renders as *"This install is **/**, not native AMD64 — the run executes through emulation…"*, and the detail page paints a `/` tag.

Scope is narrow but real: only installs with no recorded platform, i.e. pre-contract installs and `your_own` fuses — a post-contract pull records the platform and that correctly wins. Run provenance is unaffected (it reads `installed.json`). Worth noting it bites hardest on an ARM host that fell back to amd64, which is exactly the case this fallback exists for.

Suggest requiring both halves to be non-empty:

```python
os_, _, arch = raw.partition("/")
if result.returncode == 0 and os_ and arch:
    value = raw
```

`host_platform` at L179 carries the same guard and should get the same treatment.

frontend/src/features/missions/mission-card.tsx
```diff
@@ -1,31 +1,76 @@
...
+};
+
+/**
+ * The platform tags to render for a mission, marking which one is the installed image.
+ *
+ * The installed platform is unioned into the validated list (and de-duplicated) so an install
+ * still names its architecture even against a pre-contract catalog that offers no list — the
+ * one thing a lone "installed" badge could never tell you: WHICH arch you actually have.
+ */
+export function platformTags(c: CatalogEntry): PlatformTag[] {
```
guru-xorciseai, line 45: **Static missions have no architecture — `platformTags` should return nothing for them.**

The live test catalog serves all 7 `type: "static"` missions with `image: null` and `image_size_bytes: null`, but still `platforms: ["linux/amd64"]`. So this function paints an `AMD64` badge on missions that have no container image at all — cipher-beacon, derelict-manifest, molten-perimeter, pipe-dream, rogue-uplink, severed-stream, verify-you-are-human, every one of them attachment-only.

The client degrades correctly everywhere else on those rows (`compatible: null`, `base_major: null`), so the card is the one place the stale catalog field leaks to the user as a claim about a runtime that does not exist.

Cheapest fix is here — return `[]` when the mission is static (or, more defensively, when it has no image), so a platform tag only ever reflects an actual artifact. Getting the catalog to stop emitting `platforms` for static missions is the other half of it, but the UI shouldn't depend on that being fixed first.

src/xorcise/core/rest/docker_runtime.py
christo-xorciseai: Confirmed and fixed in `eec976b`.

Reproduced it here from the opposite direction — arm64 host (Docker 29.6.2, containerd snapshotter) inspecting a freshly pulled amd64 image:

```
$ docker image inspect --format '{{.Os}}/{{.Architecture}}' busybox:1.36
rc=0  raw=/
$ docker image inspect busybox:1.36 --format 'Os={{.Os}} Arch={{.Architecture}}'
Os= Arch=
```

So it is not specific to amd64 hosts or to a re-tagged ref — any foreign-arch local image under that snapshotter does it.

Took your suggestion, factored into a shared `_both_halves()` so the host probe at L179 and the image probe get the identical guard rather than two copies that can drift. Unknown now stays `None`, which the callers already render as no claim.

Covered by a parametrised test over `linux/amd64`, `/`, `linux/`, `/amd64`, `""` on both probes — `linux/` and `/amd64` because a half-empty field is the same class of bug and `"/" in raw` would have accepted those too.

Agreed the scope is narrow, and worth restating for anyone reading later: only installs with no recorded platform, and run provenance was never affected since it reads `installed.json`.

frontend/src/features/missions/mission-card.tsx
christo-xorciseai: Fixed in `eec976b`. Confirmed against the live catalog — all 7 static missions come back with `image: null` and `platforms: ["linux/amd64"]`, exactly as you describe.

Took the defensive form you suggested, gating on the artifact as well as the type:

```ts
if (c.type === "static" || !c.image) return [];
```

Before relying on `!c.image` I checked it does not also suppress tags while browsing, since that is where platforms matter most. It tracks **type**, not install state:

```
lab     n=21  with image=21  installed=15
static  n=7   with image=0   installed=7
```

All six uninstalled lab missions still carry an image and still render their tags.

That did surface one thing worth flagging: three detail-page tests broke, because the `BASE` fixture described an installed lab mission with **no** image — a shape that cannot occur. The guard was right and the fixture was wrong, so it now carries the image it always implied. If it had been the other way round I would have come back to you rather than widen the guard.

Agreed the catalog should stop emitting `platforms` for static missions; not addressed here, and the UI no longer depends on it.

Issue #62, opened 2026-08-20
doctor: control-plane check verifies the container, not the address routers dial — reports healthy while every run fails


christo-xorciseai:
`xorcise doctor` reports the control plane healthy by checking the **container**, not the address
routers actually dial. When the two disagree, doctor is green and every run fails.

## What happens

`doctor` probes `control_plane(settings.headscale_container)` — a container-level health check. It
never verifies that `headscale_url`, the address the run's router logs in to, is reachable.

The two diverge whenever the host's address changes after `up` provisioned Headscale — a laptop
moving networks, or a DHCP lease change is enough. Headscale stays published on the old address,
the container stays perfectly healthy, and `doctor` says:

```
Services
  ✓ port 3001 (rest): in use by the running XORCISE service
  ✓ port 4318 (otlp): in use by the running XORCISE service
  ✓ run network ready (Headscale)

No problems found.
```

Meanwhile the control plane is unreachable even from the host itself.

## Why it is worth fixing

The resulting failure points somewhere else entirely. The router comes up, cannot log in, and sits
at `NeedsLogin`; it never gets a tailnet address; the readiness window expires and the run ends as:

```
readiness: closing out <run> — not ready within the readiness window
```

Nothing in that mentions the control plane, so it reads as "this mission will not deploy". The
mission stack is fine — every service starts and stays up. Confirming that the control plane is
the actual cause means going into the fused container, finding the router, and reading its
tailscaled log, which is a long way to travel for a stale IP in `config.toml`.

`doctor` exists precisely to shorten that path, and here it actively points away from the fault.

## Suggested fix

Probe `headscale_url` for reachability rather than (or in addition to) container health, and when
it fails say what was tried and what to do — the remediation is just `xorcise down && xorcise up`,
which re-detects the host address and re-provisions. Worth also mentioning the control plane in the
readiness-timeout message when the router never obtained an address, since that is the point where
someone is actually looking.

## Reproducing

1. `xorcise up` on a host, note the address recorded in `~/.xorcise/config.toml`
2. Change the host's address (move networks, or renew the lease)
3. `xorcise doctor` → `No problems found`
4. Create any run → `not ready within the readiness window`

PR #63, opened 2026-08-20
feat(net): make the tailnet the only path between agent and mission

### LINKED ISSUE #61: Per-run networking: the mission cannot reach the agent, and mission networks are not confined to the run
Two related gaps in the per-run network model. They are separate problems, but they share a
cause — the run's tailnet is currently the *intended* path between agent and mission rather than
the *only* one — and fixing either alone leaves the other worse, so they are filed together.

## 1. The mission cannot open connections back to the agent

The per-run ACL only ever grants the agent → mission direction. There is no rule for the reverse,
so a connection initiated by a mission service toward the agent is dropped by the agent node's
packet filter, and nothing in the run is configured to carry one.

That makes an entire mission shape unbuildable: any mission where the target calls the agent
rather than the other way round. Three missions in the catalog are written this way and cannot
be completed as authored:

- `border-control` — registers a callback API that the target POSTs to in bursts
- `guardrail-proxy` — the compliance checker drives requests through an agent-hosted proxy
- `vanishing-point` — the endpoint checker exercises an agent-hosted REST API

It also rules out reverse-shell and beacon-style payloads generally, which is a limitation on
what missions can be authored rather than a deliberate design choice.

The failure mode is unhelpful: the connection simply times out, with no signal distinguishing
"not permitted" from "nothing listening" from "wrong address". An agent that hits this has no way
to tell it is facing a harness limitation rather than its own mistake, and the time is spent
looking in the wrong place.

## 2. Mission networks are not confined to the run

The per-run compose networks are created without `internal: true`, so they are not restricted to
the run — they retain connectivity beyond it.

Two consequences:

**Mission integrity.** A mission whose premise depends on the environment being closed cannot rely
on it. `border-control` is the clearest case: the whole exercise is that an offline database beats
an online lookup, and its flag literally says so — but nothing in the environment enforces the
constraint the challenge is built on. Any mission making a similar assumption inherits the same
weakness silently.

**Isolation.** The per-run tailnet, the Headscale ACL and the subnet carving exist to make the run
a bounded environment. Leaving the networks unconfined means that boundary is weaker than the rest
of the design assumes.

Worth noting the two interact: because of (2), the callback shape in (1) can appear to work over a
path that was never meant to carry it, so closing (2) without addressing (1) would turn a quiet
integrity problem into a visible functional failure.

## Notes for whoever picks this up

- Confinement has to cover **every** network a mission declares, not just the agent-facing ones.
  Some missions attach a service to more than one network, and confining only the entry network
  leaves those services unconfined — and they tend to be exactly the agent-facing ones.
- The run's router still needs its own way out for the control plane, so it cannot simply be
  confined alongside the mission services.
- Whether the agent is reachable, and whether a mission may reach beyond its run, are separate
  decisions and want separate switches.

A fix for both is in flight.

christo-xorciseai:
Closes #61.

Makes the run's tailnet the only path between agent and mission, in both directions.

Two problems, one cause. The ACL only ever granted agent → mission, so a mission could not open
a connection back — which made callback missions unwinnable. And mission networks were not
confined to the run, which left a non-tailnet path between them that made the first problem
invisible: callbacks "worked" over a path nothing intended. Fixing either alone makes the other
worse, so both are here.

## What changed

**ACL** — the fence renders an inbound rule per run, sourced only from that run's router tag.
Ports are wildcard deliberately: the agent is a host on the mission network, and what may be
reached on it is the mission's firewall to decide, not a policy the harness imposes invisibly. A
single allowed port is a complete tunnel anyway, so scoping buys nothing against a capable agent
while ruling out reverse shells, beacons and dynamic ports for authors. `assert_policy_safe` pins
the rule's source and destination to the run and fails closed without a router tag.

**Router** — `MASQUERADE` on `tailscale0` plus a watcher that discovers the agent by its Headscale
user and points an ingress DNAT at it. LAN → tailnet is not a direction subnet routers support by
default; without the SNAT, packets leave with a source the node does not own and are discarded
with no diagnostic. Nothing is pinned to an address: the agent joins after the stack is up and may
rejoin with a different one, so the watcher re-points on change.

**Callback address** — lives inside the mission subnet rather than being the agent's tailnet
address. A mission container reaches it as a same-segment neighbour, so no route off the subnet is
needed and the fused container is untouched. It also stops the agent's tailnet identity leaking
into the mission, which reads better in evidence.

**Confinement** — every network a mission declares is `internal: true`, with a router-only egress
network. This closes the non-tailnet path for all join modes, not just the sidecar, and closes the
unrestricted mission egress that let a mission reach beyond its run. Note this must cover *every*
declared network: a service with one foot in an entry network and another in an unconfined one
keeps its way out, and those tend to be the agent-facing services.

Two follow-on changes fell out of it:

- The sidecar's SOCKS/HTTP proxies bind its own bridge address instead of `0.0.0.0`. In
  userspace mode netstack delivers inbound tailnet connections to `127.0.0.1`, so a wildcard bind
  publishes the agent's own proxy on the tailnet once ingress is allowed.
- On Linux, `auto` now prefers kernel mode, then a Docker sidecar, and only falls back to
  userspace on the host. That last mode runs netstack in the host namespace, where the same
  behaviour exposes every host loopback service to the mission network. It is now last-resort and
  says so on join. Its proxies bind `127.0.0.2`, which is still loopback-only but is not
  netstack's forward target.

## One switch, not two

The agent is **always** reachable from the mission network. That is not a per-mission setting: a
host on a network is reachable, and making the agent unreachable is the artificial state — it is
what made callback missions unwinnable and reverse shells impossible to author. It is also a
post-exploitation capability rather than an entry one, since mission services never dial the agent
unaided, so it cannot shortcut the front door of any mission. What may be reached on the agent is
the mission's own firewall to decide.

Mission networks are the mirror image, and they **do** get a switch. `allow_egress` is an escape
hatch FROM a restriction: without it a mission that legitimately needs external access is
unbuildable, and someone works around the confinement instead. It defaults off, because a mission
whose premise is "no external access here" must not be undercut by access nobody intended.

The whole catalog was checked for collisions with the reserved router and callback addresses
first — none pin them.

## Verification

Beyond unit coverage, exercised against live runs on images pulled from the registry:

- callback flow end to end — 100 callbacks delivered over the tailnet
- causal check — the callback dies when the tailnet is taken down and recovers when it returns,
  with no non-tailnet path to fall back to
- counters — DNAT hits and WireGuard packet deltas match the callback count
- `segmented-pivot` and `operation-tessera`, the two multi-network missions: all networks
  confined, egress blocked from every multi-homed service, and their pivots and segmentation
  intact (cross-segment resolution still correctly fails; same-segment still works)
- a mission that declares neither switch, to confirm the defaults reach it and normal outbound
  operation is unaffected

## Reviewer notes

- There is no per-run persistence for reachability and no migration for it. An earlier revision
  of this branch carried both, to survive the ACL being re-rendered from the database; once
  reachability stopped varying they were dead weight and were removed while the migration was
  still unshipped and free to delete.
- Confinement needs the full network list, which only the mission's compose has — read from the
  image at deploy. Resolving it at ingestion would fail open for already-installed missions.
- No schema change: this PR adds no migration.
- One behaviour change worth flagging: pwn missions gain reverse shells as an option. The graded
  difficulty is unchanged — code execution is still the hard part — but runs before and after are
  not strictly comparable.
- Residual, deliberately not fixed here: a Linux agent host with neither a TUN nor Docker still
  runs netstack in the host namespace. Narrow after the mode-selection change, and self-announcing.
- #62 is related but separate — `doctor` reports the control plane healthy while it is unreachable,
  which is not addressed by this PR.


christo-xorciseai:
Pushed `bd64007` — removed the `agent_ingress` switch entirely. Net **−112 lines**, and this PR no
longer carries a schema change.

### Why

It shipped defaulting to on and no mission ever set it off, so it was plumbing for a decision
nobody was making: a database column, a migration, and the value threaded through the manifest
contract, `NetworkSpec`, `RunNetwork`, the fence port and its client, the repository and the ACL
provider — fourteen files.

Most of that weight was persistence, and persistence existed for exactly one reason: the ACL is
re-rendered from the database, so a restart would drop the inbound rule unless the value survived
there. That only matters if the value can vary. It cannot, so `active_run_networks()` goes back to
two-tuples, `reserve_run` and `create_run_network` lose a parameter, and the test churn those
signature changes caused reverts with them.

The safety invariant gets easier to hold onto too — from *"every run gets an inbound rule if its
persisted flag says so"* to *"every run gets exactly one inbound rule, sourced from its own router
tag"*.

### Why `allow_egress` stays

The asymmetry is the point. `allow_egress` is an escape hatch **from** a restriction: without it a
mission that legitimately needs external access is unbuildable, and someone works around the
confinement instead — worse than the hatch. `agent_ingress` was a restriction nobody asked for, and
without it nothing is blocked; a mission simply has a capability it does not exercise. That is also
why `allow_egress` needs no column: it is consumed at deploy time, not re-derived by the ACL
provider.

### Why now

Migration `0003` had not shipped. Today it was a clean delete; after release it would have needed a
second migration and care about existing rows. If a mission ever needs a genuinely one-way network,
re-adding this costs exactly what it cost here, and would be designed against a real requirement
rather than an imagined one.

### Verified

Gates green (mypy, ruff check + format, import walls, single-distribution guard) and all three fast
lanes pass. Smoke-tested on a live run: the inbound ACL rule still renders with no flag anywhere in
the system, mission networks stay confined, and the router still arms the callback address —
`10.200.1.254` on `eth1`, alongside its pinned `10.200.1.2`, with the egress network on `eth0`.

src/xorcise/core/runner/netoverride.py
```diff
@@ -104,14 +273,21 @@ def build_net_override(
         f"TS_HOSTNAME={project}-router",
         "TS_STATE_DIR=/var/lib/tailscale",
     ]
+    router_nets: dict[str, object] = {
+        name: {"ipv4_address": _host_in_subnet(cidr, ROUTER_OCTET)}
```
guru-xorciseai: **Blocker — `ROUTER_OCTET = 2` collides with Docker's first dynamic address.**

The gateway is `.1`, so `.2` is the *first* address Docker's dynamic IPAM hands out. Pinning the
router there puts it in a race with every unpinned mission service on the same network.

Reproduced against real Docker/Compose: a stack with two unpinned services plus a router pinned to
`.2` on a `/24` failed `docker compose up -d` in **3 of 5 runs** with
`failed to set up container networking: Address already in use`. The fused entrypoint is `set -eu`
and runs `docker compose ... up -d`, so the outer container dies and the run reports FAILED — with
an error that names neither the router nor the colliding service.

`_assert_reserved_addresses_free` can't catch this: it only walks `static_ips`, i.e. *statically
pinned* services. The colliding service is by definition unpinned.

This is new behaviour — before this PR the router took a dynamic address and never contended.

Suggest moving the router to the top of the range next to ingress (`.253` against
`ingress_address`'s `.254`), which is where the reservation check is already looking.

src/xorcise/core/runner/netoverride.py
```diff
@@ -125,11 +301,31 @@ def build_net_override(
...
+            'printf "%s" "$$XORCISE_HEADSCALE_CA_B64" | base64 -d > /tmp/headscale-ca.pem'
+        )
+    if entry_subnets:
+        if not agent_user:
+            raise ValueError(
+                "agent ingress requires agent_user — the router discovers the agent's tailnet "
+                "address by its headscale user and cannot arm ingress without it"
+            )
+        # One ingress endpoint, on the first carved subnet by name so the address is stable.
+        net_name = sorted(entry_subnets)[0]
```
guru-xorciseai: **Blocker — ingress is armed on one subnet, but `internal: true` leaves the others with no route
to it.**

`net_name = sorted(entry_subnets)[0]` arms exactly one callback address, and the confinement change
in this same PR marks every mission network `internal: true`. A Docker `--internal` network gives
its containers an on-link subnet route and **no default route**, so an address in a sibling subnet
is simply unreachable:

```
# container on 10.211.7.128/25, pinging a sibling internal subnet
ping: sendto: Network unreachable
```

For a two-entry mission (`zulu_net`→10.200.1.0/25, `alpha_net`→10.200.1.128/25) this picks
`alpha_net` and arms 10.200.1.254. Anything on `zulu_net` that dials the address the prompt handed
the agent gets `Network unreachable` — and which network wins is decided by **alphabetical name**,
not by which one is agent-facing.

This is the exact shape of `segmented-pivot` and `operation-tessera`. I suspect the verification
covered confinement and segmentation but not a callback *originating from a non-first segment* —
worth re-testing that specific case before anything else here.

Either arm ingress on every entry network, or take the agent-facing network from the manifest
rather than from string collation.

src/xorcise/core/runner/netoverride.py
```diff
@@ -61,6 +91,125 @@ def target_ips_for(
...
+#
+# The agent is found by its HEADSCALE USER, which is derived from the run id and therefore known
+# here; its tailnet ADDRESS is not knowable at compose time (the agent joins later, and may rejoin
+# with a different one), so the DNAT target is discovered and re-pointed rather than pinned.
+_INGRESS_PROLOGUE = r"""
+rip="%%ROUTER_IP%%"
+dev=$$(ip -4 -o addr show 2>/dev/null | awk -v r="$$rip" '$$4 ~ "^"r"/" {print $$2; exit}')
+[ -n "$$dev" ] || dev=eth0
+ip addr add %%INGRESS_CIDR%% dev "$$dev" 2>/dev/null || true
+iptables -t nat -C POSTROUTING -o tailscale0 -j MASQUERADE 2>/dev/null ||
```
guru-xorciseai: **High — the prologue shells out to bare `iptables`, reintroducing the exact dependency this file
pins away from 30 lines below.**

```
$ docker run --rm --entrypoint sh tailscale/tailscale:stable -c 'ls -l $(command -v iptables)'
/usr/sbin/iptables -> iptables-legacy   # -> xtables-legacy-multi
```

The `TS_DEBUG_FIREWALL_MODE=nftables` comment below already spells out why that matters: legacy
iptables needs `ip_tables`/`iptable_nat`, this container is nested with no `/lib/modules` to
modprobe them, and on nftables hosts (Kali) it therefore installs **zero** rules.

The prologue runs under `/bin/sh -c` with no `set -e`, so on exactly those hosts the `MASQUERADE`
and `DNAT` commands error to stderr and boot continues straight to `exec containerboot`. No SNAT,
no DNAT, every callback times out — while the control plane looks healthy. Same silent-data-plane
failure the pinned firewall mode was added to prevent.

`iptables-nft -t nat ...` (or `nft` directly), plus a hard failure if the rules don't install.

src/xorcise/core/runner/netoverride.py
```diff
@@ -61,6 +91,125 @@ def target_ips_for(
...
+  iptables -t nat -A POSTROUTING -o tailscale0 -j MASQUERADE
+(
+  cur=""
+  waited=0
+  while :; do
+    aip=$$(tailscale status 2>/dev/null |
+      grep -E "[[:space:]]%%AGENT_USER%%[[:space:]]" | head -1 | awk '{print $$1}')
```
guru-xorciseai: **Medium — `head -1` with no online filter means a stale registration wins permanently.**

`tailscale status` lists offline peers too. If the join script runs twice for a run — a re-issued
tool call, a re-join after a crash — Headscale holds two nodes for the same user and both rows
match the `grep`. `head -1` then deterministically returns the same row, so `[ "$aip" != "$cur" ]`
never fires and the DNAT stays pointed at the dead address.

That's precisely the case the header comment above claims to handle ("may rejoin with a different
one, so the watcher re-points on change"). The failure is silent in the worst way: the watcher has
already logged `agent ingress <ip> -> <addr>`, so the log says armed while callbacks black-hole.

Filter to online peers before `head -1` (`tailscale status --json` + `jq` on `.Peer[].Online`
would be exact, if `jq` is available in the image).

src/xorcise/core/runner/netoverride.py
```diff
@@ -61,6 +91,125 @@ def target_ips_for(
...
+    case "$$aip" in 100.*) ;; *) aip="" ;; esac
+    if [ -n "$$aip" ] && [ "$$aip" != "$$cur" ]; then
+      if [ -n "$$cur" ]; then
+        iptables -t nat -D PREROUTING -d %%INGRESS_IP%% -p tcp \
+          -j DNAT --to-destination "$$cur" 2>/dev/null || true
+      fi
+      iptables -t nat -A PREROUTING -d %%INGRESS_IP%% -p tcp -j DNAT --to-destination "$$aip"
```
guru-xorciseai: **Medium — the DNAT is `-p tcp` only, but the prompt promises the agent any port.**

`prompt.py` tells the agent to register `http://<addr>:<your-port>/` — "any port you like". A
mission that beacons over UDP (DNS-style exfil, a UDP heartbeat) or a check that pings the callback
address sends to the ingress IP, which the router *has* on its interface but has no DNAT rule for.
The packets are delivered locally to the router and dropped.

From the author's side this is indistinguishable from a mission bug. Either add `-p udp` (and let
ICMP through), or drop the protocol match entirely.

src/xorcise/core/runner/netoverride.py
```diff
@@ -61,6 +91,125 @@ def target_ips_for(
...
+    try:
+        doc = yaml.safe_load(compose_text) or {}
+    except yaml.YAMLError as exc:
+        raise ValueError(f"mission compose is not valid YAML: {exc}") from exc
+    if not isinstance(doc, dict):
+        raise ValueError("mission compose did not parse to a mapping")
+    names: set[str] = set()
+    declared = doc.get("networks")
+    if isinstance(declared, dict):
+        names.update(str(k) for k in declared)
```
guru-xorciseai, line 234: **Medium — PyYAML parses this as YAML 1.1; Docker Compose parses it as YAML 1.2. Boolean-looking
network names get mangled and the real network is left unconfined.**

On this branch:

```python
>>> compose_network_names('networks:\n  no: {}\n  on: {}\nservices:\n  a: {networks: [no]}')
('False', 'True')
```

PyYAML's 1.1 resolver turns `no`/`on`/`yes`/`off`/`y`/`n` keys into booleans and `str(k)` then
stringifies them. Compose uses `gopkg.in/yaml.v3` (YAML 1.2), where they stay the strings `no` and
`on`. So the override declares two phantom networks named `False`/`True` and never sets
`internal: true` on the networks that actually exist — a silent fail-open in the confinement
control.

`yaml.load(text, Loader=yaml.BaseLoader)` (every scalar a string) makes this read the same document
Docker will.

src/xorcise/core/runner/netoverride.py
```diff
@@ -61,6 +91,125 @@ def target_ips_for(
...
+            cidr = entry_subnets.get(net)
+            if cidr is None:
+                continue
+            claimed = _host_in_subnet(cidr, octet)
+            if octet == ROUTER_OCTET:
+                raise ValueError(
+                    f"mission service {service!r} pins {claimed} on network {net!r}, which is "
+                    f"reserved for the run's tailscale router (octet {ROUTER_OCTET})"
+                )
+            if claimed == ingress_address(cidr):
```
guru-xorciseai, line 300: **Low/Medium — this rejects a top-of-range pin on *every* entry network, but ingress only exists on
one of them.**

Given `entry_networks = ('zulu_net', 'alpha_net')`, `build_net_override` arms ingress only on
`alpha_net`. A mission that pins a service at the top of `zulu_net` still hits
`claimed == ingress_address(cidr)` here and the build fails with
`... reserved for the agent's callback address` — naming a reservation that doesn't exist on that
network. The mission is undeployable for no reason.

The mirror image of the anchoring issue at line 314: the check assumes ingress is everywhere, the
builder puts it in one place. Fixing 314 by arming every entry network would make this correct as
written.

src/xorcise/core/runs/join.py
```diff
@@ -55,16 +55,35 @@
...
 case "$TS_MODE" in
   auto)
-    if [ "$(uname -s)" = "Darwin" ]; then TS_MODE=sidecar; else TS_MODE=native; fi
+    if [ "$(uname -s)" = "Darwin" ]; then
+      TS_MODE=sidecar
+    elif [ "$(id -u)" = "0" ] && [ -w /dev/net/tun ]; then
+      # Kernel mode: a real TUN, so the host stack applies. Loopback stays private, targets are
+      # reachable directly by IP, and no proxy exists to expose. The faithful option; preferred.
+      TS_MODE=native
+    elif command -v docker >/dev/null 2>&1 && docker info >/dev/null 2>&1; then
```
guru-xorciseai: **High — `auto` now prefers the sidecar on Linux, and the sidecar path has no fallback.**

A non-root Linux host with Docker but no `/dev/net/tun` used to land on `TS_MODE=native` and pull
the pinned client from `$RC_BASE/tailscale.tgz` — the script's own documented air-gapped path ("the
server caches it, no public egress needed").

It now lands on `sidecar`, which needs `tailscale/tailscale:v1.98.9` **from Docker Hub**.
Run-control serves the tarball but not the image. On an offline host `docker run` fails, the script
does `echo "tailnet sidecar failed to start"; exit 1`, and there's no fall-through to native — so
the agent can't join at all. That's a regression for a case the script explicitly supports.

Falling back to native on sidecar-start failure would keep the security win (sidecar preferred
whenever it's actually available) without breaking the offline path.

Minor, same block: `docker info` now runs twice per join — once for detection here, once for the
guard below. On an unreachable daemon that's two full timeouts on every join.

src/xorcise/core/runs/join.py
```diff
@@ -116,7 +142,17 @@
...
+    # Bind the proxies to this container OWN bridge address, not 0.0.0.0. In userspace-networking
+    # mode netstack delivers inbound tailnet connections to 127.0.0.1 INSIDE this namespace, so a
+    # wildcard bind publishes the agent SOCKS/HTTP proxy ON the tailnet — reachable by the mission
+    # once agent ingress is allowed, which would hand mission code a relay carrying the agent
+    # identity. The docker published-port proxy dials the container address, so binding it keeps
+    # the host side working while removing the tailnet side. (No apostrophes: this whole script is
+    # a single-quoted argument to sh -c.)
+    bip=$(ip -4 -o addr show eth0 2>/dev/null |
+      sed -n "s|.*inet \([0-9.][0-9.]*\)/.*|\1|p" | head -1)
+    [ -n "$bip" ] || bip=0.0.0.0
```
guru-xorciseai: **High — the `0.0.0.0` fallback silently restores the exposure this change exists to remove.**

```sh
[ -n "$bip" ] || bip=0.0.0.0
```

If `ip` isn't present, the interface isn't named `eth0` (custom bridge setups), or the address
lookup is raced at container start, this binds SOCKS5/HTTP on the wildcard — exactly what the
comment directly above explains must not happen. In userspace mode netstack delivers inbound
tailnet connections to `127.0.0.1` inside the namespace, so with the new agent-ingress rule now
open, the mission gets a working SOCKS5/HTTP relay carrying the agent's tailnet identity.

The join prints success either way, so nobody finds out.

For a security control the safe default is fail-closed: `exit 1` with the reason, rather than
falling back to the insecure bind.

src/xorcise/core/runner/service.py
```diff
@@ -102,12 +109,21 @@ def _deploy_env(self, run_id: RunId, network: NetworkSpec) -> tuple[tuple[str, s
         entry_subnets = dict(zip(network.entry_networks, network.routes, strict=True))
         routes = network.routes
         ca_path = _CA_PATH if network.ca_cert else ""
+        # Every network the mission's compose will create, read from the image itself — the
+        # only authoritative source (the installed bundle carries the manifest, not the compose).
+        # Confining just the carved entry networks would leave a multi-homed service its route off
+        # box, so this is what makes the tailnet the ONLY path to the agent.
+        compose_text = self.driver.read_image_file(image, MISSION_COMPOSE_PATH) if image else None
```
guru-xorciseai: **High — a `None` here silently disables confinement, against the explicit warning in
`read_image_file`'s own docstring.**

`DockerDriver.read_image_file` is a non-abstract base method returning `None`, and its docstring
says:

> None is the stub's answer ..., never a real driver's way of reporting failure, because a silent
> empty answer would leave networks unconfined.

But nothing here enforces that. `compose_text = ... if image else None`, then
`all_networks = compose_network_names(compose_text) if compose_text else ()`. So `None` — or a
driver subclass that simply doesn't override the method, or `_deploy_env`'s own `image: str = ""`
default — yields `all_networks = ()`, and only the carved entry networks get `internal: true`.

A multi-homed mission service then keeps its route off box, which is the exact hole the PR's own
description calls out ("a service with one foot in an entry network and another in an unconfined
one keeps its way out, and those tend to be the agent-facing services"). Nothing logs, nothing
raises.

Raise when a real driver returns `None`, and make the base method `NotImplementedError` rather than
a `None`-returning default.

src/xorcise/core/runner/service.py
```diff
@@ -39,6 +39,11 @@
 _RUN_ID_RE = re.compile(r"[0-9a-f]{32}")
 
 
+# Where the fused image keeps the mission's authored compose. Read at deploy to enumerate the
+# networks the run will create so every one of them can be confined.
+MISSION_COMPOSE_PATH = "/mission/docker-compose.yml"
```
guru-xorciseai: **Low — hardcodes `docker-compose.yml` while the manifest already carries the authoritative name.**

`build.py` and `preflight.py` both honour `manifest.environment.compose_file` (default
`docker-compose.yml`, but freely settable), and `_write_fused_dockerfile` does `COPY bundle
/mission` — so the compose can legitimately land at `/mission/compose.yaml`.

`read_image_file` then calls `container.get_archive("/mission/docker-compose.yml")`, which raises
`docker.errors.NotFound`. Only `ImageNotFound` on `create` is translated, so deploy 500s with a raw
docker error instead of a domain error.

Worth threading `environment.compose_file` through rather than adding a third hardcoded copy of the
path (the mission-base entrypoint has the other).

src/xorcise/core/headscale/policy.py
```diff
@@ -110,3 +126,33 @@ def assert_policy_safe(
...
             if dst not in allowed:
                 raise ValueError(f"Policy contains foreign dst {dst!r} for {net.agent_user}@")
+
+    # Inbound rules. The invariant is NOT "no wildcard port" — ports are deliberately wildcard
+    # (see render_policy). It is that every router-sourced rule names THIS run's router tag as its
+    # only source and THIS run's agent user as its only destination, so a run's router can never
+    # be handed a path to another run's agent.
+    ingress_dsts = {f"{n.agent_user}@:*" for n in networks}
+    for net in networks:
+        if not router_tag:
```
guru-xorciseai: **High — the fail-closed check is loop-invariant but sits inside the loop, so an empty `networks`
skips the whole gate.**

`assert_policy_safe(policy_text, [], router_tag="")` runs `for net in networks:` zero times, so it
never reaches `if not router_tag: raise`, and then the `if router_tag:` foreign-dst sweep below is
skipped too. The result is a policy certified as safe without any router-sourced rule having been
checked at all.

The `router_tag: str = ""` default compounds it — any future caller can opt out of a gate the
docstring calls fail-closed just by not passing the argument. Three existing tests
(`test_safe_rejects_missing_cidr`, `test_assert_policy_safe_rejects_a_foreign_dst`,
`test_assert_policy_safe_raises_when_agent_rule_missing_from_acls`) still call it without the tag
and now pass for reasons that depend on check ordering rather than on intent.

Hoist the check above the loop and make `router_tag` a required keyword argument.

src/xorcise/core/runner/docker/driver.py
```diff
@@ -165,6 +165,41 @@ def image_platform(self, image: str) -> str | None:
...
+        # translates into "…(re)build it". Same error, same remediation, whichever hits first.
+        try:
+            container = self._client.containers.create(image, entrypoint=["/bin/true"], command=[])
+        except docker.errors.ImageNotFound as exc:
+            raise ImageNotInstalledError(f"image {image!r} is not in the local store") from exc
+        try:
+            stream, _stat = container.get_archive(path)
+            blob = b"".join(stream)
+        finally:
+            container.remove(force=True)
```
guru-xorciseai: **Medium — this leaks one anonymous volume per deploy, and an unlabelled container if the process
dies mid-call.**

```
$ docker image inspect docker:29.7.1-dind --format '{{json .Config.Volumes}}'
{"/var/lib/docker":{}}
```

Every fused mission image is `FROM xorcise/mission-base` → `FROM docker:29.7.1-dind`, so
`containers.create(image, ...)` allocates an anonymous volume. `container.remove(force=True)`
doesn't pass `v=True`, so it's orphaned on **every** deploy. It carries no
`com.docker.compose.project` label, so `_remove_run_resources`'s volume sweep never reaps it.

The container is also created without `MANAGED_LABEL`/`RUN_ID_LABEL`, so if the server is killed
between `create` and `remove`, `reap_managed` can't find it either.

Two one-liners: `remove(force=True, v=True)`, and add the managed labels at `create`.

guru-xorciseai (commented):
Reviewed the branch at `001af9b`. The thesis is right and the writeup is unusually clear — "two problems, one cause" is the correct diagnosis, and treating agent reachability as a property rather than a switch is the right call. My comments are all about execution, not direction.

Worth stating up front: **`ruff`, `ruff format --check`, `mypy` (488 files) and the unit lane (2107 passed) are all green on this branch, and none of the below is caught by any of them.** Several were checked against real Docker/Compose and the actual `tailscale/tailscale` image rather than by reading.

**Two look startup-breaking:**

- `ROUTER_OCTET = 2` races Docker's first dynamic address — `compose up` failed 3 of 5 live runs with `Address already in use` on a stack with unpinned services. `_assert_reserved_addresses_free` structurally can't see it, since it only walks `static_ips`.
- Ingress is armed on `sorted(entry_subnets)[0]` only, but `internal: true` leaves containers with no default route — so any service on a *different* segment gets `Network unreachable` dialling the address the prompt handed the agent. This is the `segmented-pivot` / `operation-tessera` shape; I think the verification covered confinement and segmentation but not a callback originating from a non-first segment.

**Three are fail-open paths in the new controls** — each one silently reopens the hole this PR closes: the `bip=0.0.0.0` fallback in `join.py`, `None` from `read_image_file` yielding `all_networks = ()`, and the loop-invariant `router_tag` check in `policy.py` that no-ops entirely on an empty `networks`. That pattern repeating in three subsystems feels worth a deliberate pass: for a confinement control, refusing is nearly always better than quietly doing nothing, and two of the three docstrings already say so.

**One is a regression for offline hosts:** Linux `auto` now prefers the sidecar, which needs `tailscale/tailscale:v1.98.9` from Docker Hub. Run-control serves the tarball but not the image, so the air-gapped path the script itself advertises is dead, with no fall-through to native.

The rest (bare `iptables`, stale-peer `head -1`, PyYAML 1.1 vs Compose's YAML 1.2, the leaked anonymous volume, TCP-only DNAT, `agent_ingress_addr` on every lab run, the over-strict reserved-address check, the duplicate step 2) are real but none block on their own.

Happy to take any of these on directly if that's easier than round-tripping.

src/xorcise/core/rest/run_create.py
```diff
@@ -756,6 +779,10 @@ def _create_run_with_cidr(
         # Size the prompt's disclosed-intel count off the SAME policy filter the run-control gate
         # applies, so the advert never promises intel this run may not disclose.
         intel_available=len(allowed_intel(intel_policy, installed.manifest.intel)),
+        # Callback missions: the mission-network address the router forwards to the agent.
+        # Derived from the SAME carved subnet the override pins the router on, so the
+        # prompt and the deployed DNAT can never name different addresses.
+        agent_ingress_addr=_ingress_addr_for(entry_subnets) if entry_subnets else "",
```
guru-xorciseai, line 803: **Medium — this is populated for every lab run, so every prompt claims the mission calls back.**

`entry_networks = tuple(installed.environment.entry_networks) or ("default",)` is never empty for a
lab mission, so `entry_subnets` is always truthy and the ternary always yields an address.

`contracts/connect.py` documents the field as "Callback missions only ... Empty when the mission
never calls back", and `prompt.py` gates the whole callback paragraph on truthiness alone. So a
plain web-exploitation mission now tells the agent it will be called back, and instructs it to
stand up a server and register an address nothing will ever dial — wasted budget and a false
premise in the brief.

Needs an actual signal for "this mission calls back" (a manifest flag, or the presence of a check
that consumes a callback) rather than "this mission has networks".

src/xorcise/core/runs/prompt.py
```diff
@@ -130,6 +132,24 @@ def _join_lines(mission: MissionPrompt) -> list[str]:
...
         "in pwntools/PySocks rather than connecting directly.",
     ]
+    if mission.agent_ingress_addr:
+        # Callback missions. Two things the agent cannot discover: the address to register (its
+        # own tailnet IP is NOT it — mission containers have no route there; the run's router owns
+        # this mission-network address and forwards it across the tailnet), and WHERE the server
+        # has to run. In sidecar mode the tailnet node is the sidecar container, so a port bound
+        # on the host is unreachable no matter how the mission is configured.
+        lines.append(
+            "2. This mission calls BACK to you. Register the address "
```
guru-xorciseai: **Low — this renders as a second "2.", and breaks step 1's cross-reference.**

For a lab run the prompt now emits:

```
1. Join the run's tailnet ...
   ($BASE and $BEARER are the run-control base URL and bearer header shown in step 2.)
2. This mission calls BACK to you. ...
2. Run-control (REST) — authenticate EVERY call ...
```

The run-control step below stays `"2."` whenever `join_key` is set, so there are two step 2s, and
an agent told to look in "step 2" for the base URL and bearer finds the callback paragraph — which
contains neither.

Nothing asserts prompt step numbering, so the unit lane is green. Numbering the steps off a counter
rather than by literal would stop this recurring.

src/xorcise/core/runner/netoverride.py
christo-xorciseai: Confirmed and fixed in ea7a977. Reproduced it here too — 4 of 5, and the mechanism is exactly as described: when the router wins, the unpinned services get .3/.4; when they win, the router gets `Address already in use`.

Moved to the top of the range as suggested, next to ingress: `ROUTER_OFFSET = 2` against `INGRESS_OFFSET = 1`, both derived from the broadcast address by the same helper, so they stay in range for a carved /25. Same stack, 0 failures in 6.

`_assert_reserved_addresses_free` now compares against `router_address(cidr)` rather than testing `octet == ROUTER_OCTET`, so the reservation moves with the address instead of needing to be kept in sync by hand.

src/xorcise/core/runner/netoverride.py
christo-xorciseai: Confirmed and fixed in ea7a977. Verified the routing claim directly — a container on an internal `10.200.1.0/25` has only the on-link route, and `10.200.1.254` in the sibling `/25` is `sendto: Network unreachable`.

Took both halves of your suggestion, because they fix different things:

- Ingress is now armed on **every** entry subnet. There is no single address that can work across segments once they are internal, so one address per segment is the only shape that does. The watcher DNATs the whole set to the agent.
- The address the prompt advertises now comes from `next(iter(entry_subnets))` — manifest order — rather than `sorted()[0]`. Your `zulu_net`/`alpha_net` example is the case: collation order is not something the author controls, so the agent-facing network was being chosen for them.

Live check on the two-segment shape: `10.200.1.126` and `10.200.1.254` each armed on their own interface, and the router resolved them by matching its pinned IP rather than assuming `eth0` — alpha landed on eth0 and zulu on eth1, so the ordering assumption would have been wrong there anyway.

src/xorcise/core/runner/netoverride.py
christo-xorciseai: Confirmed and fixed in ea7a977. `iptables -> iptables-legacy` in the image, `iptables-nft` present, no `nft` binary, and no `jq` either (which rules out your suggestion on the next comment — see there).

The prologue now resolves `IPT=iptables-nft` with a fallback, and every rule that matters is checked explicitly: the SNAT failing aborts the boot rather than continuing to `exec containerboot`, and a DNAT failing in the watcher signals PID 1. Booting unarmed is no longer a thing it can do.

One thing worth recording from the live check, since it looks alarming and is not: once tailscaled is up in `nft-forced` mode, `iptables-nft -t nat -S` reports ``table `nat' is incompatible`` — it cannot render tailscale's native nft tables. That is display-only. `-A` and `-C` against the same table both return 0, which is all the prologue and the watcher use.

src/xorcise/core/runner/netoverride.py
christo-xorciseai: Confirmed and fixed in ea7a977, though not quite the way you suggested — there is no `jq` in the image (`command -v jq` comes back empty on `tailscale/tailscale:stable`), so the exact `--json` filter is not available.

Filtering the plain output on `offline` instead: `grep -v "offline"` before `head -1`. That is the column `tailscale status` prints for a peer with `Online == false`, so it catches the stale-registration case you describe — the dead node is precisely the one that is offline. If both rows are momentarily online the choice is still arbitrary, but the watcher re-points on change, so it converges rather than sticking.

A transient empty result now leaves the existing DNAT alone rather than tearing it down, which is the behaviour you want when `status` is briefly unavailable.

src/xorcise/core/runner/netoverride.py
christo-xorciseai: Agreed and fixed in ea7a977 — dropped the protocol match entirely rather than adding `-p udp`. Verified `iptables-nft -t nat -A PREROUTING -d <ip> -j DNAT --to-destination <ip>` is accepted with no `-p`, which covers TCP, UDP and ICMP in one rule.

Your framing is the reason I went that way rather than enumerating protocols: the prompt promises the agent any port, and a mission author should not have to know which protocols the fence happens to forward. There is a test asserting no protocol match appears on the DNAT line, so it cannot quietly come back.

src/xorcise/core/runner/netoverride.py
christo-xorciseai: Confirmed and fixed in ea7a977. Reproduced your exact output before changing anything — `compose_network_names` on that document returned `('False', 'True')`.

Switched to `yaml.load(text, Loader=yaml.BaseLoader)`, which leaves every scalar a string and now returns `('no', 'on')`. Test added with your document as the case.

Worth noting why this one rated a fix rather than a shrug despite how exotic the input looks: it is not a parse error, it is the confinement pass and Compose silently disagreeing about a network's *name*, which means the override confines a network that does not exist and leaves the one that does untouched.

src/xorcise/core/runner/netoverride.py
christo-xorciseai: Correct, and fixed as you predicted — arming ingress on every entry network (the fix on line 314) makes this check right as written, so it needed no change of its own beyond moving off the octet comparison. The reservation now exists on every entry network, which is what the check was already assuming.

src/xorcise/core/runs/join.py
christo-xorciseai: Confirmed and fixed in ea7a977. This was the one I was most glad you caught — the offline host is precisely the host where the sidecar cannot start, so `auto` preferring it turned a supported path into a dead one.

`auto` now records that it chose the sidecar (`TS_FALLBACK=yes`), and a sidecar that fails to start falls through to native with a message saying what is being given up (host userspace mode puts netstack on this host's loopback). An explicit `XORCISE_TAILSCALE_MODE=sidecar` never falls back — someone naming the mode is diagnosing something and wants the failure, not a quiet downgrade.

The double `docker info` is gone too: one probe into `HAVE_DOCKER`, read by both the detection and the guard.

src/xorcise/core/runs/join.py
christo-xorciseai: Agreed, fixed in ea7a977 — it now exits 1 with the reason rather than binding the wildcard.

Also widened the lookup while I was there, since one of your triggers was the interface not being named `eth0`: it selects the first non-loopback address from all interfaces rather than reading `eth0` specifically. So the fail-closed path is now genuinely the last resort rather than something a custom bridge setup could trip.

src/xorcise/core/runner/service.py
christo-xorciseai: Agreed on both counts, fixed in ea7a977.

`read_image_file` is now `@abstractmethod` — no base `return None` to inherit, so a driver declining to read has to write that down. `StubDockerDriver` overrides it explicitly and says why.

The caller half needed a discriminator, since the stub legitimately returns `None`: `DockerDriver.reads_image_files` (False on the base, True on `DockerSdkDriver`). `_deploy_env` refuses to deploy when a driver that *has* an image store yields nothing and the mission is being confined. The stub creates no networks, and an `allow_egress` mission is not being confined, so neither is held to it.

Tests cover all four combinations.

src/xorcise/core/runner/service.py
christo-xorciseai: Agreed, fixed in ea7a977 — threading `environment.compose_file` through `MissionRef` rather than adding a third hardcoded copy. `MISSION_COMPOSE_PATH` became `MISSION_BUNDLE_DIR`, joined with the authored name at the call site.

`docker.errors.NotFound` from `get_archive` is now translated to `NotFoundError` as well, so a wrong path is a domain error rather than a 500. Though with the fail-closed guard on the previous comment it would have been a refused deploy rather than a silent unconfined one either way.

src/xorcise/core/headscale/policy.py
christo-xorciseai: Agreed, fixed in ea7a977 — hoisted above the loop and `router_tag` is now a required keyword argument, which also puts it in line with `render_policy`, which already required it.

Your point about the three existing tests was the useful part. They now pass the tag explicitly so they test what their names claim, and `test_safe_refuses_to_certify_without_a_router_tag` is parametrised over `[net]` and `[]` — the empty case being the one that actually regressed, since a zero-iteration loop skipped the gate and the foreign-dst sweep together.

src/xorcise/core/runner/docker/driver.py
christo-xorciseai: Confirmed both and fixed in ea7a977. `docker image inspect docker:29.7.1-dind` shows `{"/var/lib/docker":{}}` here too, so every deploy was orphaning one.

`remove(force=True, v=True)`, and `MANAGED_LABEL` at `create`. Left `RUN_ID_LABEL` off deliberately — this call knows an image, not a run, and a run-id label carrying an image ref would be worse for diagnostics than no label. `reap_managed` filters on `MANAGED_LABEL` alone, so the reap half works.

src/xorcise/core/rest/run_create.py
christo-xorciseai: Agreed on the diagnosis, fixed in ea7a977 — though by changing what the prompt claims rather than by adding a signal.

You are right that `entry_subnets` is never empty for a lab, so the address is always populated. But that is now correct: agent reachability is a property of every lab run in this PR, so the address genuinely does exist on every one of them. What was wrong was the sentence — asserting *this mission calls BACK to you* as an event when all we know is that it can.

It now reads *If you need the mission to reach YOU — a callback, a beacon, a reverse shell — it can, at ...*, and closes with a line saying nothing here means the mission will call back and to set this up only if the mission asks for an address. That removes the false premise and the wasted budget without needing a manifest flag whose only consumer would be prose.

If a real signal turns up later — a check that consumes a callback, say — gating on it would be a strict improvement, but I would rather not add a flag authors have to remember to set for a paragraph that is now accurate either way.

src/xorcise/core/runs/prompt.py
christo-xorciseai: Confirmed and fixed in ea7a977. Both steps numbered 2, and step 1 pointing at the wrong one.

Numbered off `_run_control_step()` now, which both the cross-reference and the run-control heading read from, so they cannot drift apart. Parametrised test over the three shapes — join + callback + run-control, join + run-control, and static — asserting the emitted step numbers are `[1,2,3]`, `[1,2]`, `[1]` and that step 1 points at whichever number run-control actually landed on.

christo-xorciseai:
All fifteen addressed in ea7a977. I reproduced each empirical claim before changing anything rather than reading and agreeing; every one held, and two were worse than reported.

**The two blockers.** The `.2` race reproduced 4 in 5 here, and isolating it showed the mechanism exactly as you described — when the router wins the unpinned services get .3/.4, when they win the router gets `Address already in use`. At the top of the range: 0 in 6. (An early "`.253` also fails 2/5" reading of mine was a broken `sed` that never applied — the line ends in `}` — so it was still testing `.2`.) The internal-network claim reproduced verbatim, `sendto: Network unreachable` for a sibling `/25`.

**One thing your live check would have caught next.** Fixing 314 by arming every entry network exposed that the interface lookup mattered more than I had credited: on the two-segment stack, alpha landed on **eth0** and zulu on **eth1**. Resolving the interface by the router's pinned IP rather than by name was already right, but that ordering is not stable and anything assuming it would have armed the wrong interface.

**A regression neither of us had flagged, found while validating live.** `EnvironmentSpec` is `extra="forbid"`, and the catalog is versioned independently of this code. Stripping `agent_ingress` in 001af9b turned every already-published manifest that sets it into a hard install failure:

```
$ xorcise run create --agent cc-opus-4.8 --mission segmented-pivot
error: pull failed — 1 validation error for MissionManifest
environment.agent_ingress
  Extra inputs are not permitted
```

That is any client that did nothing but upgrade, on the missions this PR most needs to work. The field is accepted and ignored again, with a test, until the catalog stops emitting it. No unit test caught it because the unit lane validates manifests it constructs itself.

**Verified live, end to end**, on `segmented-pivot` pulled fresh from the registry: both mission networks `internal=true` — including `internal_net`, which only the compose declares, so the read-from-image path and the authored `compose_file` are both exercised — router at `.253` with `.254` armed beside it, and the dual-homed `web` service reaching its own-segment callback address while getting `Network unreachable` for `1.1.1.1`.

One observation worth recording, since it looks like a failure and is not: once tailscaled is up in `nft-forced` mode, `iptables-nft -t nat -S` reports ``table `nat' is incompatible`` — it cannot render tailscale's native nft tables. Display only; `-A` and `-C` against that table both return 0, which is all the prologue and watcher use.

Lanes: 2300 passed across unit/topology/adapters (+21), mypy clean on 488 files, ruff + format + import contracts clean. `test-heavy` is still skipped on this PR and this is the change that most wants it — if there is a label that opts a PR in, worth using here.

guru-xorciseai:
Pulled `ea7a977` and reran everything you claimed — it all reproduces: **2301 passed** across unit/adapters/topology, **mypy clean on 488 files**, ruff + format + `lint-imports` (4 kept / 0 broken) clean. The `agent_ingress` catch is the one I'd have missed too; `extra="forbid"` against an independently-versioned catalog is a nasty shape, and you're right that the unit lane structurally can't see it because it validates manifests it builds itself.

I then ran a review over the diff and verified the findings myself rather than pass them along. Four hold, one I reported and then had to withdraw, and several I'm listing as questions rather than defects because I haven't proven them.

---

## 1. The inbound ACL is tag-scoped, not run-scoped — cross-run reachability

`policy.py:81` emits `src: [router_tag]` per network, but `router_tag` is one global value, not per-run:

```
config/__init__.py:144   router_tag: str = "tag:router"      # one value
run_create.py:198        router_tag=settings.router_tag       # same for every run
controller.py:126        tags=[self._router_tag]              # every router gets it
```

With runs A and B live the policy holds `{src:[tag:router], dst:[A-agent:*]}` **and** `{src:[tag:router], dst:[B-agent:*]}`. B's router carries `tag:router`, so it matches A's rule — any port, by design.

The comment above the rule says *"scoped to this run's router as the ONLY source."* It's scoped to the tag, and the tag is shared. `assert_policy_safe` can't catch it: it builds `ingress_dsts` as a union across all networks, and `test_safe_rejects_a_router_sourced_dst_for_a_foreign_agent` only ever renders one run, so it passes on a policy that has the property it's meant to reject.

This rule is added by this PR (`+` against `eec976b`), so it isn't inherited. This is the one I'd treat as a blocker.

## 2. A mission network named `xorcise-egress` silently loses confinement

Verified by calling `build_net_override` directly:

```
dmz_net         -> {"ipam": {...}, "internal": true}
xorcise-egress  -> {}
```

The loop sets `internal: True`, then `networks[EGRESS_NET] = {}` at line 313 overwrites it. That service keeps full off-box egress *and* shares an L2 segment with the run's router, while every sibling network is confined — and the run reports itself confined. Needs a mission author to pick the name, so it's not reachable by accident, but there's no reserved-name guard and no error.

## 3. The watcher logs "armed" after a fatal DNAT failure

`netoverride.py:163` — the `|| { echo FATAL; kill -TERM 1; }` block has no `break`. The `for iip in $INGRESS_IPS` loop continues, then falls through to `cur="$aip"` and `echo "xorcise: agent ingress [$INGRESS_IPS] -> $aip"`. The last lines in the log claim ingress is armed for addresses that were never DNATed — the same misleading-log shape the offline-peer test was added to prevent.

## 4. `read_image_file` orphans are reported reaped but never removed — **pre-existing, not yours**

The throwaway container carries `MANAGED_LABEL` and deliberately no `RUN_ID_LABEL`. `reap_managed` returns `[c.name for c in managed]` (including it) but only *removes* via `_remove_run_resources(run_id)`, keyed on the label it doesn't have. So it's printed as reaped and survives every future reap, with its anonymous `/var/lib/docker` volume.

The gap is in `reap_managed` (driver.py:392-404), which this PR doesn't touch — your new call site just becomes a second producer of orphans it can't clear. Worth a follow-up issue rather than a change here. The comment at line 197 does claim the opposite, though, so it's worth fixing the comment at least.

---

## One I withdraw

I had flagged `yaml.BaseLoader` missing merge-key networks as a confinement fail-open. **That's wrong** — checked empirically:

| case | `compose_network_names` |
|---|---|
| service `<<: *anchor`, net declared top-level (valid compose) | `('hidden_net',)` — caught |
| service `<<: *anchor`, net not declared top-level | missed, but that compose is invalid |
| merge key inside top-level `networks:` | `('<<', 'hidden_net')` |

Compose requires the top-level declaration and that path catches it, so confinement holds. The real residue is only the spurious `'<<'` entry in the third case, which would create a junk network. Minor, and your BaseLoader reasoning for the YAML 1.1 `no`/`on`/`yes` problem stands.

Also withdrawing the `ip addr add ... || true` / `dev=eth0` fail-open at `netoverride.py:135` — it didn't survive checking. The weaker true statement is just that the `2>/dev/null || true` swallows a genuine add failure, which is a diagnosability issue, not a fail-open.

---

## Unverified — questions, not claims

I ran out of time to prove these, so please treat them as things to check rather than defects I'm asserting:

- `service.py:140` — the new fail-closed `ValueError` and `read_image_file`'s `NotFoundError` don't appear in `runs.py`'s except chain, so they may surface as a bare 500. (The `except Exception: delete_run(); raise` rollback that skips `teardown_run_network` is pre-existing, not this PR.)
- `service.py:129` — confinement reads the manifest's `compose_file`, but `entrypoint.sh` hardcodes `-f /mission/docker-compose.yml`. Can those two ever name different files?
- `join.py:85` — `auto` picking the sidecar on non-root Linux writes `XORCISE_SOCKS5` where the userspace path wrote `ALL_PROXY`/`HTTPS_PROXY`/`HTTP_PROXY`. Does anything source `ts-env.sh` expecting the latter?
- `driver.py:188` — `read_image_file` calls `containers.create` without `platform=`, while `run()` passes `spec.platform or self._platform`.
- `run_create.py:548` — the prompt advertises only the first entry segment's ingress address; under confinement, is that reachable from the other segments?
- `netoverride.py:310` — does `internal: true` land on networks a mission declares `external: true`, and does Compose reject that combination?
- `prompt.py:152` — the callback step number is back-computed as `rc_step - 1` rather than taken from the shared counter.

---

On your last point: **agreed, and it generalises.** `test-heavy` being PR-skipped just cost us `main` — it's been red since 19 Aug on `test_up_serves_ui_then_down`, which turned out to be a 20-40s Docker pre-warm started with `asyncio.to_thread` gating process exit (fix in #65). Same blind spot, found independently on the same day. Whatever we do for this PR, the lane being a post-merge tripwire rather than a gate looks like the actual problem.

christo-xorciseai:
Pushed as `e51b337`. Validated each finding empirically before touching anything; all four hold, and I worked the seven questions too — several turned into fixes, one confirmed latent-only, one was already correct.

## The blocker — cross-run reachability

Reproduced exactly as you described: a two-run render put `{src:[tag:router], dst:[A-agent:*]}` and `{src:[tag:router], dst:[B-agent:*]}` in the same policy, and `assert_policy_safe` passed it (the union-of-dsts hid it — every dst was "allowed" for the one shared tag).

Fix: each run's router now carries a **per-run tag** `tag:router-<agent-user>` alongside the base tag. The base tag survives only to auto-approve the one route identical across runs (the collector /32); every per-run subnet and every inbound rule is scoped to the run's own tag. `assert_policy_safe` now pairs each router tag with its single permitted agent — the crafted cross-run rule **and** the old bare-`tag:router`→agent shape both fail the gate now.

Verified against real headscale with two simultaneous `segmented-pivot` runs:
```
tagOwners: tag:router, tag:router-run-70a0485f-agent, tag:router-run-fe721a90-agent
autoApprovers: 10.200.1.0/24 -> [tag:router-run-70a0485f-agent]
               10.200.2.0/24 -> [tag:router-run-fe721a90-agent]
inbound: [tag:router-run-70a0485f-agent] -> [run-70a0485f-agent@:*]
         [tag:router-run-fe721a90-agent] -> [run-fe721a90-agent@:*]
```
Both routers online carrying both tags, routes approved per-run, each stack confined.

## The other three

- **xorcise-egress collision** — reproduced (`internal:true` overwritten by `{}`). Reserved-name guard refuses it, scoped to confining runs (an `allow_egress` mission may still use the name).
- **watcher false-armed log** — right, `kill -TERM 1` doesn't stop the subshell. Added `exit 1` on the fatal DNAT path so it stops before the success log.
- **reap_managed** — you're right and my round-1 reply was wrong: it reports by `MANAGED_LABEL` but removes by `_remove_run_resources(run_id)`, so the label-less throwaway was reported-reaped-and-left. Since that container is the orphan class **this PR introduced**, I closed it here (reap_managed force-removes managed containers with no run id, `v=True`) rather than only fixing the comment. The general report/remove divergence you flagged as pre-existing is filed as #66, and the misleading comment is corrected. Real-daemon adapter test extended to assert the label-less container is actually gone.

## The withdrawals

Confirmed both independently — the merge-key case only leaves a spurious `'<<'` entry (BaseLoader reasoning for `no`/`on` still stands), and the `ip addr add ... || true` is diagnosability, not fail-open. No change on either.

## The seven questions

1. **runs.py except chain** — real: the confinement `ValueError` and `NotFoundError` were falling through to a bare 500. Gave the confinement refusals a typed `EnvironmentConfigError` (a `ContractError` that also subclasses `ValueError`, so the pure builders stay `pytest.raises(ValueError)`-compatible) and mapped both to 409 with the collision named.
2. **compose_file vs the hardcoded entrypoint** — real divergence, and it traced to your own round-1 suggestion to thread `compose_file`: doing that on the read side while the entrypoint stayed hardcoded is exactly the two-sources-that-can-disagree you spotted. Fixed both to read one source — the entrypoint now composes `$XORCISE_COMPOSE_FILE` (default `docker-compose.yml` preserved, backward compatible; base MAJOR unchanged since it's non-breaking).
3. **ts-env.sh var names** — nothing internal consumes it; the consumer is the agent and the prompt is mode-agnostic, so no functional break. But you're right the names diverged now that Linux `auto` can pick the sidecar. Made `XORCISE_SOCKS5` the portable handle written in every proxy mode.
4. **read_image_file platform=** — added, mirroring run()/pull(). Static-file extraction doesn't depend on arch, but one selection rule beats two.
5. **first-segment ingress reachability** — latent-only: **zero** installed missions declare more than one entry network (the segmentation in segmented-pivot/tessera is on internal, non-entry nets). The per-segment DNAT mechanism is already in place should a multi-entry callback mission ever ship; documented the single-address advertisement as the known limitation rather than expand the connect contract speculatively.
6. **internal:true on external:true** — compose accepts the combo and `up` succeeds, but `internal` is a **no-op** on an external network, so confinement would silently not hold. No mission uses one and the fresh inner dockerd makes them non-functional anyway, but silent-false-confine is the exact fail-open this PR closes elsewhere — so a confined run now refuses an external network.
7. **rc_step - 1** — replaced with `_step_numbers`, one running counter assigning each section its number by position, so a new conditional step can't desync the numbers.

## On test-heavy

Strong agree, and the `main` breakage makes the case better than I could. The lane being a post-merge tripwire rather than a gate is the actual problem — I'll defer to you on whether to opt this PR in before merge; my validation here was again local (real headscale, two live runs, both multi-network stacks), unwitnessed by CI.

Lanes: 2309 passed across unit/topology/adapters (+9), mypy clean on 488 files, ruff + format + import contracts clean.

PR #80, opened 2026-09-07
fix(runner): probe nested containers at the mission's platform, not a hard-pinned amd64

### LINKED ISSUE #79: Nested-container probe hard-pins linux/amd64: arm64 Linux hosts are refused with a misleading remediation, even for native-arm64 missions (+4 related findings)
## Summary

On an arm64 Linux host, `xorcise run create` for a lab mission fails at the nested-container precondition (HTTP 503, CLI exit 1) with:

> this host cannot run a mission's containers inside the run container — the DinD probe's inner daemon never came up (exit 255). XORCISE runs each mission's containers INSIDE its own container, which this host could not do. Check that Docker can run privileged containers, then re-run 'xorcise doctor'. To bypass this check on a host you know is fine, set XORCISE_NESTED_CONTAINER_CHECK=skip

Privileged containers are not the problem. The Tier-2 probe (`verify_nested_amd64`, `src/xorcise/core/runner/docker/rosetta.py:288`) starts the DinD wrapper at `platform="linux/amd64"` **unconditionally**, so on any non-amd64 host without x86 emulation the wrapper cannot exec at all (`exec format error`, status 255). The probe reads that text and discards it, and off macOS the remediation selector can only ever emit `_FIX_GENERIC`. Net effect: a host that nests perfectly well natively, running a mission that publishes a native arm64 image (19 of the 27 catalog missions do), is refused — and told to fix the wrong thing. Five smaller findings compound it (below).

Consolidated from an operator feedback sheet. Every claim was re-validated against `main` @ 48caafb and reproduced empirically on an amd64 host by mirroring the architecture (recipe at the end).

**Reporter environment**

| | |
|---|---|
| XORCISE | 0.1.3 (pipx) |
| OS | Kali Linux, `aarch64`, kernel `6.18.12+kali-arm64` |
| Docker | arm64-native engine (client and server arm64), overlay2, runc |
| x86 emulation | none registered at first failure (no qemu binfmt handler) |
| Python | not reported |

### 1. Linux remediation is always "check privileged containers" — the real cause is captured and thrown away (High)

**Code**

- `rosetta.py:288` — `platform="linux/amd64"` on the Tier-2 wrapper.
- `rosetta.py:319-321` — the "inner daemon never came up (exit N)" branch returns without the container logs it has already read into `text`.
- `rosetta.py:393-401` — `rosetta_at_fault = macos and (...)`; off macOS the only reachable remediation is `_FIX_GENERIC`.

**Empirical** (amd64 host, arm64 DinD pulled by digest, no binfmt handler — the mirror of the reporter's setup; the real `verify_nested_amd64` + `check_nested_support(on_macos=False)` with the client's `platform` kwarg forced to the foreign arch):

```
detail      : the DinD probe's inner daemon never came up (exit 255)
wait status : {'StatusCode': 255}
container logs the probe read and DISCARDED on this branch:
    'exec /usr/local/bin/dockerd-entrypoint.sh: exec format error\n'
remediation : XORCISE runs each mission's containers INSIDE its own container, which this host could not do. Check that Docker can run privileged containers, then re-run 'xorcise doctor'
```

Byte-for-byte what the reporter saw. `tests/unit/test_rosetta_probe.py::test_a_linux_failure_gets_the_generic_fix_not_rosetta` pins the current wording, so a fix must update it.

### 2. The probe proves *amd64* nesting, not *host* nesting (High)

The module docstring (`rosetta.py:11`, "can this host run a container inside a container?") and the `NESTED_PROBE_IMAGE` comment (`rosetta.py:82`, "This probes HOST capability (can this machine nest at all)") describe a host check; the code performs an amd64-only check. Meanwhile the pull spine already picks a platform per mission (`rest/mission_pull.py::_select_platform`, native-first since #57), and the live catalog publishes `linux/arm64` for 19/27 missions (`operation-tessera` included). So a native-arm64 mission that deploys and grades fine once the check is bypassed is blocked by a probe for a capability it never needs.

**Feasibility.** At `rest/run_create.py:637` the gate runs after `installed` is resolved, and `installed.platform` is already consumed at lines 837 and 947 — the amd64-specific probe can be conditioned on the mission's resolved platform. The auto-pull precheck (`run_create.py:609`) runs before download, but the catalog's `platforms` list is known there. `require_nested` currently takes no arguments.

### 3. `nested_container_check = "skip"` in `config.toml` has no effect until the server restarts, and the error only names the env var (Medium)

`config/__init__.py:227` — `get_settings` is `@lru_cache`; the TOML source is read once per process. The server's only `cache_clear()` calls are in `rest/config_view.py`, for `.env` writes from the settings UI (which does not expose this key). `xorcise up` spawns `serve` as a subprocess that inherits `os.environ` at launch. The remediation at `rest/docker_runtime.py:145-146` mentions only `XORCISE_NESTED_CONTAINER_CHECK=skip`. The key is not documented on docs.xorcise.ai (checked reference/configuration, reference/cli, operate/configuration, operate/troubleshooting, start/quickstart).

**Empirical**

```
1. first load, config.toml=enforce      -> enforce
2. after editing config.toml to skip    -> enforce (same object: True )
3. after get_settings.cache_clear()     -> skip
4. env=skip beats config.toml=enforce   -> skip
```

### 4. An emulated DinD cannot bring up its netfilter (qemu-user + nf_tables), and the probe then reports a bare timeout (Medium)

With x86 emulation registered, a plain amd64 container runs, but the amd64 `docker:29.7.1-dind` inner daemon dies at `iptables: Failed to initialize nft: Protocol not supported`. Reproduced in mirror (qemu-aarch64 via `tonistiigi/binfmt`, arm64 DinD on the amd64 host):

```
wrapper arch: aarch64 | alpine 3.24.1 | iptables: Failed to initialize nft: Protocol not supported
INNER DAEMON NEVER CAME UP (waited 76s)
dockerd.log: iptables: Failed to initialize nft: Protocol not supported
```

The probe image is Alpine 3.24 / iptables 1.8.13 on the nf_tables backend; `iptables-legacy` is present in the image (a candidate fallback for the emulated case). Two consequences:

- In this state the real probe's inner wait loop (90 × `docker info` + `sleep 1`, each slow under qemu) outlasts the 180 s `container.wait` ceiling. The operator gets `nested amd64 probe failed: ConnectionError: … Read timed out.` plus the privileged-containers advice — no daemon log, no hint. Measured: 180.5 s.
- `_select_platform`'s "AMD64 under the host's emulation layer" fallback (the AS4 notice) is therefore not a viable execution path for the 8 amd64-only missions on arm64 **Linux**; it works on macOS only because Rosetta is not qemu-user. Nothing in code or docs distinguishes the two, and the docs have no mention of emulation limits, qemu, binfmt or privileged containers.

### 5. "Mission images are built amd64-only" is stale in code and docs (Medium)

`config/__init__.py:119-121` still says an arm64 host "must request linux/amd64"; the lines directly beneath (added in #57) describe the native-first behaviour that supersedes it. The same sentence is live on docs.xorcise.ai under reference/configuration and reference/cli ("Mission images are built for amd64 only, so an arm64 host must request linux/amd64 and run them under emulation"). Live catalog (`/v1/catalog`): 19 missions publish amd64+arm64; 8 are amd64-only (`breachpoint`, `cipher-beacon`, `derelict-manifest`, `molten-perimeter`, `pipe-dream`, `rogue-uplink`, `severed-stream`, `verify-you-are-human`).

### 6. `xorcise run events <run-id>` → "No such command" (Low — by design, but no hint)

`cli/commands/run.py:45-49` — `events` is a Typer sub-group whose only command is `export`; the docs give the syntax as `xorcise run events export RUN_ID`. `cli/_errors.py` carries a usage hint for `xorcise run events export` but none for the bare group, so a run id lands on the parser's generic "No such command" (exit 2) rather than a pointer to `export`.

## Expected behaviour

- A host that nests natively runs missions that publish a native image, with no bypass needed.
- When the probe fails, the operator sees the actual cause (exec format error, nft init failure, timeout) and an arch-aware fix. On non-amd64 Linux without an x86 handler: say so, and name `qemu-user-static` or an amd64 host. Do not point at privileged containers.
- `xorcise doctor` states plainly whether amd64-only missions can run on this host.
- The skip remediation names the config key and says a running server must be restarted (or the setting is re-read per run).
- Code comment and docs reflect the multi-arch catalog.

## Suggested direction

- [ ] Condition the amd64 probe on the mission's resolved platform (`installed.platform` / catalog `platforms`); gate the general precondition on a host-native DinD probe.
- [ ] On the never-came-up branch, surface `text` (tail-clipped like `clip_detail`), and have `_NESTED_SCRIPT` print the tail of `/tmp/dockerd.log` on failure; size the inner wait loop so it cannot outlast `NESTED_PROBE_TIMEOUT`.
- [ ] Arch-aware Linux remediation: Tier 1 already runs a privileged alpine and reads `binfmt_misc`; extend it to detect an x86 handler (`qemu-x86_64`) when `platform.machine()` is not x86_64.
- [ ] Remediation text: add `nested_container_check = "skip"` in `~/.xorcise/config.toml` and "restart the server (`xorcise down && xorcise up`)".
- [ ] Fix the `config/__init__.py:119` comment; update docs reference/configuration and reference/cli; add a troubleshooting entry covering the nested probe, the skip key, and "amd64 missions on arm64 Linux via qemu are unsupported / best-effort".
- [ ] Either mark the AS4 emulation fallback macOS-only in `_select_platform`, or have `doctor` warn on arm64 Linux.
- [ ] `_errors.py`: usage hint for bare `xorcise run events <id>` → `xorcise run events export <id>`.

## Reproducing on an amd64 host (mirror)

```sh
# 1. foreign-arch DinD present, no emulation  → exit 255 + "exec format error"
DIG=$(docker buildx imagetools inspect docker:29.7.1-dind \
      | awk '/^  Name:/{n=$2} /Platform:.*linux\/arm64/{print n; exit}' | sed 's/.*@//')
docker pull docker@$DIG   # by digest — under overlay2 a tag holds ONE platform, and the SDK
                          # 404s ("platform does not match") instead of pulling if the tag is amd64
# wrap docker.from_env() so containers.run(...) receives platform="linux/arm64", then call
#   rosetta.verify_nested_amd64(client, image=f"docker@{DIG}")
#   rosetta.check_nested_support(skip=False, probe_tier1=..., probe_tier2=lambda _: p, on_macos=False)

# 2. with emulation → nft init failure; the real probe times out at 180 s
docker run --privileged --rm tonistiigi/binfmt --install arm64
docker run --rm --privileged --platform linux/arm64 docker@$DIG sh -c \
  'unset DOCKER_HOST DOCKER_TLS_VERIFY DOCKER_CERT_PATH; dockerd-entrypoint.sh dockerd >/tmp/d.log 2>&1 &
   sleep 75; docker info >/dev/null 2>&1 && echo UP || grep -i "nft" /tmp/d.log'
docker run --privileged --rm tonistiigi/binfmt --uninstall qemu-aarch64
```

Related: #45 (multi-arch mission-base), #57 (native-first platform selection). The 53 existing probe/platform unit tests pass on `main`; they pin the current behaviour.

christo-xorciseai:
## What changed

The nested-container probe (Tier 2, `runner/docker/rosetta.py`) is now asked about the **execution platform of the mission about to run** instead of being hard-pinned to `linux/amd64`:

- `verify_nested(client, platform=…)` starts the DinD wrapper at that platform (`None` = the daemon's native one, no `--platform` anywhere), runs an inner child of the same platform and checks the child reports the machine that platform implies (`x86_64` / `aarch64`; unknown platforms must match the wrapper's own `uname -m`).
- `check_nested_support(...)` takes `platform` + `host_platform` and picks the remediation from them. A foreign platform on a Linux host gets a new fix that names both platforms and says emulation cannot bring up the inner daemon; the Rosetta fix is now restricted to *amd64 asked of an Apple Silicon Mac*; everything else stays "check privileged containers". `NestedSupport` records the platform the verdict is about.
- `rest/docker_runtime.py` memoises **one verdict per platform**. A native probe (what boot pre-warms and `doctor` runs) is also filed under the daemon's explicit platform spelling, so the first run of a native mission still reads the warm verdict. `require_nested_support(..., platform=…)`.
- The pull spine calls `precheck(selected_platform)` **after** `_select_platform`, and run-create passes the install's recorded platform to the gate (falling back to the local image's platform for installs that predate the record).
- The probe **reads the container output on every failure path**: the exec error when the wrapper cannot start (`exec format error`), the inner dockerd's log tail when the daemon dies (emitted by the script inside a 90 s wall-clock budget), and whatever was printed so far when the SDK wait times out. Previously all of that was discarded and the operator saw `inner daemon never came up (exit 255)` or a bare `ConnectionError: Read timed out`.
- Under overlay2 a tag holds one platform and docker-py only re-pulls on a *missing* image, so a present-but-wrong-platform tag 404s ("platform … does not match"). The probe now pulls the wanted platform and retries once.
- The probe container is removed with `v=True` — the dind image declares `VOLUME /var/lib/docker`, so every probe leaked one anonymous volume.
- The refusal message names both spellings of the bypass (`XORCISE_NESTED_CONTAINER_CHECK=skip`, `nested_container_check = "skip"` in `config.toml`) and that either needs `xorcise down && xorcise up`.
- `xorcise run events <run-id>` now suggests `xorcise run events export <run-id>` with the id carried over (generic: an unknown token on a single-subcommand group whose subcommand takes a positional).
- Stale comment in `config/__init__.py` ("mission images are built amd64-only") replaced.

## Why

Closes #79.

On an arm64 Linux host with no x86 emulation, every `xorcise run create` for a lab mission was refused with *"the DinD probe's inner daemon never came up (exit 255). … Check that Docker can run privileged containers"* — including for the 19 of 27 catalog missions that publish a native arm64 image and would run perfectly. The probe's wrapper could not exec at all (foreign arch, no binfmt handler); the text saying so was read and thrown away, and off macOS the remediation selector could only ever produce the privileged-containers advice. Nesting is a property of the *(host, platform)* pair, and the gate now asks about the right pair.

Findings 3–6 of #79 (config.toml bypass frozen until restart, qemu DinD failing at nftables with no diagnosis, the stale amd64-only comment, `run events <id>` rejected without a hint) are addressed in the same PR because they all sit on the message an operator reads when this gate refuses them.

## Testing

Lint / types / layers:

```
$ uv run ruff check . && uv run ruff format --check .
All checks passed!
514 files already formatted
$ uv run mypy
Success: no issues found in 489 source files
$ uv run lint-imports
Contracts: 4 kept, 0 broken.
```

Test lanes:

```
$ uv run pytest -m unit -n auto -q
2150 passed
$ uv run pytest -m "adapters or topology" -q
179 passed
```

New/updated tests: `tests/unit/test_rosetta_probe.py` (platform-aware remediation: foreign-on-Linux → emulation fix, native-on-Apple-Silicon → not Rosetta, exec-format-error marks foreign even with unknown host), `tests/adapters/test_rosetta_probe_wiring.py` (wrapper follows the requested platform, native pins nothing, output tail on exec error / daemon log / SDK timeout, wrong-platform tag re-pull, anon volume removal), `tests/unit/test_nested_container_support.py` (memo per platform, native pre-warm answers the explicit spelling, message names both bypass spellings + restart), `tests/unit/test_cli_ux.py` (`run events <id>` hint; misspelt subcommand gets the close match instead).

Empirical, against the real daemon on this amd64 host (`linux/amd64`, overlay2, no aarch64 binfmt handler), using the production entry points (`nested_support` / `require_nested_support`, `xorcise doctor`):

| Case | Before (main @ 48caafb) | After |
|---|---|---|
| Native probe (`platform=None`) | pass, 8.2 s | pass, 8.8 s — `nested linux/amd64 verified (x86_64 inside a linux/amd64 DinD)` |
| Explicit `linux/amd64` right after the native pre-warm | n/a (one global memo) | memo hit, 0.000 s, same verdict object |
| Foreign platform (`linux/arm64` on this amd64 host, **no** binfmt) — the mirror of the reporter's situation | `inner daemon never came up (exit 255)` + "check privileged containers" | refused in 16 s: `… (exit 255): exec /usr/local/bin/dockerd-entrypoint.sh: exec format error. this mission's image is linux/arm64, but this host executes linux/amd64 natively and cannot run linux/arm64 containers nested — … Choose a mission that publishes a linux/amd64 image …, or run XORCISE on a linux/arm64 host. To bypass … XORCISE_NESTED_CONTAINER_CHECK=skip … or nested_container_check = "skip" in ~/.xorcise/config.toml, then restart it (xorcise down && xorcise up)` |
| Foreign platform **with** a qemu binfmt handler (`tonistiigi/binfmt --install arm64`, removed afterwards) — mirror of finding 4 | 180 s, `ConnectionError: Read timed out`, nothing else | wrapper execs (`XORCISE_OUTER_ARCH=aarch64`), inner dockerd dies, script hands out its log tail and exits 1 after the 90 s budget (93 s wall, `OOMKilled=false`): `… never came up (exit 1): … modprobe: can't change directory to '/lib/modules' … iptables: Failed to initialize nft: Protocol not supported` + the same emulation remediation |
| Wrong-platform local tag (`docker:29.7.1-dind` held arm64 after the mirror run, native probe asked for amd64) | 404 `platform (linux/arm64) does not match the specified platform (linux/amd64)` → "cannot nest" | re-pulled + retried; `xorcise doctor` green: `✓ nested containers: nested linux/amd64 verified …`; tag back to `linux/amd64` |
| Anonymous volumes (`docker volume ls -qf dangling=true | wc -l`) before / after a probe | +1 per probe | 352 / 352 |
| `xorcise run events 0123…cdef` | `No such command '0123…'` exit 2, no hint | exit 2 + `Did you mean? xorcise run events export 0123…cdef` |

Host state after the runs: no `xorcise.managed` containers, no aarch64 binfmt handler, `docker:29.7.1-dind` tag = `linux/amd64`.

Not run here: a real arm64 Linux host (the reporter's). The arm64 branch of the code is exercised only through the mirror above and the unit/adapters tests; the `full-ci` label would be worth adding if an arm64 runner is available.

**Test lanes affected**

- [x] `unit` — no Docker, fast
- [x] `adapters` — port/adapter contracts
- [x] `topology` — manifest / compose / import / extras parity
- [x] `integration` — needs Docker or a live server process (run-create gate; not run locally end-to-end)
- [ ] `frontend` — Next.js typecheck / vitest

## User-facing impact

- [ ] No user-visible change (internal refactor, tests, docs only)
- [ ] **Breaking change** — describe the break and the migration path below

Behaviour change, not an interface change: hosts that could never run a lab mission before (arm64 Linux, native-arm64 missions) now can; refusals carry a different, platform-specific message; `run events <id>` gains a hint. No CLI options, config keys, REST contracts or schemas change. Internal signatures that changed: `verify_nested_amd64` → `verify_nested(platform=…)`, `RunCreateDeps.require_nested: Callable[[str | None], None]`, `pull_mission(precheck=Callable[[str | None], None])`.

## Documentation

- [ ] No documentation change needed
- [ ] Documentation updated in this PR
- [x] Documentation needed and tracked in a follow-up issue (link it)

The docs site (separate repo) has no page for `nested_container_check` and nothing on arm64 Linux / emulation limits — tracked as findings 3 and 4 of #79; this PR puts the same facts into the error message and the setting's comment.

## Dependencies

- [x] No dependency changes

## Checklist

- [x] **No secrets, credentials, tokens, keys, internal hostnames or customer data** appear in the diff
- [x] No references to internal-only systems in code or documentation

---

### DIFF
--- a/src/xorcise/core/rest/docker_runtime.py
+++ b/src/xorcise/core/rest/docker_runtime.py
@@ -7,12 +7,15 @@ mission, and a "no" must fail the run.
 
 Two separate questions live here:
 
-  * HOST capability — can this machine run a container inside a container at all? A behavioural
-    probe (runner/docker/rosetta.py). Expensive (~20-40 s for Tier 2), so its verdict is memoised
-    IN MEMORY for the server's lifetime (host capability does not change without a Docker restart,
-    which restarts the server). `doctor` bypasses the memo (fresh=True) to report current health.
-    There is deliberately NO on-disk cache: a persisted verdict turned a transient probe failure
-    (a registry blip) into a permanent refusal keyed on a fingerprint that never cleared.
+  * HOST capability — can this machine run a container inside a container AT A GIVEN PLATFORM?
+    A behavioural probe (runner/docker/rosetta.py), asked per execution platform: the native one
+    for a mission that pulled a native image, `linux/amd64` for one running under emulation on an
+    arm64 host. Expensive (~10-40 s for Tier 2), so each verdict is memoised IN MEMORY per
+    platform for the server's lifetime (host capability does not change without a Docker
+    restart, which restarts the server). `doctor` bypasses the memo (fresh=True) to report
+    current health. There is deliberately NO on-disk cache: a persisted verdict turned a
+    transient probe failure (a registry blip) into a permanent refusal keyed on a fingerprint
+    that never cleared.
 
   * ARTIFACT compatibility — was THIS mission's fused image built on a base generation this
     XORCISE can run? A cheap label/tag check (require_base_compatible), no container.
@@ -31,6 +34,7 @@ import subprocess
 import threading
 import time
 from collections.abc import Callable, Mapping
+from typing import Any, cast
 
 from xorcise.core.config import Settings
 from xorcise.core.contracts.errors import (
@@ -46,17 +50,19 @@ from xorcise.core.runner.docker.build import (
 from xorcise.core.runner.docker.rosetta import (
     NestedSupport,
     binfmt_signal,
+    canonical_platform,
     check_nested_support,
-    verify_nested_amd64,
+    verify_nested,
 )
 
 log = logging.getLogger(__name__)
 
-# In-memory verdict, memoised for the process lifetime. The lock serialises the (slow) probe so
-# concurrent first run-creates queue behind ONE probe instead of each launching a privileged DinD
-# and racing a cache file — the failure the removed on-disk cache used to have.
+# In-memory verdicts, one per execution platform (None = the daemon's native platform), memoised
+# for the process lifetime. The lock serialises the (slow) probe so concurrent first run-creates
+# queue behind ONE probe instead of each launching a privileged DinD and racing a cache file —
+# the failure the removed on-disk cache used to have.
 _MEMO_LOCK = threading.Lock()
-_memo: NestedSupport | None = None
+_memo: dict[str | None, NestedSupport] = {}
 #: The last verdict line logged, so a long-lived server logs only when the verdict CHANGES
 #: (a single value, not an unbounded set of every distinct failure string ever seen).
 _last_logged: str | None = None
@@ -64,22 +70,46 @@ _last_logged: str | None = None
 
 def _log_support(support: NestedSupport) -> None:
     global _last_logged
-    line = f"{'ok' if support.ok else 'unavailable'}: {support.detail}"
+    line = (
+        f"{'ok' if support.ok else 'unavailable'} [{support.platform or 'native'}]: "
+        f"{support.detail}"
+    )
     if line == _last_logged:
         return
     _last_logged = line
     (log.info if support.ok else log.warning)("nested containers — %s", line)
 
 
-def _probe(client_factory: Callable[[], object]) -> NestedSupport:
-    """Run both tiers against one client, closing it after. Never memoises — caller decides."""
+def _daemon_platform(client: object) -> str | None:
+    """What the daemon executes natively, from its own version report — or None if unknowable."""
+    try:
+        version = cast(Any, client).version() or {}
+        os_name, arch = version.get("Os"), version.get("Arch")
+    except Exception:  # noqa: BLE001 — unknown, never a crash; only sharpens the memo + message
+        return None
+    return canonical_platform(f"{os_name}/{arch}") if os_name and arch else None
+
+
+def _probe(
+    client_factory: Callable[[], object], platform: str | None
+) -> tuple[NestedSupport, str | None]:
+    """Run both tiers for `platform` against one client, closing it after. Never memoises — the
+    caller decides. Also returns the daemon's native platform, so a native verdict can be filed
+    under its explicit spelling too."""
     client = client_factory()  # may raise (daemon down); the caller wraps it into a typed error
     try:
-        return check_nested_support(
+        native = _daemon_platform(client)
+        # Probe with an EXPLICIT platform whenever one is known: "native" spelt out lets the
+        # wrapper insist on the right image when the local tag holds another platform's copy
+        # (docker keeps one per tag under overlay2), instead of silently running whatever is there.
+        support = check_nested_support(
             skip=False,
             probe_tier1=lambda: binfmt_signal(client),
-            probe_tier2=lambda _t1: verify_nested_amd64(client),
+            probe_tier2=lambda _t1: verify_nested(client, platform=platform or native),
+            platform=platform,
+            host_platform=native,
         )
+        return support, native
     finally:
         close = getattr(client, "close", None)
         if callable(close):
@@ -90,32 +120,48 @@ def nested_support(
     settings: Settings,
     client_factory: Callable[[], object],
     *,
+    platform: str | None = None,
     fresh: bool = False,
 ) -> NestedSupport:
-    """Can this host nest containers? Memoised in memory; `fresh` re-probes (doctor).
+    """Can this host nest containers at `platform`? Memoised in memory; `fresh` re-probes (doctor).
+
+    `platform` is the execution platform of the mission about to run — the one its install
+    recorded — or None for the daemon's native platform (what boot pre-warms and `doctor`
+    reports). The two spellings of "native" share one verdict: a probe made with None is also
+    filed under the platform the daemon reported, so the first run of a native mission reads the
+    pre-warmed answer instead of paying a second probe.
 
     A skipped check answers immediately without touching Docker. Otherwise the verdict is computed
-    once and held; `fresh=True` recomputes AND refreshes the memo, so an operator who fixes Rosetta
-    and re-runs `doctor` unblocks subsequent runs without restarting the server.
+    once per platform and held; `fresh=True` recomputes AND refreshes the memo, so an operator who
+    fixes Rosetta and re-runs `doctor` unblocks subsequent runs without restarting the server.
     """
+    key = canonical_platform(platform)
     if settings.nested_container_check == "skip":
-        return NestedSupport(True, "nested-container check skipped by configuration")
-    global _memo
+        return NestedSupport(
+            True, "nested-container check skipped by configuration", platform=key or ""
+        )
     with _MEMO_LOCK:
-        if _memo is not None and not fresh:
-            return _memo
-        support = _probe(client_factory)
-        _memo = support
+        hit = _memo.get(key)
+        if hit is not None and not fresh:
+            return hit
+        support, native = _probe(client_factory, key)
+        _memo[key] = support
+        if key is None and native is not None:
+            _memo[native] = support  # "native" and its explicit spelling are the same question
+        elif key is not None and key == native:
+            _memo[None] = support
         _log_support(support)
         return support
 
 
 def prewarm_nested_support(settings: Settings, client_factory: Callable[[], object]) -> None:
-    """Compute the verdict ahead of the first run-create (called at boot, best-effort).
+    """Compute the NATIVE verdict ahead of the first run-create (called at boot, best-effort).
 
     Warming the memo at startup means run-create reads it instantly instead of the first real run
-    paying the 20-40 s probe under a client timeout. Never raises: a warm-up failure just leaves
-    the memo cold, and the first run recomputes it."""
+    paying the 10-40 s probe under a client timeout. Only the native platform is warmed: it is
+    what every mission with a native image runs at, and probing a foreign platform at every boot
+    would start a privileged emulated DinD for a case most operators never hit. Never raises: a
+    warm-up failure just leaves the memo cold, and the first run recomputes it."""
     try:
         nested_support(settings, client_factory)
     except Exception as exc:  # noqa: BLE001 — a safety-net warm-up must never break boot
@@ -125,26 +171,42 @@ def prewarm_nested_support(settings: Settings, client_factory: Callable[[], obje
 def require_nested_support(
     settings: Settings,
     client_factory: Callable[[], object],
+    *,
+    platform: str | None = None,
 ) -> None:
-    """Raise NestedContainersUnavailableError unless this host can nest containers.
+    """Raise NestedContainersUnavailableError unless this host can nest containers at `platform`.
 
     Call this BEFORE anything stateful or expensive (subnet reservation, control-plane fence,
     image pull) so a host that cannot run missions costs an error message rather than a
     half-built run that has to be torn down.
     """
-    support = nested_support(settings, client_factory)
+    support = nested_support(settings, client_factory, platform=platform)
     if support.ok:
         return
     # The probe's detail often already ends in a period (it quotes a runtime error verbatim), so
     # only add one where it is missing — a stray ".." reads like a typo in the one message an
     # operator sees when nothing works.
     detail = support.detail.rstrip(". ")
+    where = f" ({support.platform})" if support.platform else ""
     raise NestedContainersUnavailableError(
-        f"this host cannot run a mission's containers inside the run container — {detail}. "
-        f"{support.remediation}. "
+        f"this host cannot run a mission's containers inside the run container{where} — "
+        f"{detail}. {support.remediation}. "
+        # Both spellings of the escape hatch, and the restart: the server reads its settings
+        # once at boot, so editing config.toml (or exporting the variable) while it runs changes
+        # nothing until `xorcise down && xorcise up`.
         "To bypass this check on a host you know is fine, set "
-        "XORCISE_NESTED_CONTAINER_CHECK=skip"
+        "XORCISE_NESTED_CONTAINER_CHECK=skip in the server's environment or "
+        'nested_container_check = "skip" in ~/.xorcise/config.toml, then restart it '
+        "(xorcise down && xorcise up)"
     )
+
+
+def reset_nested_support_memo() -> None:
+    """Drop every memoised nesting verdict (tests)."""
+    global _last_logged
+    with _MEMO_LOCK:
+        _memo.clear()
+    _last_logged = None
 
 
 # Memoised daemon platform: the arch of a running daemon cannot change, but "docker was down,
--- a/src/xorcise/core/rest/docker_runtime.py
+++ b/src/xorcise/core/rest/docker_runtime.py
@@ -52,6 +52,7 @@ from xorcise.core.runner.docker.rosetta import (
     binfmt_signal,
     canonical_platform,
     check_nested_support,
+    host_is_macos,
     verify_nested,
 )
 
@@ -132,8 +133,11 @@ def nested_support(
     pre-warmed answer instead of paying a second probe.
 
     A skipped check answers immediately without touching Docker. Otherwise the verdict is computed
-    once per platform and held; `fresh=True` recomputes AND refreshes the memo, so an operator who
-    fixes Rosetta and re-runs `doctor` unblocks subsequent runs without restarting the server.
+    once per platform and held. `fresh=True` recomputes AND refreshes the memo — for the platform
+    asked about and, when that is the native one (what `doctor` asks), for EVERY other platform a
+    run has asked about since boot. That second half is what makes "fix Rosetta, re-run `doctor`,
+    the next run works" true: the refusal that needs clearing is the amd64 entry on an Apple
+    Silicon host, and a native re-probe alone would leave it standing while doctor printed green.
     """
     key = canonical_platform(platform)
     if settings.nested_container_check == "skip":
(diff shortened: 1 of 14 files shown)


src/xorcise/core/rest/docker_runtime.py
```diff
@@ -90,32 +120,48 @@ def nested_support(
...
-        if _memo is not None and not fresh:
-            return _memo
-        support = _probe(client_factory)
-        _memo = support
+        hit = _memo.get(key)
+        if hit is not None and not fresh:
+            return hit
+        support, native = _probe(client_factory, key)
+        _memo[key] = support
+        if key is None and native is not None:
```
guru-xorciseai: `doctor` calls `nested_support(..., fresh=True)` with no platform, so `key` is None and this only rewrites `_memo[None]` and `_memo[native]`. The amd64 entry is never touched.

That kills the recovery path promised in the docstring just above ("an operator who fixes Rosetta and re-runs `doctor` unblocks subsequent runs without restarting the server") for the one case Rosetta actually applies to — amd64 on Apple Silicon. `doctor` probes arm64, prints a green tick, and the next `run create` still reads the stale amd64 refusal telling them to enable Rosetta, which they just did. Only `down && up` clears it.

Related, while you're in here: the pre-warm now only warms native, so on Apple Silicon the common case (mission with no arm64 image, so `_select_platform` picks amd64) pays the cold probe inline on the first run. `role_all` says that's what made runs look hung and get retried into duplicates. It's a documented choice, not a bug — just noting it's a real change from the old single-amd64 memo, which did warm that case.

src/xorcise/core/runner/docker/rosetta.py
```diff
@@ -373,30 +536,59 @@ def check_nested_support(
...
         detail = f"{detail} (rosetta: {tier1.detail})"
-    return NestedSupport(
-        False,
-        detail,
-        tier1.fingerprint,
-        _FIX_ROSETTA if rosetta_at_fault else _FIX_GENERIC,
-    )
+    if rosetta_at_fault:
+        fix = _FIX_ROSETTA
+    elif foreign and not macos:
```
guru-xorciseai: `elif foreign and not macos` leaves macOS with no foreign-platform remediation at all, so those failures fall through to `_FIX_GENERIC` — the same wrong-fix problem you're fixing for Linux, just on the other OS.

Two ways in:

- Apple Silicon with Rosetta already **on** (`tier1.ok` True), amd64 mission, Tier 2 fails without "rosetta" in the detail — `rosetta_at_fault` needs `not tier1.ok or "rosetta" in lowered`, so it's False, and `not macos` is False too. Generic.
- Intel Mac (native amd64) asked for an arm64 mission — `amd64_wanted` is False, so again generic.

Both operators get sent to fix privileged containers, which aren't broken. `elif foreign and not rosetta_at_fault` seems closer to what the docstring above actually describes, and it drops the OS test from a decision that isn't really about the OS.

Tiny one in the same function: the `skip=True` early return at 559 is now the only path that doesn't set `platform`, so a skipped verdict reports `""` while `nested_support`'s own skip branch passes `key`.

src/xorcise/core/rest/mission_pull.py
```diff
@@ -354,6 +353,12 @@ def abort_if_cancelled() -> None:
         )
         if notice:
             log.warning("%s: %s", mission_id, notice)
+        # A lab mission with a download ahead of it — gate now (nesting AT the platform just
+        # selected), still before the pull. After selection, not before: whether this host can
+        # nest depends on WHICH platform it is asked to nest — amd64 under emulation on an arm64
+        # host is a different (and usually negative) answer from native arm64.
+        if precheck is not None:
+            precheck(selected)
```
guru-xorciseai, line 361: When `_select_platform` returns None — a pre-contract entry with empty `detail.platforms` — we precheck the native platform, but the pull can still land a foreign single-arch image, and the post-pull verification is skipped because `selected is None`. The install records the foreign platform, and run-create then gates on it and refuses, after the whole download the precheck exists to avoid.

Narrow (needs a pre-contract catalog entry), so I'm happy for it to be a follow-up — but it is the one hole left in "refused without first paying the pull".

Separately: `precheck` doesn't appear anywhere under `tests/`. The platform argument and the fact it runs *before* `driver.pull` are the two behavioural changes at the centre of this PR, and nothing pins either — a change that moved the gate below the download would pass every lane. One test asserting precheck gets the platform `_select_platform` chose, and gets it before the pull, would cover it.

guru-xorciseai (commented):
Pulled the stack down and ran all three branches locally. #80 on its own is green — ruff, `ruff format --check`, mypy strict, `lint-imports` (4 kept), 2150 unit and 179 adapters+topology — so it's safe as the first merge.

The arm64 Linux fix itself reads right to me, and thanks for doing the binfmt mirror run; that table made this much easier to review than it would otherwise have been.

Three things I'd like a look at before it goes in. None of them are on the Linux path you set out to fix — they're all on the macOS/memo side that the new per-platform keying opened up.

One aside, not yours: `xorcise run events '[/x]'` dies with a `rich.errors.MarkupError` because the token goes into a markup string unescaped. It does the same on `main`, so it's pre-existing — but you're in `_errors.py` anyway, so a `markup.escape()` there would be a cheap win. A bracketed token also silently vanishes from the suggestion (`run events '[abc]'` suggests `... export ` with nothing after it). Entirely fine to leave for a follow-up.

src/xorcise/core/rest/docker_runtime.py
christo-xorciseai: Fixed in 3b0a0ab. `nested_support(fresh=True)` with no platform (what `doctor` asks) now re-probes every other platform a run has asked about since boot, so the amd64 entry is cleared by the same doctor that prints green; `doctor` also lists those foreign verdicts as extra lines (`nested containers (linux/amd64)`, a warning while refused, plain green once cleared) via `nested_containers_foreign()`. Pinned by `test_a_fresh_native_probe_also_refreshes_every_foreign_verdict` and the doctor test.

On the pre-warm: taken as well — on macOS the boot warm-up now also probes `linux/amd64` when native is not amd64 (the Rosetta case is the expensive cold one); Linux still warms native only, since a foreign warm-up there would start a 90 s emulated DinD for the answer every arm64 Linux host gives. `test_prewarm_also_warms_amd64_on_apple_silicon`.

src/xorcise/core/runner/docker/rosetta.py
christo-xorciseai: Fixed in 3b0a0ab: `elif foreign:` — the OS test is gone, so Apple Silicon with Rosetta already on and an Intel Mac asked for arm64 both get the emulation remediation (its text is OS-neutral now; the nft failure applies inside Docker Desktop's VM the same way). The genuine Rosetta case (Tier 1 failing, amd64 on Apple Silicon) still gets the Rosetta fix. The skip verdict records its platform. Tests: `test_a_foreign_platform_on_macos_gets_the_emulation_fix_when_rosetta_is_not_the_cause`, `test_a_skipped_verdict_still_names_its_platform`.

src/xorcise/core/rest/mission_pull.py
christo-xorciseai: Both in 3b0a0ab. The hole: when `_select_platform` returns None, the spine now inspects what actually landed after the pull and, if it is not the daemon's native platform, calls `precheck(actual)` before anything is installed — the download is paid (unavoidable without a platform list) but a refused install, and the run-create refusal after it, are not. And `tests/unit/test_pull_precheck.py` pins the two facts nothing pinned: the precheck receives the platform `_select_platform` chose (native, or the operator override) and runs before `driver.pull` (event order asserted; a refusing precheck costs no download), plus the re-gate on a foreign landing.

christo-xorciseai:
Review follow-ups landed in 3b0a0ab (see the inline replies). The aside too: `_print_error` now `rich.markup.escape()`s everything it interpolates, so `run events '[/x]'` is a plain usage error and `[abc]` survives into the suggestion — `test_bracketed_tokens_render_literally_in_usage_errors`. Lanes on this branch alone: ruff/format/mypy/lint-imports clean, 2161 unit, 179 adapters+topology.

guru-xorciseai (approved):
Re-reviewed at 3b0a0ab. All three of mine are addressed, and I checked them rather than taking the replies at face value:

- **Remediation selector** — `elif foreign:` confirmed, OS test gone. Dropping the OS from a decision that was never really about the OS is the better fix; agreed the emulation text reads fine for Docker Desktop's VM.
- **`fresh=True`** — confirmed it now re-probes every non-native memo key, so the amd64 entry is cleared by the same `doctor` that prints green. The `nested_containers_foreign()` lines are a nice addition I didn't ask for; surfacing the foreign verdicts is what makes the recovery legible.
- **Pre-warm** — warming amd64 on Apple Silicon only, and leaving Linux native-only, is the right split. A 90s emulated DinD at every arm64 Linux boot for a foregone answer would have been a bad trade.
- **Pre-contract re-gate** — re-gating on what actually landed is the right call given you can't know the platform up front, and `test_pull_precheck.py` pins the two things that had nothing: the precheck gets the selected platform, and it runs before `driver.pull` (event order, so a refusal costs no download). That was the gap I cared most about.

Verified locally: 2161 unit + 179 adapters/topology on this branch alone, ruff/format/mypy/lint-imports clean. All 17 tests you cited exist and pass.

**One left, and it isn't yours** — I'd like it tracked rather than fixed here. `probe_channel` (`_diagnostics.py:234`) is the fifth loopback probe and still has no `trust_env=False`. It's byte-identical on `main`, so it's pre-existing and not a regression from this stack — but it's the same bug you just fixed in four places, and it's the one an operator hits first. Confirmed live against a healthy `--stub` server on 3971:

```
$ HTTP_PROXY=http://127.0.0.1:9 xorcise status
 Interface       Stopped   http://127.0.0.1:3971/ui
 REST API        Stopped   http://127.0.0.1:3971
 OTLP receiver   Stopped   http://127.0.0.1:4971
REST API, OTLP receiver: not responding — start XORCISE with xorcise up

$ HTTP_PROXY=http://127.0.0.1:9 xorcise ui
http://127.0.0.1:3971/ui
XORCISE is not running — start it with xorcise up
```

`status` and `ui` both lie about a perfectly healthy server. `doctor` is fine — it uses `ports_in_use` for the Services block and reports the ports correctly. A one-line change plus a follow-up issue is all it needs; happy for it to land separately so this stack isn't held up.

Approving.

guru-xorciseai:
Filed the two follow-ups so they don't get lost now this is in:

- **#86** — the `probe_channel` proxy gap I mentioned above. Pre-existing (initial commit, zero appearances in this stack's diffs), so explicitly not yours; it's the fifth loopback probe, and the four that could corrupt data are the ones you fixed in #82. Full reproduction and the one-line fix are in the issue.
- **#87** — `test-heavy` (integration + e2e) is opt-in on PRs via `full-ci` but always runs on push to `main`, so all three of these sat review-ready with those lanes never executed, even though #81 and #82 asked for the label in their bodies. I've added `full-ci` to the stack; both lanes pass on all three. The issue proposes triggering them from changed paths (or just always) so it doesn't depend on someone remembering.

Worth knowing from #87: `test-heavy` green means *33 passed, 4 skipped* — the four deepest run-lifecycle tests need `mission-base` built plus a live headscale, and neither CI nor the nightly provides those, so those paths aren't actually covered anywhere. Separate and bigger problem, noted in the issue.

Issue #43, opened 2026-08-12
serve runs every startup hook once per bind address: duplicated watchdogs, plus adjacent readiness/preflight findings


christo-xorciseai:
## Context

While validating the DinD-only branch (#41) by deploying all 27 library missions in parallel, I root-caused a duplicated close-out in the readiness gate and found several adjacent problems in the serving and readiness paths. The immediate fixes are on `fix-fuse-platform-and-boot-singletons` (based on `feat-mission-base-dind-29`); this issue records what remains open and the structural recommendation. Plan: land #41 first, then pick this up.

## Root cause of the duplicated close-out (fixed on the branch, but see recommendation)

`serve` builds **one `uvicorn.Server` per bind address and hands them all the same app object**:

```python
servers = [uvicorn.Server(uvicorn.Config(s.app, host=h, port=s.port, log_level=s.log_level))
           for s in specs
           for h in lifecycle._bind_hosts(s.host or host)]
```

The agent-facing specs resolve to two addresses (IPv4 loopback + `::1`), so every `@app.on_event("startup")` hook fires **twice** — measured: 4 `uvicorn.Server` instances across 2 apps. Two `ReadinessWatchdog`s scanned the same runs; both closed out the same failed run **3.6 ms apart** (timestamped log capture), and the loser's teardown died with `409 removal of container ... already in progress`. Each hook also assigns its instance to a `nonlocal`, so the second overwrote the first and shutdown stopped only one.

The branch guards all three startup hooks (reconcile, budget watchdog, readiness gate). Note for reviewers: the readiness guard **cannot** be a plain `is not None` check — the hook awaits `build_run_create_deps` before it can assign, and the second listener reaches the check while that await is pending. The guard claims a flag synchronously before the first await, and the regression test fires the hooks with `asyncio.gather` (sequential firing hides exactly this race).

Verified live on the same failing run: close-out logged once (was twice), zero teardown failures, zero 409s.

## Recommendation: one `uvicorn.Server` per app, sockets per address

The guards treat the symptom. The structural fix is to pre-bind each address and hand the socket list to a **single** Server per app via `serve(sockets=[...])` — uvicorn runs the lifespan once per Server and creates one listener per socket (`uvicorn/server.py:104`: `await self.lifespan.startup()` once, then `for sock in sockets: loop.create_server(...)`).

I prototyped this against the real apps with the exact `serve()` boot preamble, same ports, and exercised every surface:

| Surface | Result |
|---|---|
[...]

PR #81, opened 2026-09-07
fix(cli): serve one uvicorn server per app so startup hooks run once, not once per address

christo-xorciseai:
> Stacked on #80 (`fix/arm64-nested-probe`). Review the diff against that branch; merge #80 first.

## What changed

- **`serve` runs ONE uvicorn.Server per app, with a pre-bound socket per address** (`commands/serve.py`: `bind_specs` → `serve_bound`). Every address `_bind_hosts` resolves for a spec is bound up front (`_preflight.bind_listener`, same options asyncio's `create_server` would set: `SO_REUSEADDR` on POSIX, `IPV6_V6ONLY` on IPv6) and the socket list is handed to `Server.serve(sockets=…)` — uvicorn's gunicorn-worker path, which runs the ASGI lifespan once per Server and opens one listener per socket. Before, `serve` built one Server per *(app, address)* around the same app object, so every `@app.on_event("startup")` hook fired once per bind address.
- **Binding is the preflight.** A conflict surfaces as a per-address `EADDRINUSE` before any server starts and is printed through `conflict_message` naming the host — not as uvicorn's `sys.exit(1)` from inside `Server.startup()`. The IPv6 loopback companion is the one address allowed to be missing (IPv6 disabled ⇒ one warning, IPv4 still serves); a spec with no bindable address at all is fatal.
- **Port preflight sees IPv6.** `_preflight._bindable` hardcoded `AF_INET`; it now probes in the address's own family, and `find_free_port` also requires `[::1]` free (role:all binds it), while an address family that does not exist on the host reads as *free*, not *taken* (otherwise every port would look busy on an IPv6-disabled host).
- **The startup hooks are guarded** (`roles/boot/role_all.py`): reconcile, budget watchdog and readiness gate each run once per app whatever the ASGI host does. The readiness guard claims a flag *before* its first `await` — an `is not None` check is a check-then-act race across `build_run_create_deps`. This is commit b5cb829 from the unmerged `fix-fuse-platform-and-boot-singletons` branch, re-applied (the rest of that branch was superseded by #41/#57).

## Why

Fixes the root cause and finding 1 of #43.

On Linux role:all resolves three agent-facing addresses (loopback, `::1`, the docker bridge gateway), so the REST app got **three** reconciles, **three** budget watchdogs and **three** readiness gates scanning the same runs; two gates closed out the same failed run milliseconds apart and the loser's teardown died on `409 removal of container … already in progress`. Each hook assigns to a `nonlocal`, so shutdown stopped only the last instance. Separately, a listener on `[::1]:3001` with `127.0.0.1:3001` free passed the IPv4-only preflight and crashed uvicorn from inside `Server.startup()` with a raw traceback.

**Not in this PR** (still tracked on #43): finding 2 (close-out destroys the failure evidence), finding 3 (`compose_service_states` → `()` → READY fails open), finding 4 (`_check_one` tears down twice), finding 5 (90 s readiness default). Those change readiness *behaviour* and deserve their own PR and their own empirical run.

## Testing

```
$ uv run ruff check . && uv run ruff format --check . && uv run mypy && uv run lint-imports
All checks passed! / 514 files already formatted / Success: no issues found in 491 source files / Contracts: 4 kept, 0 broken.
$ uv run pytest -m unit -n auto -q          → 2161 passed
$ uv run pytest -m "adapters or topology" -q → 179 passed
```

New tests: `tests/unit/test_boot_singletons_once.py` (hooks fired 1×/2×/3× concurrently with `asyncio.gather`, exactly one budget watchdog, one readiness gate, one reconcile each time), `tests/unit/test_serve_sockets.py` (one socket per address per app; `[::1]` squatter → `BindConflict` naming `::1` and every socket bound so far released; IPv4 squatter; missing IPv6 loopback is a warning; `serve_bound` hands ALL of an app's sockets to ONE Server), `tests/unit/test_preflight.py` (+3: `find_free_port` walks past a `[::1]` squatter; an unavailable address family reads as free; `bind_listener` family/bound/held-once-listening). Four existing `serve` tests re-seamed from `ports_in_use` to `bind_listener`.

**Empirical — the real `xorcise serve --stub` (role all, real uvicorn, isolated `XORCISE_HOME`), booted from the previous tree and from this branch on this Linux host, with counters wrapped around the production singletons and nothing else changed:**

| | Before (one Server per address) | After (one Server per app) |
|---|---|---|
| Listeners (`ss -tlnH`) | 6: `127.0.0.1`, `[::1]`, `172.17.0.1` × REST, OTLP | 6, same set |
| REST reachable ipv4 / ipv6 / gateway | 200 / 200 / 200 | 200 / 200 / 200 |
| `BudgetWatchdog` built / started | **3 / 3** | 1 / 1 |
| `ReadinessWatchdog` built / started | **3 / 3** | 1 / 1 |
| `reconcile_all_on_startup` runs | **3** | 1 |
| SIGTERM → exit, time, ports left listening | exit 0, 936 ms, 0 | exit 143, 418 ms, 0 (see note) |
| SIGINT (Ctrl-C) → exit, ports left listening | — | exit 0 ("shutting down"), 0 |
| Immediate rebind of both ports after stop | ok | ok |
| Squatter on `[::1]:P`, `127.0.0.1:P` free, `serve --port P` | banner printed, then uvicorn `OSError: [Errno 98] … ('::1', P, 0, 0)`, **109-line traceback**, exit 1 | resolution now sees the IPv6 side: `--port P` relocates to P+1 and serves; no traceback (the `BindConflict` path is covered by the unit tests) |

Note on exit 143: uvicorn's `capture_signals` re-raises the captured signal after a graceful shutdown, so a single Server terminates by SIGTERM (143), which is uvicorn's own CLI behaviour. The previous exit 0 was an artefact of six nested handler restorations. `down` waits on process exit and never reads the status; Ctrl-C still exits 0.

**Test lanes affected**

- [x] `unit`
- [ ] `adapters`
- [ ] `topology`
- [x] `integration` — `serve`/`up` boot path; validated by hand above, not in CI
- [x] `e2e` — full `xorcise up` (uses `serve`); please add `full-ci`
- [ ] `frontend`

## User-facing impact

- [x] No user-visible change (internal refactor, tests, docs only)

Behaviour: `serve`/`up` boot the same listeners; conflicts on any address are reported cleanly; a port whose IPv6 side is taken is skipped by auto-increment. No interface change.

## Documentation

## Dependencies

- [x] No dependency changes

## Checklist

- [x] No secrets, credentials, tokens, keys, internal hostnames or customer data in the diff
- [x] No references to internal-only systems

---

### DIFF
--- a/src/xorcise/core/cli/_preflight.py
+++ b/src/xorcise/core/cli/_preflight.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 import errno
 import os
 import socket
+from collections.abc import Sequence
 from collections.abc import Set as AbstractSet
 
 # Human label per known plane port, for the conflict message.
@@ -42,15 +43,19 @@ def address_unavailable(exc: OSError) -> bool:
 
 
 def bind_listener(host: str, port: int) -> socket.socket:
-    """Bind (not yet listening) `host:port` exactly the way asyncio's `create_server(host=…)`
-    would, and hand the socket back for the server to adopt. Raises OSError on failure.
+    """Bind AND listen on `host:port` the way asyncio's `create_server(host=…)` would, and hand
+    the socket back for the server to adopt. Raises OSError on failure.
 
-    Binding here, once, is the preflight: the same socket is then passed to uvicorn, so there is
-    no gap in which another process can take the port between the check and the listen. Options
-    mirror CPython's base_events.create_server — SO_REUSEADDR on POSIX (so a port left in
-    TIME_WAIT by the previous server rebinds immediately) and IPV6_V6ONLY on an IPv6 socket (so
-    the IPv6 loopback listener never shadows or collides with the IPv4 one). asyncio calls
-    listen() itself when it adopts the socket."""
+    The listen() is what makes this an atomic preflight. `SO_REUSEADDR` deliberately lets two
+    sockets bind the same address while NEITHER is listening (that is how a port in TIME_WAIT is
+    reused), so a bound-but-not-listening socket holds nothing: a concurrent `serve` would bind
+    it too, and the loser would only find out at listen() — inside uvicorn's socket-adoption
+    path, after the app's startup hooks had already run, with no shutdown to follow. Listening
+    here means the conflict surfaces in THIS call, where the caller reports it cleanly, and the
+    address is held from this moment on. asyncio's listen() on adoption merely re-applies the
+    backlog. Options mirror CPython's base_events.create_server — SO_REUSEADDR on POSIX (so a
+    port left in TIME_WAIT by the previous server rebinds immediately) and IPV6_V6ONLY on an IPv6
+    socket (so the IPv6 loopback listener never shadows or collides with the IPv4 one)."""
     sock = socket.socket(address_family(host), socket.SOCK_STREAM)
     try:
         if os.name == "posix":
@@ -58,6 +63,7 @@ def bind_listener(host: str, port: int) -> socket.socket:
         if sock.family == socket.AF_INET6 and hasattr(socket, "IPPROTO_IPV6"):
             sock.setsockopt(socket.IPPROTO_IPV6, socket.IPV6_V6ONLY, 1)
         sock.bind((host, port))
+        sock.listen()
     except OSError:
         sock.close()
         raise
@@ -100,8 +106,22 @@ def ports_in_use(host: str, ports: list[int]) -> list[int]:
     return [p for p in ports if not _bindable(host, p)]
 
 
+def _probe_addresses(hosts: str | Sequence[str]) -> list[str]:
+    """The addresses a port must be free on for the listener set `hosts`: every address the
+    server will bind, plus the IPv4 wildcard when any of them is a specific IPv4 interface — a
+    wildcard probe refuses a listener on ANY single interface, so it catches a squatter on an
+    interface the set does not name. Nothing else: the IPv6 loopback is probed exactly when it
+    is in the set (role:all on a local topology), not for a role whose spec never binds it — a
+    `[::1]:8800` listener must not relocate `serve --role runner`."""
+    listen_on = [hosts] if isinstance(hosts, str) else list(hosts)
+    probe = list(dict.fromkeys(listen_on))
+    if any(address_family(h) == socket.AF_INET and h != "0.0.0.0" for h in listen_on):
+        probe.append("0.0.0.0")
+    return probe
+
+
 def find_free_port(
-    host: str,
+    host: str | Sequence[str],
     start: int,
     attempts: int = _SCAN_ATTEMPTS,
     taken: AbstractSet[int] = frozenset(),
@@ -109,28 +129,26 @@ def find_free_port(
 ) -> int:
     """First free port at-or-above `start` (walks start, start+1, … for `attempts` probes).
 
-    A port only counts as free when `host`, the IPv4 wildcard ("0.0.0.0") AND the IPv6
-    loopback ("::1") are all bindable: a wildcard probe refuses a listener on any single IPv4
-    interface, and the IPv6 probe sees the `[::1]` listener an IPv4 probe cannot — role:all
-    binds every agent-facing plane on the IPv6 loopback too, so a port that passes here cannot
-    pass preflight and then crash uvicorn on an address-specific conflict. Reuses _bindable,
-    keeping the SO_REUSEADDR/TIME_WAIT semantics (and "IPv6 disabled" reads as free, not
-    taken). `taken` holds ports already promised to other planes in the same resolution round.
+    `host` is the address the server will bind, or the WHOLE list of addresses it will bind
+    (`_bind_hosts` of the role's spec) — the scan probes exactly that set, plus the IPv4
+    wildcard for any specific IPv4 interface in it (see _probe_addresses). Handing the real
+    listener set in, rather than a hand-maintained guess of it, is what keeps this from
+    drifting behind `_bind_hosts`/`bind_specs` again. Reuses _bindable, keeping the
+    SO_REUSEADDR/TIME_WAIT semantics (and "IPv6 disabled" reads as free, not taken). `taken`
+    holds ports already promised to other planes in the same resolution round.
     """
+    addresses = _probe_addresses(host)
     for candidate in range(start, min(start + attempts, _MAX_PORT + 1)):
         if candidate in taken:
             continue
-        if (
-            _bindable(host, candidate)
-            and (host == "0.0.0.0" or _bindable("0.0.0.0", candidate))
-            and (host == IPV6_LOOPBACK or _bindable(IPV6_LOOPBACK, candidate))
-        ):
+        if all(_bindable(address, candidate) for address in addresses):
             return candidate
     raise PortScanError(label, start, attempts)
 
 
-def resolve_ports(host: str, wanted: dict[str, int]) -> dict[str, int]:
-    """Resolve each plane's requested port to the first free one at-or-above it.
+def resolve_ports(host: str | Sequence[str], wanted: dict[str, int]) -> dict[str, int]:
+    """Resolve each plane's requested port to the first free one at-or-above it, on the
+    address (or the full listener set) the planes will bind.
 
     Planes resolve in dict order; earlier choices feed into `taken` so two planes
     contending for the same base port never resolve to the same one. Raises
--- a/src/xorcise/core/cli/_preflight.py
+++ b/src/xorcise/core/cli/_preflight.py
@@ -6,6 +6,8 @@ actionable conflict message. Used by `up` (before spawning) and `serve` (before 
 
 from __future__ import annotations
 
+import errno
+import os
 import socket
 from collections.abc import Set as AbstractSet
 
@@ -15,6 +17,51 @@ _PLANE = {3001: "rest", 4318: "otlp", 8800: "runner", 8080: "headscale"}
 # How far past the requested port the auto-increment scan walks before giving up.
 _SCAN_ATTEMPTS = 50
 _MAX_PORT = 65535
+
+# The IPv6 loopback every agent-facing spec also listens on (role_all + _bind_hosts). A probe on
+# the IPv4 side alone cannot see a squatter here, which is how a `[::1]:3001` listener used to
+# pass preflight and then crash uvicorn from inside Server.startup() with a bare sys.exit(1).
+IPV6_LOOPBACK = "::1"
+
+# bind() errors that mean "this address does not exist here" rather than "someone holds it":
+# IPv6 disabled on the host, or a configured host that is not a local interface.
+_ADDRESS_UNAVAILABLE = frozenset(
+    e for e in (errno.EADDRNOTAVAIL, errno.EAFNOSUPPORT, getattr(errno, "EPFNOSUPPORT", None)) if e
+)
+
+
+def address_family(host: str) -> int:
+    """AF_INET6 for a literal IPv6 address, AF_INET otherwise."""
+    return socket.AF_INET6 if ":" in host else socket.AF_INET
+
+
+def address_unavailable(exc: OSError) -> bool:
+    """Whether a failed bind means the ADDRESS is not usable on this host at all (as opposed to
+    the port being held by someone). Such an address cannot be "in use"."""
+    return exc.errno in _ADDRESS_UNAVAILABLE
+
+
+def bind_listener(host: str, port: int) -> socket.socket:
+    """Bind (not yet listening) `host:port` exactly the way asyncio's `create_server(host=…)`
+    would, and hand the socket back for the server to adopt. Raises OSError on failure.
+
+    Binding here, once, is the preflight: the same socket is then passed to uvicorn, so there is
+    no gap in which another process can take the port between the check and the listen. Options
+    mirror CPython's base_events.create_server — SO_REUSEADDR on POSIX (so a port left in
+    TIME_WAIT by the previous server rebinds immediately) and IPV6_V6ONLY on an IPv6 socket (so
+    the IPv6 loopback listener never shadows or collides with the IPv4 one). asyncio calls
+    listen() itself when it adopts the socket."""
+    sock = socket.socket(address_family(host), socket.SOCK_STREAM)
+    try:
+        if os.name == "posix":
+            sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        if sock.family == socket.AF_INET6 and hasattr(socket, "IPPROTO_IPV6"):
+            sock.setsockopt(socket.IPPROTO_IPV6, socket.IPV6_V6ONLY, 1)
+        sock.bind((host, port))
+    except OSError:
+        sock.close()
+        raise
+    return sock
 
 
 class PortScanError(RuntimeError):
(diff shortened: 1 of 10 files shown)


src/xorcise/core/cli/_preflight.py
```diff
@@ -16,6 +18,51 @@
...
+    TIME_WAIT by the previous server rebinds immediately) and IPV6_V6ONLY on an IPv6 socket (so
+    the IPv6 loopback listener never shadows or collides with the IPv4 one). asyncio calls
+    listen() itself when it adopts the socket."""
+    sock = socket.socket(address_family(host), socket.SOCK_STREAM)
+    try:
+        if os.name == "posix":
+            sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        if sock.family == socket.AF_INET6 and hasattr(socket, "IPPROTO_IPV6"):
+            sock.setsockopt(socket.IPPROTO_IPV6, socket.IPV6_V6ONLY, 1)
+        sock.bind((host, port))
```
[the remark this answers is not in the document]
christo-xorciseai: Not a new exposure — resolving.

`bind_listener` is reached with `"0.0.0.0"` from exactly one place: `find_free_port`'s wildcard probe (`_bindable("0.0.0.0", candidate)`), which exists on `main` unchanged (`_preflight.py:74`) and binds-then-closes in the same call — it never listens, so nothing is exposed. CodeQL sees a new alert only because the bind moved into the shared helper.

The path that does listen (`serve` → `bind_specs`) binds only what `_bind_hosts` yields from role_all's narrowed agent-facing set (loopback, `::1`, the docker bridge gateway). `0.0.0.0` is bound only when the operator sets `XORCISE_HOST=0.0.0.0` explicitly, or on the documented Linux fallback when the gateway cannot be determined — the same addresses uvicorn bound before this PR. The alert instance is on the PR merge ref only; there is no open alert for this rule on `main`.

src/xorcise/core/cli/commands/serve.py
```diff
@@ -80,6 +89,122 @@ def _warn_experimental_role(role: str) -> None:
...
+            ),
+            b.sockets,
+        )
+        for b in bound
+    ]
+    tasks = [asyncio.create_task(server.serve(sockets=socks)) for server, socks in servers]
+    await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
+    for server, _ in servers:
+        server.should_exit = True
+    results = await asyncio.gather(*tasks, return_exceptions=True)
```
guru-xorciseai, line 209: If an app's ASGI lifespan startup raises, `serve()` returns without an exception, so this filter finds nothing and `serve` exits 0 having served nothing.

uvicorn's path (`server.py:93-108`): `startup()` awaits `lifespan.startup()`, and if that failed it sets `should_exit = True` and returns early — so `self.started` stays False, `main_loop()` is skipped by `if not self.should_exit`, `shutdown()` is skipped by `if self.started`, and `serve()` falls off the end returning None.

`_start_watchdog` is the one hook with no try/except, so a config error in `get_settings()` or the `BudgetWatchdog(...)` construction would do it. From the outside `up` reports only the generic "did not become healthy in 10s", which is a rough thing to debug.

The comment at 279 already states the contract ("A crashed server plane must not exit 0") — checking `not server.started` alongside the exception would make it true.

src/xorcise/core/cli/_preflight.py
```diff
@@ -16,6 +18,51 @@
...
+    """Whether a failed bind means the ADDRESS is not usable on this host at all (as opposed to
+    the port being held by someone). Such an address cannot be "in use"."""
+    return exc.errno in _ADDRESS_UNAVAILABLE
+
+
+def bind_listener(host: str, port: int) -> socket.socket:
+    """Bind (not yet listening) `host:port` exactly the way asyncio's `create_server(host=…)`
+    would, and hand the socket back for the server to adopt. Raises OSError on failure.
+
+    Binding here, once, is the preflight: the same socket is then passed to uvicorn, so there is
```
guru-xorciseai: The docstring's "no gap in which another process can take the port between the check and the listen" isn't right, and I think the gap matters here.

`SO_REUSEADDR` on a bound-but-not-**listening** socket explicitly permits a second bind of the same address (Linux only conflicts when either socket lacks `SO_REUSEADDR` or the other is already `TCP_LISTEN`). Checked it:

```
both bind 127.0.0.1:33701 -> second bind SUCCEEDED (no listen yet)
  first listen():  ok
  second listen(): OSError errno=98 (Address already in use)
```

We don't listen — uvicorn does, later — so two concurrent `up`/`serve` processes can both come through `bind_specs` clean, and the loser only discovers it at `listen()`.

The part that makes this worse than before: it surfaces inside uvicorn's `sockets is not None` branch, which has no `try/except OSError` at all, unlike the host/port branch right below it (which logs and `sys.exit(1)`s). So the `OSError` escapes `Server.startup()` *after* `lifespan.startup()` has already run — watchdog started, reconcile launched — and with `started` still False, meaning `shutdown()` and `lifespan.shutdown()` never fire. Previously uvicorn bound and listened atomically, so this couldn't happen.

Fine if you'd rather scope the docstring and handle the `OSError` than solve the race properly, but as written the comment promises atomicity we don't have.

src/xorcise/core/roles/boot/role_all.py
```diff
@@ -51,8 +51,21 @@ def build_rest_app() -> FastAPI:
...
+    # two of which closed out the same failed run milliseconds apart, the loser's teardown dying
+    # on 409 "removal of container … already in progress" — and, since each hook assigns its
+    # instance to a nonlocal, shutdown stopped only the last one. `serve` now runs one Server per
+    # app (sockets per address), which fires each hook once; the guards stay because they hold
+    # under ANY ASGI host arrangement, which this module cannot see.
     _watchdog: BudgetWatchdog | None = None
     _readiness: ReadinessWatchdog | None = None
+    # Separate from `_readiness` because the gate cannot be assigned until after an await; see
+    # the claim in _start_readiness_gate.
+    _readiness_claimed = False
```
guru-xorciseai, line 71: None of the shutdown hooks clear these. `_stop_watchdog` leaves `_watchdog` set, `_cancel_reconcile` leaves `_reconcile_task` set, `_stop_readiness_gate` leaves `_readiness_claimed` True, and `_prewarmed` stays True too.

So a startup → shutdown → startup on the same app object — two `with TestClient(app)` blocks, or any host that restarts the lifespan — comes back up with all three guards already claimed: a *stopped* BudgetWatchdog, no readiness gate, no reconcile. Silently, which is the bad part; budget deadlines just stop being enforced.

The block comment at 54-63 says the guards hold under any ASGI host arrangement, and a lifespan restart is one they don't. A one-line reset in each matching shutdown hook covers it.

Broader thought, take it or leave it: that's five closure flags implementing one idea, and `_readiness` plus `_readiness_claimed` track the same singleton. The claim-before-await rule you correctly identify applies to all of them, so it might be worth one small guard (a `_claimed: set[str]`, or moving to `lifespan=` with the singletons in one object) rather than five variables whose invariants are explained in three separate comments.

src/xorcise/core/cli/_preflight.py
```diff
@@ -61,17 +109,22 @@ def find_free_port(
...
+    taken). `taken` holds ports already promised to other planes in the same resolution round.
     """
     for candidate in range(start, min(start + attempts, _MAX_PORT + 1)):
         if candidate in taken:
             continue
-        if _bindable(host, candidate) and (host == "0.0.0.0" or _bindable("0.0.0.0", candidate)):
+        if (
+            _bindable(host, candidate)
+            and (host == "0.0.0.0" or _bindable("0.0.0.0", candidate))
+            and (host == IPV6_LOOPBACK or _bindable(IPV6_LOOPBACK, candidate))
```
guru-xorciseai: This makes *every* role require `[::1]` free, but only role:all on a local topology actually binds it — `_bind_hosts` adds the IPv6 companion only when the host is `0.0.0.0`, and control/collector/runner/headscale all resolve to `["127.0.0.1"]`.

So any unrelated `[::1]:8800` listener now relocates `serve --role runner` even though nothing will ever bind `::1` there, and a spread of `::1` listeners can walk the whole 50-port window and fail with a misleading "no free port for runner in 8800-8849".

Worth scoping to the roles that bind it. Which is really the point of the last comment I'd make here — the scan's address set (`host`, `0.0.0.0`, now `::1`) is a hand-maintained guess that has to stay in sync with `_bind_hosts`/`bind_specs`, and this is the second widening after a production miss. Now that binding *is* the preflight, handing the resolution round the spec's real host list would stop it drifting again.

src/xorcise/core/cli/commands/serve.py
```diff
@@ -80,6 +89,122 @@ def _warn_experimental_role(role: str) -> None:
...
+async def serve_bound(bound: list[BoundSpec]) -> list[BaseException]:
+    """Run one uvicorn.Server per bound app until the first exits; return the failures."""
+    import uvicorn
+
+    servers = [
+        (
+            uvicorn.Server(
+                uvicorn.Config(
+                    b.spec.app,
+                    # Only labels the "Uvicorn running on" line — the listeners are the sockets
```
guru-xorciseai: This comment is now wrong, and it costs us something: uvicorn logs "Uvicorn running on …" only when it did the binding (`if sockets is None: self._log_started_message(...)`, `server.py:188`), and the only other emitter is `Config.bind_socket()`, which we never call. So `host=` labels nothing at all, and `serve.log` has lost the six "running on http://…" lines that were the only record of which addresses actually bound.

`_serve_banner` doesn't fill the gap either — it prints one `ui_url()` from config, not the resolved set.

`up`'s failure path tells the operator "see {serve_log}", so it'd be worth printing the address set from `serve` itself. You already have `b.hosts` and `b.spec.port` right here.

guru-xorciseai (commented):
3/3/3 → 1/1/1 is a good result, and one Server per app with a socket per address is the right shape for it. Verified locally on top of #80: ruff, format, mypy, lint-imports clean, 2161 unit and 179 adapters+topology.

Two things I'd want addressed. The bind isn't actually atomic, so the preflight doesn't quite promise what the docstring says it does; and `serve` can exit 0 having served nothing. Both inline.

The CodeQL `0.0.0.0` alert — your resolution reads correctly to me. `find_free_port`'s wildcard probe binds and closes in the same call and never listens, and it's unchanged from `main`. Agreed it's an alert about the code moving, not new exposure.

src/xorcise/core/cli/commands/serve.py
christo-xorciseai: Fixed in 3b9a270: `serve_bound` now treats a Server whose `started` is still False after `serve()` returns as a failure (`RuntimeError("the app on port N never started — its startup failed (see the log above)")`), so `serve` exits 1. Reproduced first with an app sending `lifespan.startup.failed` — uvicorn logs "Application startup failed. Exiting." and `serve_bound` returned `[]` — and pinned by `test_a_failed_lifespan_startup_is_reported_as_a_failure`.

src/xorcise/core/cli/_preflight.py
christo-xorciseai: You were right, and I had confirmed the same thing while fixing a test without drawing the conclusion. Fixed in 3b9a270 by solving it rather than scoping the docstring: `bind_listener` now `listen()`s before returning, so the address is held from that moment — the rival's SO_REUSEADDR bind fails at ITS bind(), inside our `bind_specs`, where it is reported cleanly; asyncio's listen() on adoption only re-applies the backlog. Reproduced the escape first (second bind succeeded, uvicorn raised `EADDRINUSE` after `startup` had run, no `shutdown`) and pinned the fix: `test_bind_specs_holds_the_address_so_a_concurrent_binder_loses_immediately` and the updated `test_bind_listener_hands_back_a_bound_socket_of_the_right_family`.

src/xorcise/core/roles/boot/role_all.py
christo-xorciseai: Fixed in 3b9a270: every shutdown hook releases its guard (`_watchdog`, `_reconcile_task`, `_readiness`/`_readiness_claimed`, and a new `_reset_prewarm` for `_prewarmed`), so a lifespan restart on the same app builds live singletons. Reproduced first (startup → shutdown → startup built nothing) and pinned by `test_a_lifespan_restart_builds_fresh_singletons`. I kept the separate flags rather than one guard object — the reset in each matching shutdown hook keeps the claim-before-await rule visible next to the code it protects — but no objection if you would rather see the consolidation as a follow-up.

src/xorcise/core/cli/_preflight.py
christo-xorciseai: Fixed in 3b9a270, along the lines you suggested: `find_free_port`/`resolve_ports` take the address (or the whole listener set) the planes will bind, and `_resolve_role_ports` hands them `_role_bind_hosts(role, host)` — role:all on a local topology gets `role_all.agent_facing_loopbacks(host)` (the same function `apps()` builds its set from; the docker gateway it adds later is a specific IPv4 interface the wildcard probe already covers, so resolution needs no docker call), every other role the configured host. So `::1` is probed exactly when a spec binds it. Verified live: `serve --role runner` with a `[::1]:PORT` squatter stays on its port; role all with the same squatter relocates (it binds ::1). Tests: `test_role_bind_hosts_is_the_set_serve_binds`, updated `test_find_free_port_sees_a_squatter_on_the_ipv6_loopback`.

src/xorcise/core/cli/commands/serve.py
christo-xorciseai: Fixed in 3b9a270: the comment (and the `host=`) are gone, and `serve` prints one `listening on 127.0.0.1:3001, [::1]:3001, 172.17.0.1:3001` line per app to stderr — which is what `up` redirects into `serve.log`. Verified live via `up`: both lines present in `serve.log`. `test_listener_lines_name_every_bound_address`. (For the record, in this setup `serve.log` had no "Uvicorn running on" lines on `main` either — the app's logging setup does not surface uvicorn's — so the code-level point stands and the operator now gets the record they did not have before.)

christo-xorciseai:
Review follow-ups landed in 3b9a270 (inline replies on each thread). Lanes at this merge point (#80+#81): ruff/format/mypy/lint-imports clean, 2177 unit, 179 adapters+topology. Live: listener lines in `serve.log`, runner untouched by a `[::1]` squatter while role all relocates, and the concurrent-bind race now fails at the rival's bind().

guru-xorciseai (approved):
Re-reviewed at 3b9a270. Good calls on all of these, and you solved the bind properly instead of scoping the docstring, which I didn't expect.

Checked independently rather than trusting the replies:

- **Atomic bind** — confirmed with your actual `bind_listener`: the returned socket has `SO_ACCEPTCONN` set, and a rival `SO_REUSEADDR` bind is now refused at *its* `bind()` with EADDRINUSE, inside `bind_specs` where you report it cleanly. That closes it.
- **The real risk of that change** — uvicorn adopting an already-listening socket — I booted the actual server rather than trusting the unit tests: `up --stub` on an isolated home gives all 6 expected listeners (`127.0.0.1`, `[::1]`, `172.17.0.1` × REST/OTLP), `status` all healthy, `down` clean with 0 left behind. And `up` still relocates past a foreign squatter (HTML 404 on the REST port → `port 3971 (rest) is busy — using 3972 instead`), so the preflight change didn't cost the ordinary case.
- **Startup failures** — the `not server.started` branch is right, and dropping the now-inert `host=`/`port=` from `Config` is a good tidy.
- **Guard resets** — confirmed all four release in their matching shutdown hook. And I'm fine with keeping the separate flags; you're right that the reset sitting next to the claim keeps the rule visible. Not worth a follow-up.
- **Scoped port probes** — `_role_bind_hosts` driven off the same function `apps()` builds from is exactly the shape I was hoping for, and it kills the drift rather than the symptom.
- **Listener lines** — in `serve.log`, both apps, all three addresses, and correctly showing the relocated port:
  ```
  listening on 127.0.0.1:3972, [::1]:3972, 172.17.0.1:3972
  listening on 127.0.0.1:4971, [::1]:4971, 172.17.0.1:4971
  ```
  And your correction is fair: I claimed those six "Uvicorn running on" lines used to land in `serve.log`, which I hadn't actually verified — the code-level point held, the "used to" didn't. The operator ends up better off than before either way.

2161 unit + 179 on this branch, clean on the rest. Approving.

PR #82, opened 2026-09-07
fix(cli): up/down/db upgrade find a live server by pid file OR port, and never orphan or duplicate it

### LINKED ISSUE #72: `xorcise down` reports "stopped" without verifying exit or port release; a failed kill orphans the server permanently
## Summary

`xorcise down` prints "stopped" and exits 0 based on nothing but having *attempted* a kill. It never verifies the process exited or that the REST/OTLP ports were released — and when the kill fails, a `contextlib.suppress` swallows the failure and the pid file is unlinked anyway, **orphaning the server permanently**: no future `up` or `down` can name it again.

## Code (`src/xorcise/core/cli/commands/lifecycle.py`, `down`, ~858–873)

```python
pf = pid_file()
running = pf.exists()
if running:
    with (
        contextlib.suppress(ProcessLookupError, ValueError, PermissionError),
        step_progress("stopping XORCISE"),
    ):
        pid = int(pf.read_text())
        os.kill(pid, signal.SIGTERM)
        if not _await_exit(pid):
            os.kill(pid, signal.SIGKILL)
            _await_exit(pid, timeout=2.0)
    pf.unlink(missing_ok=True)            # ← unconditional, even when the kill failed
    runtime_ports_file().unlink(missing_ok=True)
```

Three distinct problems:

1. **A failed kill is indistinguishable from success.** `PermissionError` (pid alive but not signalable — e.g. a pid file left by another user, or pid reuse landing on a foreign process) aborts the suppressed block *before* anything is stopped; the code then unlinks the pid file and prints the branded "stopped" line, exit 0. The suppress treats "could not kill it" the same as "already dead".
2. **The SIGKILL escalation's outcome is ignored.** The final `_await_exit(pid, timeout=2.0)` return value is discarded — a process stuck in D-state past the timeout is still reported as stopped. (`_await_exit` itself also reads `PermissionError` on `os.kill(pid, 0)` as "exited", lifecycle.py:817.)
3. **Only the pid-file pid is considered, and no port is ever verified.** `down` never checks that the listeners on the REST/OTLP ports are gone. When the pid file has been overwritten by a second `up` (see #73 on the `up` guard), `down` kills the new pid, reports success, and the older server keeps serving.

## Empirical reproduction (isolated `XORCISE_HOME`, stub mode, xorcise 0.1.2.dev9+g1015e710e)

**PermissionError → permanent orphan:**

```
$ xorcise up --stub                      # server pid 1658004, REST :47301
$ echo 1 > $XORCISE_HOME/xorcise.pid     # pid 1 is root-owned → os.kill raises PermissionError
$ xorcise down
stopped — xorcise down                   # exit 0
$ ls $XORCISE_HOME/xorcise.pid
No such file or directory                # pid file gone — nothing can name 1658004 any more
$ ss -tlnp | grep 47301
LISTEN 127.0.0.1:47301 users:(("python3",pid=1658004,...))   # server never stopped
```

**Kills only the pid-file pid (the incident as originally reported — "after my `xorcise down`, PID 7776 was still LISTENing on :3001"):** with an orphaned server on :47301 and a second instance on :47302 (pid file names the second), `xorcise down` reported `stopped`, exit 0 — and `ss` still showed pid 1658004 LISTENing on :47301.

## Expected behaviour

`down` should report success only after verifying the server actually exited (and, ideally, that its recorded runtime ports are released). A kill that fails with `PermissionError`, or a process that survives SIGKILL past the wait, must be reported as a failure **with the pid named and the pid file left in place** — unlinking the pid file on a failed kill converts a transient failure into a permanent orphan.

## Suggested direction

- Stop suppressing `PermissionError` into the success path: keep the pid file, print the pid and a remediation (`kill <pid>` / check ownership), exit non-zero.
- Check the final `_await_exit` result after SIGKILL and report failure when it times out.
- After a confirmed exit, verify the recorded runtime ports (`runtime-ports.json`) are no longer listening before printing "stopped"; if something still listens, say who (the `/api/system` probe used by `doctor`/`_foreign_instance_home` already distinguishes our instance from a foreign one).

### LINKED ISSUE #73: `xorcise up` guards on the pid file only — a live sibling on the REST port becomes a silent second instance over the same DB
## Summary

`xorcise up`'s already-running guard consults **only the pid file** (`lifecycle.py:356–370`). It never asks whether a XORCISE instance already serves the configured REST port. With a stale or absent pid file (which the `down` defect in #72 produces), `up` treats its own live older server as a generic busy port, silently auto-increments to the next port, and boots a **second instance of the same home over the same SQLite database** — after first running `_ensure_db_ready()` against that shared DB.

## Code path (`src/xorcise/core/cli/commands/lifecycle.py`, `up`)

```python
pf = pid_file()                                   # 356: the ONLY liveness signal consulted
if pf.exists():
    try:
        os.kill(int(pf.read_text()), 0)
    except (ProcessLookupError, ValueError):
        pf.unlink(missing_ok=True)                # stale → fall through and boot
    ...
...
_ensure_db_ready()                                # 383: touches the shared DB *before* ports
...
resolved = _resolve_role_ports("all", {...})      # 388: busy port → auto-increment + notice
...
pf.write_text(str(proc.pid))                      # 420: overwrites — old server's identity lost
write_runtime_ports(resolved)                     # 421: overwrites — old ports forgotten
```

Notably, the codebase already has exactly the probe this guard needs: `_foreign_instance_home()` (lifecycle.py:521) asks `/api/system` on the REST port whose home the answering instance belongs to, and `doctor`/`status` use it to distinguish "our running server" from a foreign holder. `up` never calls it.

## Empirical reproduction (isolated `XORCISE_HOME`, stub mode, xorcise 0.1.2.dev9+g1015e710e)

```
# an orphaned instance of THIS home (pid 1658004) serves :47301; pid file absent
$ xorcise up --stub
port 47301 (rest) is busy — using 47302 instead
port 47317 (otlp) is busy — using 47318 instead
xorcise up — UI at http://127.0.0.1:47302/ui      # success banner; the live sibling never mentioned

$ ss -tlnp | grep -E '4730[12]'
LISTEN 127.0.0.1:47301 ... pid=1658004            # old instance
LISTEN 127.0.0.1:47302 ... pid=1659029            # new instance — same home, same xorcise.db
$ cat $XORCISE_HOME/xorcise.pid
1659029                                            # the older server is now unfindable
```

Both instances hold the same `~/.xorcise/xorcise.db`. The pid file and `runtime-ports.json` are overwritten, so every sibling command (`down`, `status`, `ui`) now sees only the new instance — feeding #72, where `down` kills only the pid-file pid and reports "stopped" while the old server keeps serving.

## Why this poisons the old server (the reported incident)

The original report: an orphaned 0.1.2 server (PID 7776) was still serving :3001; a newer build's `up` came through this path and the shared DB ended up migrated under it — `0002_run_provenance` renames the int `mission_version` column to `install_revision` and re-adds `mission_version` as nullable VARCHAR, so the live older server's ORM was suddenly wrong about the table it had open. Two cooperating gaps make that possible:

- `up` at 383 runs `_ensure_db_ready()` regardless of who holds the REST port (in this code version that scaffolds a fresh DB or *refuses* a stale one — but the refusal directs the operator straight to `xorcise db upgrade`, which has no liveness guard of its own; see #74);
- nothing anywhere in `up` establishes "a XORCISE from this home is already serving" — which should be the *idempotent success* case (`already running — UI at …`), not a port-conflict-shrug.

## Expected behaviour

Before auto-incrementing past a busy REST port, `up` should probe it. If the holder is an instance of **this same home** (`/api/system` home == ours), converge idempotently: report "already running", optionally repair the missing pid file / runtime-ports record, and touch nothing. If it is a foreign instance or unknown process, auto-increment is fine — but the DB must not be prepared/migrated until after that determination.

## Suggested direction

- Extend the 356 guard: when the pid file is absent/stale, probe the configured REST port with the existing `_foreign_instance_home` machinery before proceeding; same-home holder ⇒ "already running" (exit 0), recovering the runtime-ports record where possible.
- Move `_ensure_db_ready()` after port resolution + instance detection, so the shared DB is never touched while another instance of this home is live.
- Consider recording the server's pid in `/api/system` so a lost pid file can be reconstructed instead of orphaning the process.

### LINKED ISSUE #74: `xorcise db upgrade` migrates the DB while a live (older) server holds it open — the boot-time guard is one-directional
## Summary

The boot-time DB guard is one-directional. `_ensure_db_ready()` (`lifecycle.py:306–330`) refuses to boot **new code against an old DB** ("database is behind the schema — run 'xorcise db upgrade' first") — but there is no symmetric guard on the other side: `xorcise db upgrade` (`cli/commands/db.py:14–30` → `db/migrate.py:upgrade()`) will migrate the database **while an older live server still holds it open**, silently corrupting that server's view of its own tables. Worse, the refusal message actively *directs* the operator into this hazard while the old server is still running.

## Code

`xorcise db upgrade` is a bare Alembic call with no liveness check of any kind:

```python
@db_app.command("upgrade")
def db_upgrade() -> None:
    """Apply pending database migrations (explicit; never on boot)."""
    try:
        console.print(_upgrade())          # command.upgrade(config, "head") — nothing else
    except CommandError as exc:
        ...                                 # only handles migration-history mismatch
```

Neither `db.py` nor `migrate.upgrade()` consults the pid file, the runtime ports, or the health endpoint before rewriting the schema.

## Why this is destructive, not just untidy

The current migration `0002_run_provenance` uses SQLite **batch mode** on `runs` and `results` — Alembic implements that as a full copy-rebuild-and-swap of each table (`_alembic_tmp_runs` staging table, then rename over the original). Under a live server that means:

- the table the old server has open is replaced wholesale beneath its connection;
- the int `mission_version` column the 0.1.2 ORM maps is **renamed** to `install_revision`, and a new nullable VARCHAR(32) `mission_version` takes its name — so the old server's subsequent reads/writes hit a column whose type and meaning changed under it.

This is the mechanism that poisoned the orphaned 0.1.2 server (PID 7776) in the original incident: #73 got a second instance + `db upgrade` invoked while 7776 still served :3001.

## Empirical verification (isolated `XORCISE_HOME`, stub mode, xorcise 0.1.2.dev9+g1015e710e)

With **two live server instances** of the same home holding `xorcise.db` open (from the #73 reproduction), and the DB stamped back to `0001_initial`:

```
$ xorcise db upgrade        # both servers alive and serving
unexpected error: A column with name 'install_revision' is already present in table '_alembic_tmp_runs'. …
```

The failure itself is an artifact of the contrived setup (the schema was already at head; only the stamp was rewound) — the point is what it proves: **the command went straight into executing the batch table rebuild against the live DB with zero refusal or even warning that two running servers held it.** No "XORCISE is running — stop it first (xorcise down)", nothing.

## Compounding detail

`_ensure_db_ready()`'s refusal is the *only* guard in the system, and its remediation line —

> `database is behind the schema — run 'xorcise db upgrade' first`

— is printed exactly in the situation where an older server may still be live (that is *why* the DB is behind and the port is busy). Following the instruction verbatim migrates the DB under that server.

## Expected behaviour

`xorcise db upgrade` should refuse (or require an explicit `--force`) while a server appears to be using this home's database: pid file naming a live pid, or a health probe answering on the recorded runtime ports whose `/api/system` home matches this home. The refusal should say how to proceed: `xorcise down`, then upgrade, then `up`.

## Suggested direction

- Add a liveness pre-check to `db_upgrade` reusing the existing machinery: pid-file check (as in `up`'s guard) + `_foreign_instance_home`-style `/api/system` probe on the runtime-resolved REST port, refusing when the answering instance's home is this one.
- Adjust `_ensure_db_ready()`'s stale-DB message to include stopping the old server first (e.g. "run 'xorcise down' then 'xorcise db upgrade'"), so the one guard that fires doesn't route operators into the unguarded path.
- Optionally take a short-lived exclusive SQLite lock (`BEGIN IMMEDIATE` on a sentinel) before migrating, as a backstop against holders the probes can't see.

christo-xorciseai:
> Stacked on #81 (`fix/boot-singletons`), which is stacked on #80. Review the diff against `fix/boot-singletons`; merge in order.

## What changed

`up`, `down` and `db upgrade` now agree on one question — *is a server of this home running?* — and answer it from more than the pid file (`lifecycle._live_instance`): the pid file when it names a live process, otherwise `/api/system` on every REST port an instance of this home could be on (the configured port, the runtime record's port, `--port`). `/api/system` gains a `pid` field so a found instance can be stopped or re-recorded.

- **`down` (#72)** — `_stop_process` reports what actually happened (`stopped` / `gone` / `unsignalable` / `survived`) instead of `contextlib.suppress`-ing `PermissionError` into "stopped". A kill that is refused or a process that outlives SIGKILL is a **failure**: the pid named, the pid file **left in place**, exit 1. An unsignalable pid that no instance of this home claims is pid reuse: the stale record is dropped. After the recorded pid, `down` **sweeps the ports** and stops any instance of this home still answering, by the pid it reports (an older server that cannot report one is refused with an `ss -tlnp` hint, never silently left behind). "stopped" is claimed only after checking the ports are released; a foreign holder is named in one line, not treated as a failure.
- **`up` (#73)** — before touching anything, `_live_instance` runs. An instance of this home found by probe (pid file gone or stale) is **adopted**: pid file and runtime-ports record are repaired from what it reports, "already running (pid record repaired)", exit 0. No second instance, and `_ensure_db_ready` is never reached while a sibling holds the DB. A foreign home's instance still auto-increments as before.
- **`db upgrade` (#74)** — refuses while an instance of this home is live (by pid or by port), naming it and the sequence `xorcise down` → `xorcise db upgrade` → `xorcise up`; `--force` overrides. The boot-time stale-DB refusal now says to stop the server *first*, instead of routing operators straight into the unguarded path.
- `SystemInfo.pid` (contracts) / `build_system_info` sets `os.getpid()`. Additive, optional.

## Why

Closes #72, closes #73, closes #74.

One incident, three gaps: a `down` whose kill failed unlinked the pid file and printed "stopped" (the server kept serving :3001, now unfindable); the next `up` saw only a busy port, auto-incremented and booted a second instance of the same home over the same SQLite file; `db upgrade` then rebuilt the `runs`/`results` tables (SQLite batch mode: copy, swap, rename `mission_version` → `install_revision`) underneath the older server, whose ORM was suddenly wrong about tables it had open.

**Not in this PR:** the optional `BEGIN IMMEDIATE` sentinel lock suggested in #74 — the liveness checks cover the reported path; a lock would be a second line of defence with SQLite-specific semantics worth its own change. Also the checked-in `frontend/openapi.json` / `schema.ts` snapshot: regenerating it here produced formatting noise unrelated to the one added field (whatever produced the current file is not `json.dump`), so it is left for the owner of that codegen step; the frontend does not read `pid`.

## Testing

```
$ uv run ruff check . && uv run ruff format --check . && uv run mypy && uv run lint-imports
All checks passed! / already formatted / Success: no issues found in 492 source files / Contracts: 4 kept, 0 broken.
$ uv run pytest -m unit -n auto -q          → 2177 passed
$ uv run pytest -m "adapters or topology" -q → 179 passed
```

New: `tests/unit/test_lifecycle_guards.py` (15 tests — refused kill keeps the pid file and exits 1; unsignalable pid with nothing behind it is stale; survives-SIGKILL fails; orphan the pid file does not name is found by port and stopped; old server that cannot name its pid is refused with the `ss` hint; foreign holder is named without failing; `up` adopts a live instance when the pid file is gone / stale, repairs both records, asks configured + recorded + `--port`, and still boots past a foreign instance; `db upgrade` refuses by pid, refuses by port, `--force` overrides, runs when nothing is live; the stale-DB message names `xorcise down`). `test_system_rest.py` asserts `pid`. One existing `down` test re-seamed (its `os.kill` stub never let the process "exit", which the verified stop now correctly reports).

**Empirical — the real CLI (`up --stub` / `down` / `db upgrade`, isolated `XORCISE_HOME`, stub mode) from the previous tree and from this branch:**

| Scenario | Before | After |
|---|---|---|
| **A** `up`; pid file overwritten with `1` (root-owned → `PermissionError`); `down` | exit 0, "stopped", pid file **gone**, server **alive**, 3 listeners still up — permanently orphaned | `pid 1 … is not XORCISE (pid reuse) — dropping the stale record`, then `stopped an orphaned instance of this home (pid 1107211) that was still serving port 57619`; server gone, 0 listeners, exit 0 |
| **B** `up`; pid file deleted; `up` again | second instance booted on port+1 (`… is busy — using …`), pid file now names it, **2 serve processes** of one home on one DB | `already running (pid record repaired) — UI at …`; pid file restored to the first pid, **1 process**, nothing on port+1 |
| **C** `up`; `db upgrade` with the server live | exit 0, `migrations: upgraded to head` — migrated under a live server, no warning | exit 1: `XORCISE is running (pid 1110919) and holds this database open … Stop it first: xorcise down, then xorcise db upgrade, then xorcise up`; `--force` proceeds; `down` clean |
| **D** two instances of one home (orphan on the configured port, second in the pid file); `down` | exit 0 "stopped", orphan **alive** and serving | both built from this branch: `stopped an orphaned instance of this home (pid …) that was still serving port …`, both gone, both ports released, exit 0. Orphan built from the **old** tree (no `pid` in `/api/system`): exit 1, `an instance of this home still answers on port … but did not report its pid (an older server) — find it with ss -tlnp …` — refused honestly rather than reported stopped |

One thing the first run of this experiment caught: `/api/system` takes ~1.9 s cold (remote catalog probes on first hit) versus 0.15 s warm, so a 1 s probe timeout read a live server as "not ours". The instance probe uses an 8 s timeout (nothing listening still fails instantly on connection refused); `_foreign_instance_home` shares it.

**Test lanes affected**

- [x] `unit`
- [ ] `adapters`
- [ ] `topology`
- [x] `integration` — `up`/`down` lifecycle; validated by hand above
- [x] `e2e` — full `xorcise up`/`down`; please add `full-ci`
- [ ] `frontend`

## User-facing impact

`xorcise db upgrade --force` (new flag); `GET /api/system` gains optional `pid`. Behaviour changes: `down` can exit 1 where it used to print "stopped" over a live server; `up` can print "already running (pid record repaired)" instead of booting a duplicate; `db upgrade` refuses while the server runs.

## Documentation

- [x] Documentation needed and tracked in a follow-up issue (link it) — the `db upgrade --force` flag and the new `down` failure mode belong on the CLI reference page (docs site is a separate repo); noting on #74.

## Dependencies

- [x] No dependency changes

## Checklist

- [x] No secrets, credentials, tokens, keys, internal hostnames or customer data in the diff
- [x] No references to internal-only systems

---

### DIFF
--- a/src/xorcise/core/cli/commands/db.py
+++ b/src/xorcise/core/cli/commands/db.py
@@ -30,10 +30,24 @@ def db_upgrade(force: bool = _FORCE_OPTION) -> None:
     """
     from alembic.util.exc import CommandError
 
-    from xorcise.core.cli.commands.lifecycle import _live_instance
+    from xorcise.core.cli.commands.lifecycle import InstanceUndetermined, _live_instance
     from xorcise.core.config import get_settings
 
-    live = None if force else _live_instance(get_settings())
+    # `is True`: a direct (non-CLI) call gets the typer OptionInfo as the default, and that
+    # object is truthy — `if force:` would skip the guard that stops database corruption.
+    if force is True:
+        live = None
+    else:
+        try:
+            live = _live_instance(get_settings())
+        except InstanceUndetermined as exc:
+            # Cannot tell whether a server holds the DB: the safe answer is no migration.
+            err_console.print(
+                f"[err]error[/err]: {exc}. Not migrating while that is unknown "
+                "(--force overrides; unsafe.)",
+                highlight=False,
+            )
+            raise typer.Exit(1) from None
     if live is not None:
         where = f"pid {live.pid}" if live.pid is not None else f"port {live.rest_port}"
         err_console.print(
--- a/src/xorcise/core/cli/commands/db.py
+++ b/src/xorcise/core/cli/commands/db.py
@@ -11,10 +11,38 @@ db_app = typer.Typer(help="Database maintenance (advanced).", no_args_is_help=Tr
 app.add_typer(db_app, name="db", rich_help_panel="Advanced")
 
 
+_FORCE_OPTION = typer.Option(
+    False,
+    "--force",
+    help="Migrate even if a XORCISE server of this home appears to be running (unsafe).",
+)
+
+
 @db_app.command("upgrade")
-def db_upgrade() -> None:
-    """Apply pending database migrations (explicit; never on boot)."""
+def db_upgrade(force: bool = _FORCE_OPTION) -> None:
+    """Apply pending database migrations (explicit; never on boot).
+
+    Refuses while a server of this home is running: the current migrations rebuild tables in
+    place (SQLite batch mode — copy, swap, rename), so migrating under a live server replaces
+    the tables it has open and renames columns its ORM is still using. The boot-time guard
+    only ever pointed HERE ("run 'xorcise db upgrade' first"), and it fires precisely when an
+    older server may still be up — so this end has to hold the line too (#74).
+    """
     from alembic.util.exc import CommandError
+
+    from xorcise.core.cli.commands.lifecycle import _live_instance
+    from xorcise.core.config import get_settings
+
+    live = None if force else _live_instance(get_settings())
+    if live is not None:
+        where = f"pid {live.pid}" if live.pid is not None else f"port {live.rest_port}"
+        err_console.print(
+            f"[err]error[/err]: XORCISE is running ({where}) and holds this database open — "
+            "migrating it underneath a live server rewrites the tables it is using. "
+            "Stop it first: [value]xorcise down[/value], then [value]xorcise db upgrade[/value], "
+            "then [value]xorcise up[/value]. (--force overrides; unsafe.)"
+        )
+        raise typer.Exit(1)
 
     try:
         console.print(_upgrade())
--- a/src/xorcise/core/cli/commands/lifecycle.py
+++ b/src/xorcise/core/cli/commands/lifecycle.py
@@ -12,7 +12,7 @@ import signal
 import subprocess
 import sys
 import time
-from dataclasses import asdict, replace
+from dataclasses import asdict, dataclass, replace
 from pathlib import Path
 
 import httpx
@@ -339,8 +339,13 @@ def _ensure_db_ready() -> None:
         db.upgrade()
         console.print("prepared the database")
     elif state == "stale":
+        # Name the WHOLE sequence. This message fires exactly when an older server may still be
+        # serving (that is why the DB is behind), and `db upgrade` alone would migrate the tables
+        # underneath it — the stop comes first.
         err_console.print(
-            "[err]database is behind the schema — run 'xorcise db upgrade' first[/err]"
+            "[err]database is behind the schema[/err] — stop any running XORCISE first "
+            "([value]xorcise down[/value]), then run [value]xorcise db upgrade[/value] and "
+            "start it again"
         )
         raise typer.Exit(1)
 
@@ -369,20 +374,22 @@ def up(
     Already-running is success, not an error — safe to re-run any time.
     """
     pf = pid_file()
-    if pf.exists():
-        try:
-            os.kill(int(pf.read_text()), 0)
-        except (ProcessLookupError, ValueError):
-            pf.unlink(missing_ok=True)
-        except PermissionError:
-            # The pid exists but belongs to another user — treat as running.
-            console.print(f"already running — UI at {_effective_ui_url()}")
-            return
+    live = _live_instance(get_settings(), extra_ports=[port] if port else [])
+    if live is not None:
+        if live.source == "probe":
+            # An instance of THIS home is serving but the pid record is gone or names a dead
+            # process (a `down` whose kill failed, an earlier `up` that overwrote it). Booting
+            # here would start a SECOND server over the same database — and migrate it under
+            # the first. Adopt the live one instead: repair the record so `down`/`status`/`ui`
+            # can find it again, and converge.
+            _repair_instance_record(live)
+            console.print(f"already running (pid record repaired) — UI at {_effective_ui_url()}")
         else:
             # Idempotent converge: 'already true' is exit 0, matching `down` on
             # not-running — scripts and agents re-run `up` defensively.
             console.print(f"already running — UI at {_effective_ui_url()}")
-            return
+        return
+    pf.unlink(missing_ok=True)  # absent, corrupt, or names a process that is gone
 
     # Prerequisites BEFORE anything is written. A missing Compose v2 plugin used to
     # surface ~20 lines in, after the home was created, the DB migrated and two RSA
@@ -538,7 +545,9 @@ def _foreign_instance_home(host: str, rest_port: int) -> str | None:
     this home's — else None. Lets doctor say 'held by another XORCISE instance'
     instead of accusing an unknown process."""
     try:
-        info = httpx.get(f"http://{host}:{rest_port}/api/system", timeout=1).json()
+        info = httpx.get(
+            f"http://{host}:{rest_port}/api/system", timeout=_INSTANCE_PROBE_TIMEOUT
+        ).json()
     except (httpx.HTTPError, ValueError):
         return None
     their_home = str(info.get("home") or "") if isinstance(info, dict) else ""
@@ -548,6 +557,153 @@ def _foreign_instance_home(host: str, rest_port: int) -> str | None:
         if Path(their_home).resolve() == Path(xorcise_home()).resolve():
             return None
     return their_home
+
+
+@dataclass(frozen=True)
+class LiveInstance:
+    """A running server of THIS home: how it was found, where it listens, and its pid if known.
+
+    `source` is "pidfile" when the pid file names a live process, "probe" when the pid file was
+    absent/stale but an instance of this home answered on a REST port this home would use —
+    the shape a failed `down` or an overwriting second `up` leaves behind (#72, #73)."""
+
+    pid: int | None
+    rest_port: int
+    otlp_port: int | None
+    source: str
+
+
+def _pid_from_file() -> int | None:
+    """The pid the pid file names, when that process is alive; None when the file is absent,
+    corrupt, or names a process that is gone. A pid that exists but cannot be signalled
+    (another user's) still counts as alive — it is not ours to declare dead."""
+    pf = pid_file()
+    if not pf.exists():
+        return None
+    try:
+        pid = int(pf.read_text())
+    except ValueError:
+        return None
+    try:
+        os.kill(pid, 0)
+    except ProcessLookupError:
+        return None
+    except PermissionError:
+        return pid
+    return pid
+
+
+# `/api/system` is cheap when warm but its FIRST answer waits on remote catalog probes (measured
+# well over a second cold), and a timeout here reads as "not ours" — which is how a live server
+# got dropped as pid reuse. Nothing listening still fails instantly (connection refused).
+_INSTANCE_PROBE_TIMEOUT = 8.0
+
+
+def _same_home_instance_on(host: str, rest_port: int) -> dict[str, object] | None:
+    """`/api/system` of whatever answers on `host:rest_port`, when it is an instance of THIS
+    home — else None (nothing there, not XORCISE, or another home's instance)."""
+    try:
+        info = httpx.get(
+            f"http://{host}:{rest_port}/api/system", timeout=_INSTANCE_PROBE_TIMEOUT
+        ).json()
+        if not isinstance(info, dict):
+            return None
+        their_home = str(info.get("home") or "")
+        if not their_home:
+            return None
+        same = Path(their_home).resolve() == Path(xorcise_home()).resolve()
+    except Exception:  # noqa: BLE001 — a best-effort probe; anything odd means "not ours"
+        return None
+    return info if same else None
+
+
+def _candidate_rest_ports(settings, extra: list[int] = []) -> list[int]:  # type: ignore[no-untyped-def]  # noqa: B006
+    """Where a server of this home could be listening: the configured REST port, the one the
+    runtime record names (an auto-incremented instance) and any the caller was told about
+    (`up --port N`), configured first, without duplicates."""
+    ports = [settings.rest_port]
+    recorded = (read_runtime_ports() or {}).get("rest")
+    for candidate in (recorded, *extra):
+        if isinstance(candidate, int) and candidate not in ports:
+            ports.append(candidate)
+    return ports
+
+
+def _otlp_from_info(info: dict[str, object]) -> int | None:
+    """The OTLP port an instance reports in its `/api/system` planes, if it does."""
+    planes = info.get("planes")
+    if not isinstance(planes, list):
+        return None
+    for plane in planes:
+        if isinstance(plane, dict) and plane.get("name") == "otlp":
+            _host, _, port = str(plane.get("location") or "").rpartition(":")
+            if port.isdigit():
+                return int(port)
+    return None
+
+
+def _live_instance(settings, extra_ports: list[int] = []) -> LiveInstance | None:  # type: ignore[no-untyped-def]  # noqa: B006
+    """The running server of THIS home, if any: by pid file first, then by asking the REST
+    port(s) this home would use whether an instance of this home answers.
+
+    The pid file alone is not enough — `up` used to guard on it only, so once it was stale or
+    gone (a failed `down`, see #72) a live sibling on the REST port read as a generic busy port,
+    `up` auto-incremented past it and booted a second instance over the same SQLite file (#73).
+    """
+    pid = _pid_from_file()
+    if pid is not None:
+        recorded = read_runtime_ports() or {}
+        return LiveInstance(
+            pid, recorded.get("rest", settings.rest_port), recorded.get("otlp"), "pidfile"
+        )
+    for port in _candidate_rest_ports(settings, extra_ports):
+        info = _same_home_instance_on(settings.host, port)
+        if info is not None:
+            reported = info.get("pid")
+            return LiveInstance(
+                reported if isinstance(reported, int) else None,
+                port,
+                _otlp_from_info(info),
+                "probe",
+            )
+    return None
+
+
+def _repair_instance_record(live: LiveInstance) -> None:
+    """Re-create the pid file and runtime-ports record for an instance found by probe, so the
+    sibling commands can name it again."""
+    if live.pid is not None:
+        pid_file().write_text(str(live.pid))
+    ports = {"rest": live.rest_port}
+    if live.otlp_port is not None:
+        ports["otlp"] = live.otlp_port
+    write_runtime_ports(ports)
+
+
+def _stop_process(pid: int) -> str:
+    """SIGTERM `pid`, escalate to SIGKILL, and say what actually happened:
+
+    "stopped" — it exited; "gone" — it was already gone; "unsignalable" — it is alive but not
+    ours to signal (another user's process, or pid reuse); "survived" — it outlived SIGKILL.
+    `down` used to swallow the last two into "stopped" and unlink the pid file regardless,
+    which turned a failed kill into a permanently orphaned server (#72)."""
+    try:
+        os.kill(pid, signal.SIGTERM)
+    except ProcessLookupError:
+        return "gone"
+    except PermissionError:
+        return "unsignalable"
+    # Wait for it to actually exit (release its ports) so a following `up` isn't racing a
+    # still-terminating server; escalate to SIGKILL if it won't stop.
+    if _await_exit(pid):
+        return "stopped"
+    try:
+        os.kill(pid, signal.SIGKILL)
+    except ProcessLookupError:
+        return "stopped"
+    except PermissionError:
+        return "unsignalable"
+    return "stopped" if _await_exit(pid, timeout=2.0) else "survived"
 
 
 def _note_foreign_instance(host: str, rest_port: int) -> None:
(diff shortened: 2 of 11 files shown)


src/xorcise/core/cli/commands/lifecycle.py
```diff
@@ -855,28 +1011,102 @@ def down(
...
+            )
+            raise typer.Exit(1)
         pf.unlink(missing_ok=True)
         runtime_ports_file().unlink(missing_ok=True)  # the record dies with the server
+    # The pid file names ONE process. An instance of this home can be serving without being in
+    # it — a second `up` overwrote the record (#73), or the record was lost — so ask the ports
+    # this home would use, and stop any instance of this home that still answers, by the pid it
+    # reports. A foreign home's instance is left alone (and named). Bounded: one per candidate
+    # port, never a loop on something we cannot stop.
+    for port in _candidate_rest_ports(settings):
```
guru-xorciseai: This is the blocker for me. `_candidate_rest_ports` reads `runtime-ports.json`, but we unlink that file six lines up at 1064 — so by the time the sweep runs, the recorded port is gone and this only ever looks at the configured one.

Which means the instance that auto-incremented to 3002 — i.e. exactly the #73 scenario this PR exists to handle — is never probed. And because the release check at 1098 also only asks `settings.rest_port`, we print "stopped" and exit 0 over a server that's still serving. That's #72 again, just reached differently.

Repro against this branch, using your own `home` fixture:

```python
(home / "xorcise.pid").write_text("999999")   # stale: dead pid
(home / "runtime-ports.json").write_text('{"rest": 3002, "otlp": 4319}')
# only 3002 answers, reporting a live pid 4242
monkeypatch.setattr(lifecycle, "_same_home_instance_on",
                    lambda h, p: _instance(4242, rest=3002) if p == 3002 else None)
runner.invoke(app, ["down"])
```

gives:

```
OUTPUT: stopped — xorcise down
EXIT:   0     KILLED: []
```

`os.kill` never called, orphan still up, exit 0.

`test_down_stops_an_orphaned_instance_the_pid_file_does_not_name` passes only because its stub answers on the configured 3001, so deleting the record never shows up.

Fix looks like computing the candidate list once at the top of `down`, before anything is unlinked, and verifying release against that same list rather than `{settings.rest_port, settings.otlp_port}`.

src/xorcise/core/cli/commands/db.py
```diff
@@ -11,11 +11,39 @@
...
+    the tables it has open and renames columns its ORM is still using. The boot-time guard
+    only ever pointed HERE ("run 'xorcise db upgrade' first"), and it fires precisely when an
+    older server may still be up — so this end has to hold the line too (#74).
+    """
     from alembic.util.exc import CommandError
 
+    from xorcise.core.cli.commands.lifecycle import _live_instance
+    from xorcise.core.config import get_settings
+
+    live = None if force else _live_instance(get_settings())
```
guru-xorciseai: `force` needs to be `force is True`. Its default is `_FORCE_OPTION`, and `bool(typer.Option(False, "--force"))` is `True` — so any direct `db_upgrade()` call (a test, a script, a future internal caller) skips the liveness check entirely and migrates under a live server.

This codebase already knows the trap and defends against it in two places — `status` has ``if as_json is True:  # `is True` guards direct (non-CLI) calls where the typer default is an OptionInfo``, and `_resolve_role_ports` uses `isinstance(override, int)`.

Nothing calls it that way today, so it isn't urgent, but the failure direction is fail-open on the guard that stops database corruption, so I'd rather it were airtight.

src/xorcise/core/cli/commands/lifecycle.py
```diff
@@ -535,6 +544,153 @@ def _foreign_instance_home(host: str, rest_port: int) -> str | None:
...
+# well over a second cold), and a timeout here reads as "not ours" — which is how a live server
+# got dropped as pid reuse. Nothing listening still fails instantly (connection refused).
+_INSTANCE_PROBE_TIMEOUT = 8.0
+
+
+def _same_home_instance_on(host: str, rest_port: int) -> dict[str, object] | None:
+    """`/api/system` of whatever answers on `host:rest_port`, when it is an instance of THIS
+    home — else None (nothing there, not XORCISE, or another home's instance)."""
+    try:
+        info = httpx.get(
```
guru-xorciseai: `httpx.get` defaults to `trust_env=True`, and httpx applies no implicit localhost bypass — I checked against the pinned 0.28.1:

```
loopback, HTTP_PROXY set  -> pool=HTTPProxy
loopback, no proxy env    -> pool=ConnectionPool
loopback, HTTP_PROXY + NO_PROXY=127.0.0.1 -> pool=ConnectionPool
```

So in a dev container or a corporate shell with `HTTP_PROXY`/`ALL_PROXY` set and `NO_PROXY` not covering loopback, this probe goes to the proxy, comes back 502 or times out, the bare `except Exception` at 600 turns that into "not ours", and `up` decides nothing is running — then auto-increments past the live sibling and boots the second instance over the same SQLite file. `db upgrade` proceeds too. The exact failure the PR is here to prevent, and it fails silently.

`trust_env=False` on the loopback probes is the fix.

While you're there: the bare except also can't tell "connection refused" (nothing is there) from "something went wrong" (we don't know). For a guard this load-bearing, treating the second as unknown rather than absent would let it fail closed. The 8s timeout justified in the comment above makes that worse, not better — 8s of proxy stall now reads as "nothing running".

src/xorcise/core/cli/commands/lifecycle.py
```diff
@@ -354,20 +359,22 @@ def up(
...
+    live = _live_instance(get_settings(), extra_ports=[port] if port else [])
+    if live is not None:
+        if live.source == "probe":
+            # An instance of THIS home is serving but the pid record is gone or names a dead
+            # process (a `down` whose kill failed, an earlier `up` that overwrote it). Booting
+            # here would start a SECOND server over the same database — and migrate it under
+            # the first. Adopt the live one instead: repair the record so `down`/`status`/`ui`
+            # can find it again, and converge.
+            _repair_instance_record(live)
+            console.print(f"already running (pid record repaired) — UI at {_effective_ui_url()}")
```
guru-xorciseai: When the probed instance reports no pid — the older-server case from your scenario D — `_repair_instance_record` writes no pid file (`if live.pid is not None`), so nothing is actually repaired. But we still say "pid record repaired".

And it compounds: `_runtime_port_overlay` is gated on `pid_file().exists()`, so with no pid file written `_effective_ui_url()` falls back to the configured port and we advertise `http://127.0.0.1:3001/ui` for an instance we just found on, say, 3005. Confident message, dead URL, and `down`/`status`/`ui` still can't find it — so every subsequent `up` repeats the claim.

Probably wants its own branch: adopted, but the pid is unknown, so say that and don't claim the record was fixed.

src/xorcise/core/cli/commands/lifecycle.py
```diff
@@ -855,28 +1011,102 @@ def down(
...
+    # it — a second `up` overwrote the record (#73), or the record was lost — so ask the ports
+    # this home would use, and stop any instance of this home that still answers, by the pid it
+    # reports. A foreign home's instance is left alone (and named). Bounded: one per candidate
+    # port, never a loop on something we cannot stop.
+    for port in _candidate_rest_ports(settings):
+        info = _same_home_instance_on(settings.host, port)
+        if info is None:
+            continue
+        orphan_pid = info.get("pid")
+        if not isinstance(orphan_pid, int) or orphan_pid == os.getpid():
```
guru-xorciseai: Two unrelated conditions share this message, and the second one gets told the wrong thing. If `orphan_pid == os.getpid()` fires, the instance *did* report a pid — it's us — so "did not report its pid (an older server)" sends the operator to `ss -tlnp` to hunt for the CLI they just ran.

Minor, but worth splitting since the "older server" line is otherwise a genuinely good message.

Also on this block, and the reason I'd want a second look at `_stop_process` generally: the pid now arrives from a `/api/system` JSON body, and `isinstance(orphan_pid, int)` is the only check before it reaches `os.kill`. `isinstance(True, int)` is True, so `"pid": true` gives `os.kill(1, ...)`, and `"pid": 0` gives `os.kill(0, SIGTERM)` — which POSIX defines as *every process in the caller's process group*, i.e. `down` and the operator's shell pipeline. A `0 < pid < 2**31` check before any signal would close that off. (Same for the pid file side: a huge value raises `OverflowError`, which `except ValueError` doesn't catch, so `up`/`down`/`db upgrade` all abort with a raw traceback.)

guru-xorciseai (changes requested):
Pulled the whole stack and ran each merge point separately, since these land in order and main sits at each one for a while:

| | ruff / format / mypy / imports | unit | adapters+topology |
|---|---|---|---|
| #80 alone | clean | 2150 | 179 |
| #80+#81 | clean | 2161 | 179 |
| all three | clean | 2177 | 179 |

Matches your numbers exactly, and no merge point is red — so the ordering is fine and nothing here breaks the existing suite.

The `_live_instance` idea is right, and the empirical table (especially scenario D with the old binary as the orphan) is the kind of evidence I wish every PR had. But I think the `down` path has reintroduced #72 through a different door, and I've got a repro for it — that's the one blocker. Two fail-open bits alongside it.

**One more, not inline because the test itself is untouched:** `test_db_upgrade_history_mismatch_is_a_targeted_error` (`tests/unit/test_cli_ux.py:116`) sets no `XORCISE_HOME` and stubs no probe, and there's no autouse home isolation in `tests/conftest.py` — only the opt-in `migrated_home`. So the new guard reads the developer's real `~/.xorcise/xorcise.pid`. With a server running locally:

```
$ pytest tests/unit/test_cli_ux.py::test_db_upgrade_history_mismatch_is_a_targeted_error
E  AssertionError: assert 'migration history mismatch' in
   'error: XORCISE is running (pid 216735) and holds this database open ...'
```

Green in CI, red for anyone who happens to have `xorcise up` running. Same shape for the three `down` tests in `test_cli_commands.py` (~741-776), which stub neither `_same_home_instance_on` nor `ports_in_use` and so now make real HTTP calls with an 8s timeout in the unit lane. Your new `test_lifecycle_guards.py` `home` fixture stubs all three, which is what these need.

src/xorcise/core/cli/commands/lifecycle.py
christo-xorciseai: Confirmed with your repro (exit 0, `os.kill` never called) and fixed in 54c983c exactly as you describe: `down` reads the candidate ports and the runtime record before anything is unlinked, the sweep uses that list, and the release check covers the recorded ports too. Your repro is now `test_down_probes_the_recorded_port_even_though_it_unlinks_the_record` (orphan on 3002 killed, exit 0), plus `test_down_verifies_release_on_the_recorded_ports_too`. Also verified live: a foreign web service on the configured port made `up` relocate, the pid file was made stale, and `down` found and stopped the relocated instance through the record while naming the foreign holder.

src/xorcise/core/cli/commands/db.py
christo-xorciseai: Fixed in 54c983c: `if force is True:` (with the comment naming the trap, as `status` does). `test_db_upgrade_guard_holds_for_a_direct_call` calls `db_upgrade()` directly with a live pid file and asserts the migration never runs.

src/xorcise/core/cli/commands/lifecycle.py
christo-xorciseai: Fixed in 54c983c. `trust_env=False` on every loopback probe (`_probe_system`, `_foreign_instance_home`, `_note_foreign_instance`, and `up`'s health poll). Reproduced first against a live stub server: with `HTTP_PROXY=http://127.0.0.1:9` the probe returned "not ours" in 0.06 s and `up` with a lost pid file booted a duplicate; after the fix the same run adopts the instance. And the bare except is gone: `_probe_system` distinguishes connection refused (absent) from timeout / transport error / 5xx (UNDETERMINED), and `up`, `down` and `db upgrade` fail CLOSED on undetermined — refusing with the port named and what to do (`InstanceUndetermined`). A non-JSON answer (an HTML 404 from some other web service) is foreign, so auto-increment past an unrelated service still works. Verified live with a socket that accepts and never answers: all three refuse; `--force` still migrates. Tests: `test_probes_never_use_the_proxy_environment`, `test_probe_distinguishes_refused_from_undetermined`, `test_an_undetermined_port_fails_{up,down,db_upgrade}_closed`.

src/xorcise/core/cli/commands/lifecycle.py
christo-xorciseai: Fixed in 54c983c with its own branch: when the instance reports no pid, `up` says `already running on port N — UI at http://…:N/ui. Its pid is not known (an older server), so the pid file could not be restored; 'xorcise down' will find it by port`, the URL is built from the port it was FOUND on, only the port record is written, and "repaired" is not claimed. Verified live against a simulated old server (`/api/system` with `home` but no `pid`) found on a relocated port. `test_up_adopting_an_instance_without_a_pid_says_so_and_names_the_right_port`.

src/xorcise/core/cli/commands/lifecycle.py
christo-xorciseai: Fixed in 54c983c. The two conditions have their own messages (the `os.getpid()` one now says the body named this process and refuses to signal it). Every pid that arrives from a file or a JSON body goes through `_valid_pid` — an int that is not a bool, `0 < pid < 2**31` — before it can reach `os.kill`, in `_pid_from_file`, `_live_instance`, and the sweep; a pid-file value past C long reads as stale instead of an OverflowError traceback. Tests: `test_pids_from_a_json_body_are_validated_before_any_signal` (parametrised over `True`, `0`, `-1`, `2**40`, `"4242"`, `None`: nothing signalled, exit 1) and `test_a_pid_file_past_c_long_is_stale_not_a_traceback`.

christo-xorciseai:
Review follow-ups landed in 54c983c (inline replies on each thread), including the non-inline one: an autouse fixture in `tests/conftest.py` now stubs the instance probe for the whole unit lane (tests that exercise it override it), and `test_db_upgrade_history_mismatch_is_a_targeted_error` gets its own `XORCISE_HOME` — I reproduced your failure by pointing `HOME` at a fake home holding a live pid before fixing it. Lanes at each merge point after the follow-ups: #80 alone 2161 unit / 179; #80+#81 2177 / 179; all three 2209 / 179; ruff, format, mypy strict and lint-imports clean at each. Live re-validation on this branch: your blocker repro (relocated instance found and stopped), proxy env with a lost pid file (adopted, no duplicate), a hung port (up/down/db upgrade refuse; `--force` migrates), an old-style instance without a pid (honest message, right port).

guru-xorciseai (approved):
Re-reviewed at 54c983c. The blocker is genuinely fixed — `candidates`, `recorded` and `release_ports` are all read at the top of `down` before anything is unlinked, the sweep uses that list, and the release check covers the recorded ports too.

I re-ran my original repro unchanged against this head: **passes** (orphan on the recorded 3002 signalled, exit 0). The `force is True` repro passes too. Both were failing on 64ac167.

Everything else checked out:

- **`trust_env=False`** on all four loopback probes, confirmed. The undetermined/fail-closed work goes further than I asked and I think it's right — distinguishing `ConnectError` (absent) from timeout/transport/5xx (unknown) is the correct axis, and fail-closed is the only safe direction for a guard whose failure mode is a duplicate instance on one SQLite file. `InstanceUndetermined`'s message names the port and gives both ways out, which is what makes the refusal usable rather than annoying.
- **Non-JSON is foreign** — verified live: a squatter answering an HTML 404 on the REST port still lets `up` auto-increment past it. That was the case I was most worried about breaking, so thanks for handling it.
- **`_valid_pid`** covers file and JSON on both sides, and the `os.getpid()` case has its own message now.
- **Adoption without a pid** — the port it was found on, only the port record written, no "repaired" claim. Reads honestly.
- **Test isolation** — the autouse `_no_loopback_instance_probe` fixes it properly. I re-ran the test that was failing with a fake `HOME` naming a live pid: passes now, as do the three `down` tests that were making real network calls.

Full gate on the stack tip: 2209 unit, 179 adapters+topology, ruff/format/mypy strict/lint-imports all clean, and every merge point green on its own (2161 / 2177 / 2209) so the merge order is safe.

Two small things, neither worth holding this up:

1. The candidate ports are still probed sequentially at 8s each, and `_live_instance` runs before any progress output in `up` — so a worst case of three stalling candidates is ~24s of total silence before the refusal. A spinner, or probing them concurrently, would make that legible.
2. Treating any 5xx as undetermined means a *foreign* service that's merely broken on the configured port (an nginx with a dead upstream on 3001) now blocks `up` where it used to relocate. Probably the right default, but if it bites, restricting the 5xx-is-undetermined rule to the port the runtime record names — where one of ours is actually expected — would narrow it without losing the safety.

Approving. Nice work on the turnaround, and thanks for reproducing each one before fixing it — made this pass much quicker.

Issue #76, opened 2026-08-25
[Feature]: xorcise-openeval-adapter — export mission checks/rubric + GradeResult to EvalPort's portable format


adhabnr-ux:
## What problem are you trying to solve?

XORCISE's grading output (`GradeResult` — deterministic `CheckVerdict[]` + judge `CriterionScore[]`, the 50/50 evidence-anchored verdict) is excellent but XORCISE-specific. An org running XORCISE alongside other agent/LLM eval tooling has no common wire format to line up a cyber-mission run next to, say, a DeepEval or Inspect run — every comparison today means writing a bespoke parser against `xorcise run status --json` or the REST `GradeResult` shape.

## What would you like to happen?

I maintain [EvalPort](https://github.com/adhabnr-ux/evalport), an open interchange spec + Python/TS SDKs for portable `TestCase`/`Grader`/`Result`/`ResultSet`/`GraderResult` shapes, plus adapters that convert other tools' native objects into it — e.g. [deepeval-openeval-adapter](https://github.com/adhabnr-ux/evalport/tree/main/adapters/deepeval-openeval-adapter) (real, tested against the installed `deepeval` package, 40 tests).

I'd like to write `xorcise-openeval-adapter` as a **standalone package**, same zero-footprint-on-the-target pattern — nothing to add to this repo, no new dependency here — that converts:

- `MissionManifest.checks` (`Check`: id/source/ref/op/args/weight/requires) + `.rubric` (`RubricCriterion`: id/text/weight) → EvalPort `TestCase.graders` — checks as `Grader(type="code", params={source, ref, op, args})`, rubric criteria as `Grader(type="llm_judge", weight=criterion.weight)`.
- A run's `GradeResult` → EvalPort `Result(test_case_id=mission_id, passed=overall>=threshold, grader_results=[...])`, where `CheckVerdict` → `GraderResult(type="code", score=1.0/0.0, passed, reason=error or blocked_by, metadata={source, ref, op, value, blocked_by})` and `CriterionScore` → `GraderResult(type="llm_judge", score, passed=status=="ok", reason, metadata={status})`. `breakdown`, `judge_status`, `overall_upper`, `judge_coverage`, `hard_fails`, `key_evidence`, `spans_truncated` would carry through on `Result.metadata` rather than get dropped — the evidence-anchoring disclosure is exactly what EvalPort's metadata escape hatch is for.
[...]

Issue #44, opened 2026-08-12
[Feature]: purge mission images — no supported way to reclaim disk or force a re-pull


christo-xorciseai:
### Before you file

- [x] I searched existing issues and discussions and this is not a duplicate.

### What problem are you trying to solve?

There is no supported way to reclaim the disk a mission's fused image occupies, and no way to
force a re-fetch. `xorcise mission delete` frees **zero bytes** — by design, and the code says so:

> `missions/runtime.py::delete_installed` — *"The fused image is intentionally left in the local
> store; pruning it is a separate concern."*

That separate concern was never built, and three things now compound:

**1. Base bumps strand a whole generation.** A base bump re-fuses every mission under a *new*
tag (`mis-<slug>:<content-hash>-base1` → `-base2`). Pulling gets the new one; nothing ever removes
the old. With ~19 library missions at ~1.3 GB each, a single base bump silently costs **~25 GB**.
This just happened for real — the fleet moved base1 → base2 on 2026-08-11.

**2. The documented escape hatch is harmful.** `cli/_diagnostics.py::disk_space` tells operators:

```
reclaim space with: docker system prune -a
```

On a developer machine that deletes all of their unrelated images, not just XORCISE's.

**3. A re-pushed tag is shadowed forever.** `mission_pull.py` skips the fetch entirely when the
image is already local:

```python
if image is not None and not deps.driver.image_exists(image):
    ...  # pull
```

Tags encode content-hash + base generation, so a genuinely new build does arrive naturally — but
if a tag is ever re-pushed in place, or a local image is corrupt, there is no override.

Today the only way through any of this is `docker rmi` by hand, after reading `installed.json` to
find the ref.

### What would you like to happen?

Three additions. They are independent and can land separately.

**1. `xorcise mission prune`** — remove fused images that no install record references.

This is the one with a live problem behind it. Reclaims stranded generations and anything left by
`mission delete`.

```
$ xorcise mission prune
would remove 19 images, reclaiming 24.7 GB:
  mis-breachpoint:aeccc8a17e009b57-base1        1.29 GB   (superseded by -base2)
  mis-ghostwire:8d90748c61f6a713-base1          1.31 GB   (not installed)
  ...
re-run with --yes to remove them
```

**2. `xorcise mission delete <id> --purge`** — remove the install *and* its fused image.

Precise: `installed.json` already records the exact ref, so no scanning and no heuristics.

[...]

PR #94, opened 2026-09-09
fix(ci): declare the label set in git — five configured labels never existed

### LINKED ISSUE #93: [Bug]: five labels referenced by dependabot.yml, release.yml and the issue forms do not exist, so the title/label convention is a silent no-op
Five labels referenced by this repo's own configuration do not exist. GitHub drops an
unknown label from an issue form or a Dependabot config **silently** — no error, no warning
— so the title-and-label convention this repo documents is currently a no-op, and four of
the nine release-note sections are unreachable.

## The missing labels

`feature` · `breaking-change` · `dependencies` · `internal` · `skip-release-notes`

The repo's actual label set is `bug · documentation · duplicate · enhancement ·
good first issue · help wanted · invalid · question · wontfix · ci-nightly · security ·
full-ci`.

## What that breaks, with evidence

**Every feature request arrives unlabelled.** `.github/ISSUE_TEMPLATE/feature_request.yml`
declares `labels: [feature]`. Issue #76 (`[Feature]: xorcise-openeval-adapter …`) carries no
label at all. #44 was hand-patched with `enhancement` — a *different* label than the form
intended — so the taxonomy has already drifted in practice.

**Dependabot's declared labels never apply.** `.github/dependabot.yml` sets
`labels: [dependencies]` for the Python and frontend ecosystems and
`labels: [dependencies, internal]` for GitHub Actions. Not one of the ten Dependabot pull
requests (#85, #84, #70, #77, #75, #64, #50, #49, #40, #39) carries either.

**Four release-note sections and the exclusion are unreachable.** `.github/release.yml`
routes 🚀 New features on `feature`, ⚠️ Breaking changes on `breaking-change`,
📦 Dependencies on `dependencies` and 🔧 Internal on `internal`, and excludes
`skip-release-notes`. None of those five labels can be applied, so nothing can ever land in
those sections and the exclusion never fires.

The sharpest case: **#41** `feat(runner)!: DinD-only on docker:29.7.1` — the `!` declares a
breaking change — merged with no `breaking-change` label, because there is none to apply.
⚠️ Breaking changes is the one section a user MUST read before upgrading.

`CONTRIBUTING.md` compounds it by listing all nine release-note labels as though they exist,
and by instructing contributors to "apply the `breaking-change` label" for a breaking change.

## Two adjacent gaps found while confirming the above

**Issue titles: 8 of 28 comply (29%).** `blank_issues_enabled: false` only removes the web
UI's *blank issue* button; `gh issue create` and the REST API bypass the forms entirely. That
is how #87, #86, #71, #67, #66, #62, #43 and #38 were filed without a prefix, and the nightly
workflow files `"Nightly CI failure"` unprefixed by design
(`.github/workflows/nightly.yml:225`).

**Pull-request titles: 30 of 30 comply, with zero enforcement.** Conventional Commits is
documented at `CONTRIBUTING.md:166` and every PR to date honours it, but nothing checks it.
Because PRs are squash-merged, that title becomes the permanent commit on `main` — it is
holding on discipline alone. Nine merged `feat(…)` PRs (#63, #59, #57, #56, #53, #52, #46,
#41, #32) carry no release-note label.

## Why this could not have been caught

Labels live in GitHub's database, not in git. Nothing in the repo declares the label set, so
no test or lint could compare what the config *references* against what the repo *has*. That
is the root cause worth fixing, not the five labels themselves.

## Proposed fix

1. **`.github/labels.yml`** — declare all 17 labels in git, making the set reviewable and
   diffable.
2. **`tests/topology/test_label_contract.py`** — assert every label referenced by
   `dependabot.yml`, `release.yml` and the issue templates is declared in `labels.yml`. This
   test fails today; it is the check that makes this class of defect unreproducible.
   (`tests/topology/test_mission_base_release_contract.py` sets the precedent for a topology
   test that reads `.github/`.)
3. **`.github/workflows/label-sync.yml`** — reconcile GitHub from the file on merge to
   `main`. Create-and-update only, never delete: removing a label silently strips it from
   every issue carrying it, which is irreversible.
4. **A `pr-contract` job in `ci.yml`**, wired into the existing `ci-ok` aggregate gate —
   asserts a Conventional Commits title and exactly one release-note label. `ci-ok` is
   already the only required status check on `main`, so this needs no branch-protection
   change. Bots exempt from the label half; skipped on `push`, mirroring `test-heavy`.
5. **`.github/workflows/issue-contract.yml`** — comment when an issue title is missing its
   `[Bug]:`/`[Feature]:` prefix. Exempts bot actors and `ci-nightly` so the nightly workflow
   is untouched. Advisory: issues have no CI to block.
6. **`CONTRIBUTING.md`** — an `## Issues` section stating the title prefixes and the label
   taxonomy, including the `feature` (a capability that did not exist) versus `enhancement`
   (improves something that does) distinction that `release.yml` assumes but nobody wrote
   down. `## Commit messages` extended to note the PR contract is CI-enforced.
7. **`CLAUDE.md`** — the same convention, under the CI-guarded conventions. AI assistants
   filing via `gh` are precisely how the unprefixed titles happened.

## Out of scope

The 20 existing unprefixed titles stay as history. Making the repo agent-friendly for
non-Claude assistants is tracked separately.

guru-xorciseai:
Closes #93.

## The defect

Labels live in GitHub's database; the configuration that references them lives in git, and
nothing reconciled the two. GitHub drops an unknown label from an issue form or a Dependabot
config **silently** — no error, no warning, no annotation — so five labels were referenced by
this repo's own configuration and **none of them existed**: `feature`, `breaking-change`,
`dependencies`, `internal`, `skip-release-notes`.

Every symptom looked like a human forgetting to label something:

| What was configured | What actually happened |
| --- | --- |
| `feature_request.yml` asks for `feature` | Every feature request arrived unlabelled (#76). #44 was hand-patched with `enhancement` instead, so the taxonomy had already drifted from what the form intended. |
| `dependabot.yml` declares `dependencies`, plus `internal` for Actions | Not one of its ten pull requests carried either. |
| `release.yml` routes 9 sections | Four were unreachable and the exclusion never fired. **#41 — `feat(runner)!:` — merged with no `breaking-change` label because there was none to apply**, leaving ⚠️ Breaking changes empty for the release a user most needed to read it. |

`CONTRIBUTING.md` compounded it by listing all nine labels as though they existed, and telling
contributors to "apply the `breaking-change` label".

**No test could have caught this**, because nothing in the repo declared what the label set
was. That asymmetry is the actual bug; the five labels are a symptom.

## What this does

- **`.github/labels.yml`** declares all 17 labels in git, where they can be reviewed and
  diffed.
- **`tests/topology/test_label_contract.py`** (14 tests) makes the declaration load-bearing:
  it fails if any config names a label the file does not declare, if a description exceeds
  GitHub's 100-character limit, or if the gate's inline label list drifts from `release.yml`.
  **This test fails on `main` today** — the exact output that motivated the PR:
  ```
  AssertionError: these labels are referenced by configuration but not declared in
  .github/labels.yml, so GitHub will silently drop them:
  {'.github/release.yml': ['breaking-change', 'dependencies', 'feature', 'internal',
                           'skip-release-notes'],
   '.github/dependabot.yml': ['dependencies', 'internal'],
   '.github/ISSUE_TEMPLATE/feature_request.yml': ['feature']}
  ```
- **`.github/workflows/label-sync.yml`** applies the file on merge to `main`, **additively**. A
  label removed from the file is reported as an extra, never deleted — deleting one strips it
  from every issue that carried it with no record of what it was. A test asserts the workflow
  contains no `DELETE`.
- **`.github/workflows/pr-contract.yml`** requires a Conventional Commits title and exactly one
  release-note label.
- **`.github/workflows/issue-contract.yml`** comments once when an issue title lacks its
  `[Bug]:`/`[Feature]:` prefix. Automation is exempt, so `nightly.yml`'s `"Nightly CI failure"`
  is untouched.
- **`.github/dependabot.yml`** now *states* its commit prefix instead of letting Dependabot
  infer one from history. It has inferred wrong here before — five of its pull requests are
  titled `Bump x from a to b` — and bots are exempt from `pr-contract`'s label half, never its
  title half, so an inferred title is a wedged dependency update waiting to happen.
- **`CONTRIBUTING.md`** gains `## Issues` and `## Pull requests`, including the
  `feature`-vs-`enhancement` rule that `release.yml` has always assumed but nobody wrote down:
  *`feature` is a capability that did not exist, `enhancement` improves one that did.*
- **`CLAUDE.md`** carries the same rules, because AI assistants filing via `gh` are precisely
  how 20 of the first 28 issue titles lost their prefix.

## Two decisions worth reviewing

**1. `pr-contract` needs registering as a required status check.** It is a separate workflow
rather than a job under `ci-ok`, which departs from the "one required check" design. The reason
is that it must re-run on `edited`/`labeled`, and ci.yml must not: putting `edited` on ci.yml
would re-run the whole suite — e2e included — on every title or body edit, and Dependabot
rewrites bodies on every rebase. The tempting mitigation is worse than the cost — if the
expensive jobs were skipped for those events, that run's `ci-ok` would supersede the earlier
genuine pass **for the same commit**, so a label edit could turn a tested green into a skipped
green. A five-second check therefore gets its own workflow. Happy to fold it into `ci-ok`
instead if you would rather keep exactly one required check.

**2. Two stale lines in CONTRIBUTING.md, fixed here.** #91/#92 changed only `ci.yml`, so the
docs still said the heavy lanes were opt-in via `full-ci`. They now run on any
non-documentation pull request, and the label is an override for documentation-only changes.
Called out because it is adjacent to, not part of, the label fix.

## Verification

| Check | Result |
| --- | --- |
| `pytest -m topology` | 45 passed |
| `pytest -m unit -n auto` | 2209 passed |
| `mypy` | no issues in 494 files |
| `ruff check .` / `ruff format --check .` | clean, 500 files formatted |
| `zizmor --pedantic .github/workflows/` | no findings |
| `pr-contract` shell logic, 19 cases | 19/19 — valid/invalid titles, 0/1/2 labels, bot exemption, and an injection attempt (`feat: $(touch …)`) confirmed **not** executed |
| All 60 historical PR titles vs the gate | 54 accepted; the 6 rejected are one pre-convention PR and five `Bump …` Dependabot titles from before the prefix was configured |

The label logic had a real bug caught by that harness: setting `IFS='|'` to split the applied
labels also changed how the inner loop split `RELEASE_NOTE_LABELS`, collapsing all nine names
into one token so **nothing ever matched** — it would have blocked every correctly-labelled PR.
It now uses `case` glob matching, which ignores `IFS`.

## Rollout

There are **zero open pull requests**, so nothing is retroactively blocked by the new gate.
The labels themselves do not exist until `label-sync` runs on merge, which is why this PR is
labelled `bug` rather than `internal` — `internal` becomes available for this class of change
afterwards.

The 8 open issues with no title prefix are left as history; the workflow only advises on new
ones.


CLAUDE.md
christo-xorciseai, line 1: Should we we not name this file `AGENTS.md` in line with the feature request referenced with this PR, and ensure we are agent agnostic ? We could potentially have `CLAUDE.md` refer to `AGENTS.md` for the full guide.

guru-xorciseai: Yes exactly. I am doing that once this pr is merged, and then I am raising the next PR, which makes the repo agent-friendly, since #95 needed a few additional file changes. Referring to #95

.github/workflows/issue-contract.yml
```diff
@@ -0,0 +1,85 @@
...
+          if printf '%s' "$TITLE" | grep -qE '^\[(Bug|Feature)\]: .+'; then
+            echo "ok  $TITLE"
+            exit 0
+          fi
+
+          # An HTML comment is invisible in the rendered issue and is what makes this
+          # idempotent across the `edited` trigger.
+          marker='<!-- xorcise:issue-contract -->'
+          if gh api --paginate "repos/$REPO/issues/$NUMBER/comments" --jq '.[].body' \
+             | grep -qF "$marker"; then
```
christo-xorciseai: **Duplicate advisory comments.** `grep -qF` under `set -o pipefail` makes this branch unreliable. Once grep matches and exits, `gh` gets SIGPIPE on its next write and the pipeline returns 141, so the `if` is false and the script falls through to posting another comment. That is the duplicate-nag case the marker exists to prevent, and it then fires on every later `edited` event.

Reproduced with gh 2.45 against this repo (read-only):

```bash
bash -c 'set -o pipefail; gh api --paginate "repos/xorcise-ai/xorcise/issues?state=all&per_page=2" --jq ".[].number" | grep -q .; echo $?'   # 141
```

The same shape with a single page returns 0, so it only bites once the issue has more than one page of comments (30 at the default page size) with the marker in an earlier page. Simplest robust fix is to not pipe `gh` into `grep` at all, the way the sync step already does with `live.txt`:

```bash
gh api --paginate "repos/$REPO/issues/$NUMBER/comments" --jq '.[].body' > comments.txt
if grep -qF "$marker" comments.txt; then
```

(`grep ... >/dev/null` does not help: GNU grep treats a `/dev/null` stdout as `-q`.)

.github/workflows/label-sync.yml
```diff
@@ -0,0 +1,118 @@
...
+name: Label sync
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - .github/labels.yml
+      - .github/workflows/label-sync.yml
+  # Manual run for the first sync after this lands, and for re-applying after a hand edit.
+  workflow_dispatch:
```
christo-xorciseai, line 22: **The sync can run from any branch.** The header (lines 3-5) and `test_sync_runs_only_on_main` both promise this only runs from `main`, but `workflow_dispatch` accepts any ref and the `sync` job has no `if:`. Anyone with write access can pick a branch in the Run workflow dropdown and apply an unreviewed `labels.yml` (colour and description of every existing label, plus arbitrary new labels) before review.

Suggest a job-level guard, e.g. `if: github.ref == format('refs/heads/{0}', github.event.repository.default_branch)`, and extending the test to assert `jobs.sync.if` exists and mentions the ref; today it only inspects `push.branches`.

.github/workflows/label-sync.yml
```diff
@@ -0,0 +1,118 @@
...
+              declared = yaml.safe_load(handle)["labels"]
+
+          for entry in declared:
+              name = entry["name"]
+              colour = entry["color"].lower()
+              description = entry["description"]
+              for field, value in (("name", name), ("description", description)):
+                  if "\t" in value:
+                      sys.exit(f"{name}: {field} contains a tab, which breaks this encoding")
+              print(f"{name}\t{colour}\t{description}")
```
christo-xorciseai, line 88: **Newlines in a description break the sync.** This guards the field delimiter but not the record delimiter. A `description: |` block, a `>` with a paragraph break, or a quoted string containing `\n` prints as two TSV rows. The reconcile loop then reads the continuation text as a label *name* with an empty colour, GitHub answers 422 (colour is required on create), and `set -e` aborts the run with the remaining labels unsynced. `test_descriptions_fit_githubs_limit` only checks length, so nothing catches it before `main`.

Reject `\n` here alongside `\t` (or emit one JSON object per line and read it back with `jq -r`), and add the same check to the test.

christo-xorciseai (commented):
Reviewed at `adaf76e`. The diagnosis is right and the shape is good: additive sync, titles handled via `env` only, and the test that binds the gate's inline list to `release.yml`. CI is green on the head SHA, and ruff, mypy, zizmor and the topology lane all pass locally.

The inline comments carry the correctness issues and the stale text inside the diff. The items below are in files this PR does not touch, so they cannot be anchored.

**Stale text in untouched files**

- `CONTRIBUTING.md` lines 73-74: the lanes table still says `integration` and `e2e` run "only with the `full-ci` label", contradicting lines 95 and 112-116 of the same file. The PR body says this was fixed.
- `.github/PULL_REQUEST_TEMPLATE.md` lines 35-38 and 77-80: still tells authors the heavy lanes do not run by default and to ask for `full-ci`, and carries a fourth copy of the nine labels without saying `pr-contract` fails without one. Fork authors cannot apply labels, and only CONTRIBUTING line 222 says a maintainer will.
- `.github/workflows/ci.yml` lines 467-468: `ci-ok` is still "THE single required status check. Branch protection names only this", while CONTRIBUTING now lists two.
- `.github/workflows/nightly.yml` lines 175-178: see the note on `ci-nightly` in `labels.yml`.

**Maintenance note**

The release-note taxonomy now lives in five hand-maintained places (`labels.yml`, the CONTRIBUTING table, CLAUDE.md, the PR template, the `issue-contract` comment) plus the `pr-contract` env, and only `labels.yml`, `release.yml` and `pr-contract` are test-bound to each other.

**Merge-time checklist**

1. Rename the `pr-contract` job's display name to plain `pr-contract` and register it in the `main` ruleset; until then the gate is advisory.
2. Decide on squash-only plus `PR_TITLE`, or soften the "title becomes the commit" claim.
3. Decide what to do about `ci.yml`'s `labeled` trigger now that a label is mandatory on every PR.

.github/workflows/label-sync.yml
```diff
@@ -0,0 +1,118 @@
...
+          # The live set, so an existing label is PATCHed rather than POSTed (a POST for a
+          # label that exists is a 422, which would fail the run for a no-op).
+          gh api --paginate "repos/$REPO/labels" --jq '.[].name' | sort > live.txt
+          echo "Repository currently has $(wc -l < live.txt) labels."
+
+          created=0
+          updated=0
+          while IFS=$'\t' read -r name colour description; do
+            [ -n "$name" ] || continue
+            if grep -Fxq "$name" live.txt; then
```
christo-xorciseai: **Case-insensitive label names.** GitHub label names are case-insensitively unique, but this match is case-sensitive. If the live set ever has `Bug` and the file says `bug`, this routes to POST, the API returns 422 `already_exists`, and the loop aborts; every label after it in the file is skipped and the extras report never prints. No collision exists today, so this is latent, but surviving that kind of drift is what the sync is for.

`grep -Fxqi -- "$name" live.txt`, then PATCH by the *live* name (so `new_name=$name` also corrects the casing), handles it. The `--` matters on its own: a name beginning with `-` is parsed as a grep option.

.github/workflows/pr-contract.yml
```diff
@@ -0,0 +1,148 @@
+# The pull-request contract: the title, and exactly one release-note label.
+#
+# Both are load-bearing AFTER the merge, which is what makes a human reviewer a poor place to
+# enforce them:
+#
+#   * Pull requests here are SQUASH-merged, so the title becomes the permanent commit message
+#     on `main`. A malformed one cannot be corrected afterwards without rewriting history.
```
christo-xorciseai: **This premise does not hold for the repo as configured.** The `main` ruleset allows squash, rebase and merge (`main` has merge commits for #46, #25 and #24), and `squash_merge_commit_title` is `COMMIT_OR_PR_TITLE`, so a single-commit PR squashes with the *commit* subject, not the PR title. The gate therefore validates a string that often is not what lands on `main`.

Either change the repo settings so the claim becomes true (squash only in the ruleset, `PR_TITLE` for the squash title), or soften it here and in CONTRIBUTING.md line 215 and CLAUDE.md line 69.

.github/workflows/pr-contract.yml
```diff
@@ -0,0 +1,148 @@
...
+permissions: {}
+
+concurrency:
+  # A rapid sequence of label edits should settle on the last one, not queue nine runs.
+  group: pr-contract-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  contract:
+    name: pr-contract (title, release-note label)
```
christo-xorciseai: **Check-run name.** The status-check context is the job's display name, so this registers as `pr-contract (title, release-note label)`, not `pr-contract`. A maintainer following CONTRIBUTING.md and adding the context `pr-contract` to the ruleset would create a required check that never reports, and every PR would sit at "Expected — Waiting for status".

Suggest `name: pr-contract` here, the way `ci-ok` names itself, and keep the long description in the step name.

CONTRIBUTING.md
```diff
@@ -163,14 +171,77 @@ it does not license letting dependencies rot.
...
+`feat: …`, `fix: …`, `docs: …`, `test: …`, `chore: …`, `ci: …`, `style: …`,
+`refactor: …`, `build: …`, `perf: …`, `revert: …`.
+Keep the subject imperative; use the body for the "why".
+
+## Pull requests
 
-Pull requests are **squash-merged**, so the PR *title* becomes the commit message on `main` —
-write it as the changelog entry you want.
+Two things about a pull request are checked by CI, because both are permanent once it merges.
+`pr-contract` is a required status check.
```
christo-xorciseai: The live `main` ruleset requires only `ci-ok`, so as merged `pr-contract` is advisory and blocks nothing until someone registers it. Worth phrasing this (and the row at line 100) as "must be registered as a required status check" and making the registration an explicit merge step. See also the note on the job's display name in `pr-contract.yml`: the context to register is whatever `name:` says.

.github/workflows/pr-contract.yml
```diff
@@ -0,0 +1,148 @@
...
+#     invisible until release day, by which point the notes are already wrong.
+#
+# Why this is NOT a job in ci.yml, where `ci-ok` already gates everything: the checks here must
+# re-run when a title is edited or a label is added, and ci.yml must not. Putting `edited` on
+# ci.yml's triggers would re-run the whole suite — e2e included — on every title or body edit,
+# and Dependabot rewrites bodies on every rebase. The tempting mitigation is worse than the
+# cost: if the expensive jobs were skipped for those events, that run's `ci-ok` would supersede
+# the earlier genuine pass for the same commit, and a label edit could turn a tested green into
+# a skipped green. A five-second check therefore gets its own workflow and its own required
+# status check, rather than sharing a trigger set with a fifteen-minute suite.
```
christo-xorciseai, line 27: **ci.yml already re-runs on every label, and this PR makes labelling mandatory.** The argument here is right about `edited`, but `ci.yml` triggers on `labeled` (for `full-ci`) with `cancel-in-progress` on for PRs. So every PR now gets its CI run cancelled and restarted when the release-note label lands, and the cancelled run's `ci-ok` (which runs on `always()`) reports failure. This PR's own head SHA shows it: two `CI` runs created in the same second because `bug` was applied at open, one cancelled with a failed `ci-ok`, then a second success two minutes later.

One caution about the obvious fix. A job-level `if:` that skips everything on non-`full-ci` label events would produce a *skipped* `ci-ok` check run on the same SHA, which GitHub counts as passing for a required check, the exact "skipped green" this header warns about. Realistic options are dropping `labeled` from `ci.yml` and documenting that `full-ci` takes effect on the next push, or keeping the extra run but not cancelling. The trigger predates this PR, so a follow-up is fine, but it needs a decision.

.github/dependabot.yml
```diff
@@ -124,6 +142,12 @@ updates:
...
     open-pull-requests-limit: 5
+    # See the note on the uv ecosystem above: the title is stated so `pr-contract` cannot be
+    # surprised by a style Dependabot inferred.
+    commit-message:
+      prefix: build
+      prefix-development: build
+      include: scope
     labels:
       - dependencies
       - internal
```
christo-xorciseai: **Two release-note labels on every Actions PR.** This applies both `dependencies` and `internal`. That is the state `pr-contract` rejects for humans, and `release.yml` routes `dependencies` first so `internal` is dead weight here. The `*[bot]` exemption hides it, and also means a bot PR with *zero* labels passes the gate, which is exactly the #93 symptom for the author #93 was about (if `dependencies` ever vanished again, nothing would notice). The stated reason for the exemption, a pre-existing backlog, is empty per the rollout section.

Suggest dropping `internal` here and adding a test that each `updates[].labels` intersects `selectable_labels()` in exactly one element; with that in place the bot exemption can go too.

tests/topology/test_label_contract.py
```diff
@@ -0,0 +1,271 @@
...
+    dependabot: set[str] = set()
+    for update in DEPENDABOT["updates"]:
+        dependabot.update(update.get("labels", []))
+    referenced[".github/dependabot.yml"] = dependabot
+
+    for template in sorted(ISSUE_TEMPLATE_DIR.glob("*.yml")):
+        if template.name == "config.yml":  # the chooser, not a form — it has no labels
+            continue
+        form = _load(template)
+        referenced[f".github/ISSUE_TEMPLATE/{template.name}"] = set(form.get("labels", []))
```
christo-xorciseai, line 124: **Narrower than the claim.** `labels.yml` lines 4-5, CONTRIBUTING.md line 199 and CLAUDE.md line 75 say this test covers "every label named anywhere in `.github/`", but `referenced_labels` only scans `release.yml`, `dependabot.yml` and `ISSUE_TEMPLATE/*.yml`. `full-ci` (ci.yml) and `ci-nightly` (nightly.yml, issue-contract.yml) are unbound: delete both from `labels.yml` and all 14 tests stay green. `*.yaml` issue forms are also skipped.

Either scan the workflows too (a regex over `contains(github.event.pull_request.labels.*.name, '...')`, `--label`, and `gh label create` covers the current uses) or narrow the claim in the three places that make it.

.github/labels.yml
```diff
@@ -0,0 +1,110 @@
...
+  # Machine-readable. These change what CI does, or record what it found. Not for triage.
+  # ---------------------------------------------------------------------------------------
+
+  - name: full-ci
+    color: "0e8a16"
+    description: "Run the Docker-heavy test lanes (test-heavy) on this PR"
+
+  - name: ci-nightly
+    color: "b60205"
+    description: "Automated nightly validation failure"
```
christo-xorciseai, line 111: `nightly.yml` lines 175-178 still create this label inline with a hard-coded colour and description. That is now a second definition of a label this file owns; they agree today, but an edit here will not propagate to nightly's bootstrap path, and its `|| true` swallows any failure. Fine to keep as a bootstrap for a fresh fork, but worth a pointer in one of the two places to the other, or a test that the two agree.

.github/workflows/pr-contract.yml
```diff
@@ -0,0 +1,148 @@
...
+
+          # Conventional Commits: <type>(<optional scope>)<optional !>: <subject>. The type
+          # list is the one CONTRIBUTING.md documents, plus the types already in use on `main`
+          # (`refactor`, `build`, `perf`) and `revert`.
+          #
+          # Subject LENGTH is deliberately not checked. CONTRIBUTING.md suggests ~72
+          # characters, but the descriptive titles this project actually merges run longer and
+          # read better for it; enforcing the suggestion would fail most of the history.
+          types='feat|fix|docs|test|chore|ci|style|refactor|build|perf|revert'
+          if printf '%s' "$TITLE" | grep -qE "^($types)(\([a-z0-9._/-]+\))?!?: .+"; then
```
christo-xorciseai: **Lowercase-only and single-scope.** Dry-running the regex: `Feat: x`, `feat(CLI): x`, `feat(cli,rest): x` and GitHub's own Revert button title `Revert "feat: x"` all fail, while the `::error` text and CONTRIBUTING say nothing about case or scope shape. The Revert case is the one that bites in practice: every UI revert needs a manual retitle before the gate passes.

If lowercase and single-scope are intended, say so in the error text; otherwise `grep -qiE` and `[a-z0-9._/,-]+` widen it.

.github/labels.yml
```diff
@@ -0,0 +1,110 @@
...
+#
+# Colours are bare six-digit hex, no leading `#` — GitHub's API rejects it.
+# The human-facing taxonomy, and the rule for choosing between these, is in CONTRIBUTING.md
+# under "Issues" and "Pull requests".
+
+labels:
+  # ---------------------------------------------------------------------------------------
+  # Release-note labels. `.github/release.yml` routes a merged pull request into a section of
+  # the generated notes by these, in this order. Every pull request carries exactly one; the
+  # `pr-contract` job in ci.yml enforces that, because an unlabelled pull request lands in
```
christo-xorciseai: `pr-contract` is a standalone workflow, not a job in `ci.yml`; its own header explains why it must not be one.

.github/workflows/pr-contract.yml
```diff
@@ -0,0 +1,148 @@
...
+          set -uo pipefail
+          status=0
+
+          # Conventional Commits: <type>(<optional scope>)<optional !>: <subject>. The type
+          # list is the one CONTRIBUTING.md documents, plus the types already in use on `main`
+          # (`refactor`, `build`, `perf`) and `revert`.
+          #
+          # Subject LENGTH is deliberately not checked. CONTRIBUTING.md suggests ~72
+          # characters, but the descriptive titles this project actually merges run longer and
+          # read better for it; enforcing the suggestion would fail most of the history.
```
christo-xorciseai: Three references here point at text this PR changes: CONTRIBUTING.md no longer suggests ~72 characters (line 217 now says length is not enforced); its type list at lines 205-206 now includes `refactor`, `build`, `perf` and `revert`, so "plus the types already in use" is stale; and the `::error` at line 84 sends readers to "Pull requests", but the types are documented under "Commit messages".

CONTRIBUTING.md
```diff
@@ -92,20 +92,28 @@ Every pull request runs:
...
 repository secret**. A maintainer approves the first workflow run for each new contributor.
 
 ### The `full-ci` label
 
-Ask a maintainer to add the **`full-ci`** label if your change touches the runner, the
-control plane, networking, ports, containers, or `xorcise up`. That re-runs CI with the
-Docker-heavy lanes included, and `ci-ok` then requires them to pass.
+The Docker-heavy lanes run by default on any pull request that changes something other than
+documentation, so you usually do not need this label. Ask a maintainer to add **`full-ci`** for
+a documentation-only change that should still run them anyway — a change to `docs/` that
```
christo-xorciseai: There is no `docs/` directory at this PR's head. Pick a real documentation-only example (a README or CONTRIBUTING change, say).

CLAUDE.md
```diff
@@ -53,3 +53,25 @@ uv run lint-imports
   top-level names are reserved for products (guard).
 - **Stubs are filled in place** — do NOT move packages; if one is missing, fix the scaffold,
   never ad-hoc a new location.
+- **Issue and pull-request titles are contracts** — see below. This one is aimed at you
+  specifically: filing from a terminal is what broke it.
+
+## Filing issues and pull requests
+
+The issue forms and `pr-contract` enforce these, but `gh issue create` and the REST API bypass
```
christo-xorciseai: This bullet sits under "Hard conventions (CI-guarded)" and line 61 says the forms and `pr-contract` enforce it, but issue titles are only *commented on* by `issue-contract.yml` (its header says it "comments rather than blocks"), and `pr-contract` never fires on issues. An assistant filing an unprefixed issue will expect a red check and get a comment instead. Worth saying explicitly that CI blocks the PR half and only advises on the issue half.

.github/workflows/pr-contract.yml
```diff
@@ -0,0 +1,148 @@
...
+              # shape avoids. `case` globbing ignores IFS, so no inner loop is needed.
+              recognised=" $(printf '%s' "$RELEASE_NOTE_LABELS" | tr -s '[:space:]' ' ') "
+
+              applied=""
+              # `|` separates the applied labels because a name may contain spaces ("good
+              # first issue"); no release-note label contains either character, so splitting
+              # on `|` keeps every name whole and pulls none of them apart.
+              original_ifs="$IFS"
+              IFS='|'
+              for label in $LABELS; do
```
christo-xorciseai, line 151: Minor: `$LABELS` is unquoted on purpose for the IFS split, but the split words still undergo pathname expansion. A label literally named `*` with files in the cwd would count them (reproduced locally with `LABELS='*'` and two files present: count 2, FAILED). The runner cwd is empty today since nothing is checked out, so this is cheap insurance: `set -f` before the loop, `set +f` after.

tests/topology/test_label_contract.py
```diff
@@ -0,0 +1,271 @@
...
+
+    def test_sync_workflow_exists(self) -> None:
+        assert LABEL_SYNC_PATH.exists(), "no label-sync workflow to apply labels.yml"
+
+    def test_sync_never_deletes(self) -> None:
+        text = LABEL_SYNC_PATH.read_text()
+        # A label removed from GitHub is removed from every issue and pull request that
+        # carried it, with no record of what it was. The sync reconciles additively and
+        # reports extras instead; this asserts nobody ever "tidies" that up.
+        assert "--method DELETE" not in text, (
```
christo-xorciseai: Minor: this only catches the literal `--method DELETE`. `-X DELETE`, `--method=DELETE`, `gh label delete` and `curl -X DELETE` all pass. Something like `re.search(r"(-X|--method[= ])\s*DELETE|label delete", text)` covers the realistic forms.

tests/topology/test_label_contract.py
```diff
@@ -0,0 +1,271 @@
...
+    return loaded
+
+
+# `on:` is YAML 1.1's boolean true, so a workflow's trigger block parses under the key `True`,
+# not `"on"`. Every reader here goes through this rather than `workflow["on"]`, which looks
+# correct and raises KeyError.
+def _triggers(workflow: dict[Any, Any]) -> dict[Any, Any]:
+    triggers = workflow.get(True, workflow.get("on", {}))
+    assert isinstance(triggers, dict), "workflow triggers did not parse as a mapping"
+    return triggers
```
christo-xorciseai, line 68: Minor: `test_mission_base_release_contract.py` line 42 has the same `on:` to `True` workaround inline (with the opposite preference order), and `ROOT` is now defined in four topology files. `tests/_helpers.py` exists; worth moving both there.

.github/workflows/label-sync.yml
```diff
@@ -0,0 +1,118 @@
...
+      - name: Set up uv
+        uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
+        with:
+          version: ${{ env.UV_VERSION }}
+
+      # YAML is parsed by a YAML parser, not by grep. The output is tab-separated because a
+      # description contains spaces and apostrophes but never a tab.
+      - name: Render labels.yml to a tab-separated table
+        run: |
+          uv run --no-project --with pyyaml==6.0.3 -- python - <<'PY' > labels.tsv
```
christo-xorciseai, line 65: Minor: setup-uv plus a `pyyaml==6.0.3` download on every run, to render 17 rows. The pin is invisible to Dependabot and the inline Python to ruff. `yq` ships on the ubuntu runner image and does this in one line, or the code could live under `scripts/` and run via `uv run python scripts/...` the way `ci.yml`, `nightly.yml` and `release.yml` already do.

.github/workflows/issue-contract.yml
guru-xorciseai: Fixed in 244d812. Reproduced your exact command before touching it, and the single-page control too:

```
gh api --paginate ...per_page=2  | grep -q .   -> 141
gh api ...per_page=100           | grep -q .   -> 0
```

So it would have survived every short-thread test, which is the part that makes it worth the comment. Collected to `comments.txt` and `grep -qF --` against the file, the way the sync step already does. Your parenthetical about `>/dev/null` is right and I kept it in the comment — `cat` of a 100k-line file into `grep -qF` also returns 141 here, so the short-circuit is real regardless of where stdout points.

One thing I could not bind: no test can observe this without a live paginated issue. The comment now carries the mechanism and the measurement so the next person does not "simplify" it back into a pipe.

.github/workflows/label-sync.yml
guru-xorciseai: Fixed in 244d812, as you suggested:

```yaml
if: github.ref == format('refs/heads/{0}', github.event.repository.default_branch)
```

and `test_sync_runs_only_on_main` now asserts the job-level guard exists and mentions both `github.ref` and `default_branch`, not just `push.branches`. Verified it binds: the test fails if the `if:` is removed.

Worth noting the behaviour change you implied — dispatching from a branch now shows the job as *skipped* rather than running. That is the intent, and the header already promised it.

.github/workflows/label-sync.yml
guru-xorciseai: Fixed in 244d812. The render now rejects both delimiters:

```python
if "\t" in value or "\n" in value:
```

with the reasoning you gave recorded next to it — the record delimiter is the worse of the two, because the continuation text becomes a label *name* with no colour and `set -e` aborts mid-sync.

Bound in two directions: `test_no_name_or_description_carries_a_delimiter` checks the data, and `test_the_sync_rejects_both_delimiters` checks the guard is still in the workflow. Latent today — no current description contains a newline, as you'd expect.

.github/workflows/label-sync.yml
guru-xorciseai: Fixed in 244d812, close to your suggestion but with one change:

```bash
live_name=$(grep -Fxi -m1 -- "$name" live.txt || true)
if [ -n "$live_name" ]; then
  encoded=$(jq -rn --arg s "$live_name" '$s|@uri')
  gh api --silent --method PATCH "repos/$REPO/labels/$encoded" -f "new_name=$name" ...
```

I capture the live spelling rather than using `-q`, because the PATCH has to address the *live* name in the path while `new_name` carries the declared one — otherwise a drifted `Bug` gets a 404 on the path instead of converging on `bug`. `--` kept for the reason you gave. `test_the_sync_matches_label_names_case_insensitively` pins `grep -Fxi`.

.github/workflows/pr-contract.yml
guru-xorciseai: You are right, and I checked the repository rather than taking it as read:

```
allow_squash=true  allow_merge=true  allow_rebase=true
squash_merge_commit_title=COMMIT_OR_PR_TITLE
main ruleset allowed_merge_methods: squash, rebase, merge
```

plus the merge commits for #24, #25 and #46. So the premise was false in two independent ways.

Softened in all three places in 244d812 — the workflow header, CONTRIBUTING.md and CLAUDE.md now say "normally squash-merged … usually becomes the commit message", and state outright that a single-commit request squashes under the *commit* subject and that merge and rebase are still permitted. The gate is described as a strong convention check, not a guarantee about the bytes.

I deliberately did **not** change the repository settings to make the claim true. Squash-only plus `PR_TITLE` is a policy call about how this repo merges, not a workflow edit, and it belongs to whoever owns that — flagging it rather than deciding it.

.github/workflows/pr-contract.yml
guru-xorciseai: Fixed in 244d812 — `name: pr-contract`, with the descriptive wording moved onto the step. Confirmed live on this push: the check now reports as **`pr-contract`**, so the context in CONTRIBUTING.md is the one a maintainer would actually register.

`test_the_job_names_itself_exactly_pr_contract` pins it, because this is a string whose *only* consumer is a repository setting — nothing else in the tree would ever notice a rename. Verified it fails if the name drifts back.

Renaming does leave the old `pr-contract (title, release-note label)` context behind on any PR that already reported under it, which is harmless while nothing requires it — but it is an argument for registering the context *after* this lands, not before.

CONTRIBUTING.md
guru-xorciseai: Fixed in 244d812. CONTRIBUTING.md now says it plainly, as a maintainer step rather than a statement of fact:

> **Maintainers:** that context has to be **registered as a required status check** on `main` before it blocks anything. Until it is, `pr-contract` reports and is advisory — the `main` ruleset currently requires `ci-ok` only. Register the context exactly as the job names itself: `pr-contract`.

I verified the ruleset lists exactly one context (`ci-ok`) rather than trusting the claim. Paired with the job rename above, so the string in that sentence is now the string GitHub reports.

.github/workflows/pr-contract.yml
guru-xorciseai: Fixed in 244d812 by dropping `labeled` from ci.yml — your option 1.

Your diagnosis is right and the evidence is on this PR's own head SHA: two `CI` runs and two `PR contract` runs, one of each cancelled. One correction to the mechanism, though: `if: always()` does not make `ci-ok` *report a failure* on a superseded run — a concurrency cancellation marks the queued job `cancelled` directly rather than letting it execute. The merge-gate effect is the same and the conclusion is unchanged, but the wording matters if someone later reasons from it.

What tipped it to option 1 rather than "keep the run, do not cancel": #91 already made the heavy lanes run on any non-docs change, so `full-ci` is now only needed for a docs-only PR that wants them — rare. The trigger was close to vestigial, and option 2 still pays for an extra full run on every PR plus a pending flicker on the required check. Cost of the change is that `full-ci` takes effect on the next push, which is now documented in CONTRIBUTING.md and the PR template.

Your caution about the `if:`-based fix producing a skipped-green `ci-ok` is well taken — that is why I did not go near it. `test_ci_does_not_retrigger_on_label_events` pins `labeled` out.

.github/dependabot.yml
guru-xorciseai: Fixed in 244d812: `internal` dropped, so every ecosystem block now applies exactly one release-note label, and `test_each_update_block_applies_exactly_one` binds it — verified it fails against the old config and passes against the new.

Two corrections to the write-up, having checked each part:

- Only the **github-actions** block double-labelled. The uv and npm blocks each applied `dependencies` alone, so "the other update blocks" overstates it slightly.
- "then the bot exemption can go" is where I'd push back. Both dependabot.yml and pr-contract.yml already note that Dependabot applies labels at *creation* time only — so if the taxonomy changes while an old bot PR is open (plausible at `open-pull-requests-limit: 5` on a weekly schedule), that PR's label set could stop satisfying "exactly one" with nothing able to fix it. Removing the exemption is safe *today* (zero open PRs, and every block now yields exactly one), but it is not unconditionally safe, so I have left it in place with the one-label test as the real guard.

You are right that the zero-label bot pass is a hole, and right that the backlog justification was empty — the rollout section does say zero open PRs. The one-label test closes the drift you were actually worried about; the exemption now guards a timing case rather than a labelling case.

tests/topology/test_label_contract.py
guru-xorciseai: Confirmed empirically before fixing, because the "all 14 stay green" claim is the kind worth testing: I deleted both labels from labels.yml and got `14 passed`. Exactly as you said.

Fixed in 244d812 — `referenced_labels()` now also scans `.github/workflows/*.yml` for three shapes (`contains(...labels.*.name, 'X')`, `gh label create X`, `--label X`) and picks up `*.yaml` issue forms. Re-ran the deletion afterwards:

```
assert not {'.github/workflows/ci.yml': ['full-ci'],
            '.github/workflows/nightly.yml': ['ci-nightly']}
```

One trap the first attempt hit, worth recording: a bare `labels\.\*\.name,` pattern also matches `join(github.event.pull_request.labels.*.name, '|')` in pr-contract.yml and yields `|` as a "label". The pattern is anchored on `contains(` for that reason.

I kept the claim in labels.yml/CONTRIBUTING/CLAUDE as-is rather than narrowing it, since it is now true.

.github/labels.yml
guru-xorciseai: Fixed in 244d812 with a test rather than a pointer, since a comment on either side would go stale the same way the definitions would: `test_nightly_bootstrap_matches_the_declaration` parses nightly's `gh label create ci-nightly` and asserts its `--color` and `--description` match labels.yml. Verified it fails if either drifts.

Keeping the bootstrap, for the reason you gave — a fresh fork needs it before the first sync, and `|| true` is correct there. The one nuance: the colours differ in *case* today (`b60205` vs `B60205`), so the test compares case-insensitively, which is what GitHub does anyway.

.github/workflows/pr-contract.yml
guru-xorciseai: Confirmed by dry-running the exact regex — all four of your cases fail. Fixed in 244d812, but not quite as suggested, and the difference matters:

**Your proposed fix does not cover the Revert case.** `grep -qiE` plus a comma in the scope class fixes `Feat:`, `feat(CLI):` and `feat(cli,rest):`, but `Revert "feat: x"` still fails — it starts `Revert "`, not `revert: `, so no amount of widening the type/scope pattern reaches it. Since you called that the one that bites in practice, it needed its own branch:

```bash
if printf '%s' "$TITLE" | grep -qE "^($types)(\([a-z0-9._/,-]+\))?!?: .+" \
   || printf '%s' "$TITLE" | grep -qE '^Revert "'; then
```

On case, I took your *other* option rather than `-qiE`: lowercase stays required, and the error message now says so outright. `-qiE` would also accept `FEAT: x`, which is looser than the convention we are trying to hold. Multi-component scopes are allowed.

Dry-run after the change: `feat(cli,rest):`, `Revert "feat: x"`, `Revert "fix(cli): y"`, `revert: …` all pass; `Feat:`, `feat(CLI):`, `FEAT:` all still fail, deliberately and with an error that explains why.

.github/labels.yml
guru-xorciseai: Fixed in 244d812 — it now reads "`pr-contract` workflow enforces that (its own workflow, not a job in ci.yml — see the header of .github/workflows/pr-contract.yml for why)".

Small note on the location: line 32 is a `#` comment heading the release-note group, not any label's `description:` field — `full-ci`'s actual description never mentioned pr-contract. Substance was right, so this is only worth saying because anyone verifying by reading the description would have found nothing there.

.github/workflows/pr-contract.yml
guru-xorciseai: All three fixed in 244d812. The comment now says length is deliberately unchecked *and that CONTRIBUTING.md agrees* rather than citing a ~72 that no longer exists; the type list is described as "the one CONTRIBUTING.md documents under Commit messages" with no stale "plus the types already in use"; and the `::error` points at 'Commit messages', which is where the types actually live.

The error text also picked up the lowercase rule and dropped the "cannot be corrected after the squash-merge" phrasing, since that claim did not survive your other comment.

CONTRIBUTING.md
guru-xorciseai: Fixed in 244d812 — verified there is no `docs/` in the tree, and the example is now "a rewrite of this file's runner instructions", with a parenthetical that the documentation site lives in its own repository so nobody goes looking again.

The same paragraph also gained the thing that section was missing: `full-ci` now takes effect on the next push, not immediately, because ci.yml no longer triggers on `labeled`.

CLAUDE.md
guru-xorciseai: Fixed in 244d812. The asymmetry is now stated first, and aimed at exactly the reader you describe:

> These are enforced ASYMMETRICALLY, and the difference matters to you: `pr-contract` **fails the check** on a malformed pull-request title or a missing release-note label, while an issue title only earns an advisory comment from `issue-contract` — an issue has no status check to gate, so nothing blocks. Do not wait for a red check on an issue; there will not be one.

Small note: the sentence you quoted is at line 61 under "Filing issues and pull requests", not under "Hard conventions (CI-guarded)" — two different sections. The substance holds either way, and an assistant reading the bullet under the CI-guarded heading would draw exactly the wrong conclusion, which is why it was worth fixing.

.github/workflows/pr-contract.yml
guru-xorciseai: Fixed in 244d812 with `set -f` / `set +f` scoped to the loop, and reproduced your result first (count 2 with two files present, 1 with `set -f`).

Agreed it is insurance rather than a live defect: the job has one step and no checkout, so `$GITHUB_WORKSPACE` is empty and an unmatched `*` survives as a literal. The comment says that explicitly, so nobody reads the guard as a fix for something exploitable today. It matters the day a step above it writes a file.

tests/topology/test_label_contract.py
guru-xorciseai: Fixed in 244d812, with your regex as the starting point:

```python
re.search(r"(?:-X|--method[=\s])\s*DELETE\b|\blabel\s+delete\b", text, re.IGNORECASE)
```

Case-insensitive and `\b`-anchored, so `--method=DELETE`, `-X DELETE`, `gh label delete` and `curl -X DELETE` are all caught. The assertion message now names the spelling it found, which is the part that makes a failure actionable.

tests/topology/test_label_contract.py
guru-xorciseai: Agreed, and deliberately **not** in this change — flagging rather than deferring silently.

Confirmed the duplication: `ROOT` is defined in four topology files, and `test_mission_base_release_contract.py:42` does carry the same `on:`-parses-as-`True` workaround with the opposite preference order. `tests/_helpers.py` is the right home for both.

My reason for holding it: this PR is already +862/-15 across eight files and touches three of those four test modules not at all. Moving a shared helper *into* them is a mechanical change that deserves to be reviewable on its own, rather than buried in a diff about labels. Noted in the commit message as deferred so it does not evaporate.

.github/workflows/label-sync.yml
guru-xorciseai: Fair, and deliberately deferred — with one finding that changes the shape of the fix.

`uv run python scripts/render_labels.py` **without** `--no-project` resolves and installs the whole project dependency set (typer, fastapi, sqlalchemy, docker, opentelemetry-proto …) just to render 17 rows, which is worse than the pyyaml download it replaces. So "move it under scripts/ like the other workflows" is not a like-for-like swap: it needs either `--no-project --with pyyaml` again (same pin problem, just relocated) or the `yq` route.

`yq` on the runner image is the clean answer and I would rather do it as its own change with its own verification than fold an untested render rewrite into a correctness fix. Your underlying point stands: the pin is invisible to Dependabot and the inline Python to ruff. Recorded as deferred in the commit message.

guru-xorciseai:
Addressed in `244d812`. I verified every claim against the tree or the live repository before changing anything rather than working from the review text — **all 27 held up**, and several were exact (the "delete both and all 14 tests stay green" prediction reproduced verbatim). The inline threads carry the per-item detail; this covers the four items you could not anchor, plus the checklist.

## Stale text in untouched files — all four fixed

- **CONTRIBUTING.md 73-74** — corrected. You were right that the PR body claimed this was fixed and it was not; the table still said `⚠️ only with the full-ci label` while lines 95 and 112-116 said the opposite. It now reads `✅ any PR that changes more than docs`.
- **PULL_REQUEST_TEMPLATE.md 35-38** — rewritten: the lanes run automatically on any non-docs change, `full-ci` takes effect on the next push, and it now says outright that a release-note label is required, that `pr-contract` fails without one, and that **fork authors cannot apply labels and have nothing to do** — the gap you identified between the template and CONTRIBUTING line 222.
- **ci.yml 467-468** — corrected. `ci-ok` is now described as "the required status check for everything in THIS workflow", with an explicit note that `pr-contract` is a separate context.
- **nightly.yml 175-178 / `ci-nightly`** — bound by a test rather than a comment, since a pointer on either side goes stale the same way the definitions do. `test_nightly_bootstrap_matches_the_declaration` parses nightly's `gh label create` and asserts colour and description match labels.yml.

## Maintenance note — partly closed

Your count was right in substance. Two refinements from checking each file: CLAUDE.md and `issue-contract` carry the narrower **issue-type** taxonomy (five names), not the nine-item release-note one — so the release-note list lives in five places, of which `labels.yml`, `release.yml` and `pr-contract` were test-bound. And dependabot.yml and the issue forms were *already* bound to labels.yml, which the note understates.

The two genuinely unbound copies are now bound: `test_every_release_note_label_appears_in_contributing` and `..._in_the_pull_request_template`. Verified they bind by adding a fictional tenth label to labels.yml + release.yml — both fail and name it. That closes the direction that actually caused #93: a label declared in configuration and documented nowhere an author reads.

## Merge-time checklist

1. **Job renamed** — `name: pr-contract`, and this push confirms it: the check reports as `pr-contract`. `test_the_job_names_itself_exactly_pr_contract` pins it, because a repository setting is the string's only consumer. **Registration is still yours to do**, and better done *after* this lands — renaming leaves the old `pr-contract (title, release-note label)` context behind on PRs that already reported under it.
2. **Softened, not enforced.** Verified `allow_merge=true`, `allow_rebase=true`, `squash_merge_commit_title=COMMIT_OR_PR_TITLE`, plus your merge commits for #24/#25/#46. All three places now say "normally squash-merged … usually becomes the commit message" and state that a single-commit PR squashes under the *commit* subject. I did not change the repository settings — squash-only plus `PR_TITLE` is a policy decision, not a workflow edit.
3. **`labeled` dropped from ci.yml** — your option 1. #91 already made the heavy lanes run on any non-docs change, so the trigger was close to vestigial while `pr-contract` made a label mandatory on every PR. `full-ci` now takes effect on the next push, documented in both places, and `test_ci_does_not_retrigger_on_label_events` pins it out.

## Two corrections to the review

- **The Revert fix you proposed does not reach the Revert case.** `grep -qiE` plus a comma in the scope class fixes `Feat:`, `feat(CLI):` and `feat(cli,rest):`, but `Revert "feat: x"` still fails — it starts `Revert "`, not `revert: `. Since you called that the one that bites, it needed its own branch. I also took your *other* option on case: lowercase stays required and the error now says so, because `-qiE` would accept `FEAT: x` too.
- **`if: always()` does not make a superseded `ci-ok` report failure** — a concurrency cancellation marks it `cancelled` directly. Same merge-gate effect, and your conclusion is unchanged, but worth correcting before someone reasons from the mechanism.

Also: only the **github-actions** block double-labelled (uv and npm each applied `dependencies` alone), and I have kept the bot exemption rather than removing it — both configs already note Dependabot labels are applied at creation time only, so a taxonomy change landing while an old bot PR is open could wedge it with nothing able to fix it. The new one-label-per-block test is the real guard for the drift you were worried about.

## Deferred, with reasons

- **`tests/_helpers.py`** (your `ROOT` / `on:`-workaround note) — confirmed: four topology files define `ROOT`, and `test_mission_base_release_contract.py:42` does carry the opposite preference order. Held back because this diff is already large and touches three of those four modules not at all; it deserves its own reviewable change.
- **label-sync's setup-uv + pyyaml render** — deferred, and the shape of the fix changed on inspection: `uv run python scripts/...` *without* `--no-project` installs the entire project dependency set to render 17 rows, which is worse than the pyyaml download. `yq` is the clean answer and warrants its own verification.

Both recorded as deferred in the commit message so they do not evaporate.

Gate on this head: ruff, `ruff format --check`, mypy (strict), `lint-imports`, `zizmor --pedantic`, and 2411 tests across unit/adapters/topology, all green locally. Every new test was verified to fail against the old state before being trusted.

guru-xorciseai:
Second pass in `67b5cdf`. I ran an adversarial review over my own fixes rather than stopping at yours, and it found four more — **two of them in code the previous commit had just touched**, which is the argument for doing it.

**1. The malformed-title annotation ran words together.** The block is folded into one line with `tr -d '\n'`, so every continuation needs its own leading space. It did not have them:

```
Titles follow ConventionalCommits: ... Valid types:feat fix docs ...
This title is normally whatlands on main ... corrected withoutrewriting history.
```

Four broken seams on the *common* failure — the one an author actually has to read. Two of them predate my last commit; two I added by lengthening the text, which is worse. And the two error blocks further down the same file already used the leading-space convention, so this was plain inconsistency. Fixed and re-simulated.

**2. The label scan I added missed issue events — including the file its own comment claimed to cover.** I anchored the regex on `github.event.pull_request.labels.*.name`. But `issue-contract.yml:35` gates on `github.event.issue.labels.*.name` (`ci-nightly`) — a different event object. `ci-nightly` was surviving only via nightly.yml's `gh label create` and `--label`; strip those and the declaration could be deleted with the suite green.

Verified by doing exactly that — removing nightly's references and deleting the declaration:

```
AssertionError: ... not declared in .github/labels.yml ...:
  {'.github/workflows/issue-contract.yml': ['ci-nightly']}
```

which the `pull_request`-only pattern could never have produced. Now matches both event objects.

Worth saying plainly: writing a comment that overstated what the code checked is the same defect this branch exists to fix, one level up from where you found it.

**3. Nothing bound the Conventional-Commits *type* list**, though the label list beside it was bound meticulously. It is copied in `pr-contract.yml`, `CLAUDE.md` and `CONTRIBUTING.md`, and dependabot's `commit-message.prefix: build` has to stay inside it — bots are exempt from the label half only, so `build(deps): …` satisfies the title check unaided or every dependency update fails. Three tests now bind the prefix, CLAUDE.md's copy, and CONTRIBUTING.md's coverage.

**4. The 7-day cooldown was documented as a fact and checked by nothing.** CONTRIBUTING.md says "Dependabot enforces this automatically via `cooldown.default-days: 7`"; no test read the value, so dropping it from a block would leave the guarantee documented and gone. Bound across all three ecosystems.

Also took the `comm -23 … | grep -q` in label-sync's extras report — same exits-on-first-match shape as the issue-contract guard, harmless at this data volume, but it also ran `comm` twice to print what it had just tested. Captured once.

Each new binding was verified to **fail** against the unfixed state before I trusted it: a drifted commit prefix, a type dropped from CLAUDE.md, a zeroed cooldown, and an issue-event-only label reference each fail their own test.

---

One correction to my own process, since it is the kind of thing worth admitting rather than quietly dropping. I ran a second adversarial agent whose report claimed most of your findings were "fabricated" and pointed at lines that already contained the fix. That was an artefact of *my* orchestration, not a finding: I started editing files while the verification was still running, so that pass read the post-fix tree and mistook the repaired code for the original. It says nothing about your review, which I had already verified line by line against `adaf76e` with recorded evidence — the 141 exit code, the 14-green deletion, the regex dry-runs and the live repository settings.

Gate on this head: ruff, `ruff format --check`, mypy (strict), `lint-imports`, `zizmor --pedantic`, 2415 tests across unit/adapters/topology. Test count went 14 → 27 over the two commits.

guru-xorciseai:
@christo-xorciseai all done — `244d812` and `67b5cdf`, CI green. Thanks for this, it was a genuinely good review; I checked every claim against the tree before touching anything and they all held up, several exactly.

The ones that would have bitten:

- the SIGPIPE guard — your repro gave me 141 paginated and 0 single-page, so it would have slipped past every short-thread test
- the job name, so the context is actually registerable
- the label scan, which turned out to bind almost nothing (deleting `full-ci` and `ci-nightly` really did leave all 14 tests green, exactly as you called it)
- `labeled` dropped from ci.yml, the case-insensitive match, the newline guard, the branch guard, the dependabot double-label, and the stale docs

Two small pushbacks, both in the threads: your Revert fix doesn't quite reach `Revert "…"` titles (it needs its own branch, not a wider scope class), and I kept the bot exemption — Dependabot labels are creation-time only, so removing it could wedge an old open PR.

I also ran a pass over my own fixes afterwards, which caught two things I'd broken myself: the malformed-title annotation was running words together (`ConventionalCommits`, `withoutrewriting`), and my new label scan only matched `pull_request` events while `issue-contract` gates on `github.event.issue` — so my own comment overclaimed what it covered. Both fixed. Tests went 14 → 27.

Left alone on purpose: the `tests/_helpers.py` consolidation, and the label-sync render — your `uv run scripts/…` suggestion would pull in the whole project dependency set to print 17 rows, so `yq` is the right answer and wants its own change.

**Not merging until you've had another look.** Two I'd rather you called than me:

1. registering `pr-contract` as a required check — happy to do it, but after this lands, since the rename leaves the old context behind on anything already reported
2. squash-only + `PR_TITLE`, or leave the claim softened as it is now

Over to you.

christo-xorciseai (approved):
Looks good to me 👍

PR #96, opened 2026-09-09
feat(docs): generate the documentation's screenshots from the real console

guru-xorciseai:
## Why

The screenshots on docs.xorcise.ai were shot by hand and had no way to notice the product moved. All twelve are dated **4 Aug**; the console design system landed **21 Aug** in #59. So today the live docs show:

- a sidebar reading **"Challenges"** — renamed to "Missions" back in July
- the pre-design-system **uppercase button labels** that #59 replaced
- an Agents page full of test debris: `test-cli-renamed`, `cli-json-emmit`, `walking-skeleton`, `b`

Nobody did anything wrong. Hand-shot images rot silently, and these rotted in public.

## What this adds

**`scripts/seed_docs_state.py`** — drives a throwaway instance through the REST API and OTLP receiver until every photographed page has real content: three agents on the three built-in harnesses, two missions installed against 26 available, and a run replaying a genuinely captured claude-code trace from `tests/fixtures/otlp/`. Only the `xorcise.run_id` resource attribute is rewritten; nothing else about the capture is touched.

It seeds through the same public surfaces an operator uses, so a change that breaks seeding is a change that broke the product — not a stale fixture. It refuses `:3001` without an explicit override, because it registers agents and starts runs.

**`frontend/e2e/docs-screenshots.spec.ts`** — drives the real UI. Not part of `npm run test:e2e` (it needs seeded state and writes files), so it gets its own `npm run screenshots`.

## Two design decisions worth reviewing

**Element-scoped crops, not pixel boxes.** A moved button still yields a correct crop; a hardcoded rectangle silently photographs the wrong thing.

**No baked-in numbered markers.** An annotated PNG can't be regenerated without a human re-annotating it — the exact rot this replaces. Crop to the control instead. It turned out several console forms already carry their own numbered steps, which is better than anything painted on top.

## Two limits, recorded in the spec

Found by hitting them, and written down so they aren't rediscovered:

- Under `--stub` the run environment never leaves `starting`, so `awaitingAgent` is never true and the connect prompt + launch-mode toggle never mount. Sending telemetry doesn't help — that ends the awaiting state from the other side.
- With no judge configured a run grades 0%, so the result and performance pages photograph poorly.

Both shots need a real instance. The spec **skips them naming the reason** rather than writing a misleading image.

## Verification

Ran end to end against a throwaway stub instance (`XORCISE_HOME` + port 3051, isolated from the default):

```
12 passed (9.6s)
```

Twelve images produced. Four are already merged into the docs repo — `agents`, `missions`, `runs`, `traces` plus two new step crops — after visual inspection. `dashboard`, `setup`, `results`, `performance` and `settings` were **held back**: under stub mode `/ui/` routes to Setup with a "Docker ACTION NEEDED" banner, and the graded pages show 0%. They need a real-instance pass.

## Note for the reviewer

Output is gitignored here. The PNGs live in the docs repository; a second copy beside the generator would fork the source of truth and let the two drift — the problem this exists to solve.

### DIFF
--- a/frontend/.gitignore
+++ b/frontend/.gitignore
@@ -4,3 +4,7 @@ out/
 *.tsbuildinfo
 playwright-report/
 test-results/
+# Generated documentation screenshots. They are an OUTPUT of e2e/docs-screenshots.spec.ts and
+# their home is the docs repository's images/ — committing them here would fork the source of
+# truth and let the two copies drift apart, which is the problem the generator exists to solve.
+docs-screenshots/
--- a/frontend/e2e/docs-screenshots.spec.ts
+++ b/frontend/e2e/docs-screenshots.spec.ts
@@ -1,0 +1,186 @@
+import { test, expect, type Page } from "@playwright/test";
+import { mkdirSync } from "node:fs";
+import { join } from "node:path";
+
+/**
+ * The documentation's screenshots, generated from the real console.
+ *
+ * Every image on docs.xorcise.ai used to be shot by hand. By September they were five weeks
+ * stale and showed a console that no longer existed — the sidebar still said "Challenges"
+ * (renamed to "Missions" in July), the buttons still used the pre-design-system uppercase
+ * labels, and the Agents page was full of test debris like `test-cli-renamed`. Nobody had
+ * done anything wrong; hand-shot images simply have no way to notice the product moved.
+ *
+ * So they are generated instead. This spec drives the real UI against a seeded throwaway
+ * instance and writes PNGs the docs repository consumes directly.
+ *
+ * PREREQUISITES — this is not part of `npm run test:e2e`, because it needs seeded state and
+ * writes files. Run it deliberately:
+ *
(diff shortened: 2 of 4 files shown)


guru-xorciseai:
Closing: this belongs in the docs repository, not here.

`pip install xorcise` ships the console — the static export is force-included into the wheel (`pyproject.toml` `artifacts`), and `scripts/check_wheel_contents.py` already asserts that `pip install xorcise` gives a working `xorcise up`. So the pipeline never needed a clone of this repo.

More importantly the placement was backwards. A spec living beside the UI runs against **this repo's `main`**, so it would photograph unreleased behaviour — while the docs site's policy is that it describes the latest *release*. Installing from PyPI fixes that, and it lets the generator open its own pull requests, which a cross-repo pipeline could not.

Rebuilt in Python Playwright at xorcise-ai/docs#14. Nothing is lost from this branch; the OTLP fixture is vendored there.

PR #97, opened 2026-09-09
fix(cli): status and ui probe the loopback planes with trust_env=False

### LINKED ISSUE #86: status and ui report a healthy server as stopped when HTTP_PROXY is set
## Symptom

With `HTTP_PROXY` (or `ALL_PROXY`) set in the shell, `xorcise status` and `xorcise ui` report a
perfectly healthy running server as stopped:

```console
$ xorcise up --stub
xorcise up — UI at http://127.0.0.1:3971/ui

$ xorcise status                       # no proxy — correct
 REST API        Healthy   http://127.0.0.1:3971
 OTLP receiver   Healthy   http://127.0.0.1:4971
All services are healthy.

$ HTTP_PROXY=http://127.0.0.1:9 xorcise status      # same live server
 Interface       Stopped   http://127.0.0.1:3971/ui
 REST API        Stopped   http://127.0.0.1:3971
 OTLP receiver   Stopped   http://127.0.0.1:4971
REST API, OTLP receiver: not responding — start XORCISE with xorcise up

$ HTTP_PROXY=http://127.0.0.1:9 xorcise ui
http://127.0.0.1:3971/ui
XORCISE is not running — start it with xorcise up
```

The server is fine throughout — only the report is wrong. The advice it then gives
("start XORCISE with `xorcise up`") sends the operator down a false trail.

## Root cause

`probe_channel` (`src/xorcise/core/cli/_diagnostics.py:232-238`) builds its request with httpx's
default `trust_env=True`:

```python
def probe_channel(name: str, url: str, ok_statuses: tuple[int, ...] = (200,)) -> Check:
    try:
        code = httpx.get(url, timeout=1).status_code
    except httpx.HTTPError:
        return Check(name, False, "down")
```

httpx honours `HTTP_PROXY`/`ALL_PROXY` and applies **no implicit loopback bypass** — verified
against the pinned httpx 0.28.1:

```
loopback, HTTP_PROXY set                  -> pool=HTTPProxy      <-- goes to the proxy
loopback, no proxy env                    -> pool=ConnectionPool
loopback, HTTP_PROXY + NO_PROXY=127.0.0.1 -> pool=ConnectionPool
```

So the health probe for `http://127.0.0.1:<port>/api/health` is routed to the proxy, fails, and
is reported as `down`.

## Fix

One argument, at the single call site:

```diff
-        code = httpx.get(url, timeout=1).status_code
+        # A loopback probe must never be routed through HTTP_PROXY/ALL_PROXY: httpx honours them
+        # by default and applies no implicit loopback bypass, so a proxied shell reads a healthy
+        # local server as down.
+        code = httpx.get(url, timeout=1, trust_env=False).status_code
```

`probe_channel` is only ever called with loopback plane URLs, so there is no caller that wants
the proxy:

- `lifecycle.py:548` — `ui`
- `lifecycle.py:850-851` — `status`

### Test

Assert the probe does not consult the environment, in the same shape as
`test_probes_never_use_the_proxy_environment` in `tests/unit/test_lifecycle_guards.py` (added by
#82 for the sibling probes) — e.g. monkeypatch `httpx.get` and assert `trust_env is False` in the
call kwargs.

## Scope

- **Affected:** `xorcise status`, `xorcise ui`
- **Not affected:** `xorcise doctor` — its Services block uses `ports_in_use`, not
  `probe_channel`, and correctly reports `✓ port 3971 (rest): in use by the running XORCISE
  service` under the same proxy.
- **Not affected:** `up`, `down`, `db upgrade` — #82 set `trust_env=False` on all four of their
  loopback probes (`_probe_system`, `_foreign_instance_home`, `_note_foreign_instance`, and
  `up`'s health poll).

## Provenance

Pre-existing, **not** a regression from #80/#81/#82. `probe_channel` is byte-identical on `main`
and appears zero times in all three of those diffs; the line dates to the initial commit
(1d90225, 2026-08-03).

#82 fixed this same class of bug in the four places where a wrong answer causes real damage
(a duplicate instance on one SQLite file, or a migration under a live server). This is the fifth
instance, in a file that work had no reason to touch — where the consequence is a false report
rather than corruption. Raised so it is not lost now that its siblings are fixed.

## Priority

Low. Only reachable with `HTTP_PROXY`/`ALL_PROXY` set and `NO_PROXY` not covering the loopback —
but a corporate shell or dev container hits it immediately, and the symptom is maximally
misleading.

christo-xorciseai:
Closes #86.

## The defect

With `HTTP_PROXY` (or `ALL_PROXY`) set and `NO_PROXY` not covering the loopback, `xorcise status`
reported a healthy running server as **Stopped** on every plane, and `xorcise ui` told the operator
to start a server that was already up. `probe_channel` built its request with httpx's default
`trust_env=True`; httpx applies no implicit loopback bypass, so the GET for
`http://127.0.0.1:<port>/api/health` went to the proxy, failed, and read as "down".

## The fix

One argument at the single call site — `trust_env=False` — the same fix #82 applied to the four
lifecycle probes. `probe_channel` is only ever called with loopback plane URLs (`ui`, `status`), so
no caller wants the proxy. A unit test pins that the probe does not consult the environment.

Not affected, and unchanged: `doctor` (uses `ports_in_use`), and `up` / `down` / `db upgrade` (#82).

## Verification

| Check | Result |
| --- | --- |
| `pytest tests/unit/test_diagnostics.py tests/unit/test_cli_ux.py` | 109 passed |
| `mypy` / `ruff check` / `ruff format --check` | clean |

**Live**, against `xorcise up --stub` on relocated ports, with `HTTP_PROXY=http://127.0.0.1:9`
(a dead proxy) in the environment:

- `xorcise status` → every plane **Healthy**, "All services are healthy." (was Stopped × 3)
- `xorcise ui` → prints the URL, no "XORCISE is not running" (was the false advice)
- Control: the pre-fix request shape (`httpx.get(url, timeout=1)`) under the same proxy →
  `ConnectError`; the fixed shape → `200`.


guru-xorciseai (commented):
Pulled the stack and ran this one against a real stub instance on relocated ports. The fix does what you say it does — with `HTTP_PROXY=http://127.0.0.1:9` in the environment and no `NO_PROXY`, `status` reports every plane Healthy and `ui` prints the URL instead of the false advice. I re-ran your control too, and it holds:

```
PRE-FIX  (trust_env default=True): ConnectError: [Errno 111] Connection refused
POST-FIX (trust_env=False):        HTTP 200
```

Good root cause. The "httpx applies no implicit loopback bypass" detail is the sort of thing that stays invisible until someone runs this behind a corporate proxy, and the docstring you left on `probe_channel` explains it better than the issue did.

Two other places in the same bug class, though. The first one I think is more urgent *because* of this PR, not despite it.

### 1. `rest_client.py` — every other CLI command still goes through the proxy

`probe_channel` is fixed, but `RestClient` builds all of its calls with plain `httpx.get`/`post`/`put`/`delete` (lines 63, 80, 90, 94, 111, 126, 132, 138, 144) against `self.base_url`, which `default_base_url()` builds as `f"http://{s.host}:{port}/api"` — loopback, same as the planes you just fixed.

Same instance, same dead proxy, same shell:

```
xorcise status        →  "All services are healthy."                          exit 0
xorcise run list      →  "cannot reach the XORCISE service — is it running?"  exit 1
xorcise agent list    →  same                                                 exit 1
xorcise mission list  →  same                                                 exit 1
```

Before this PR they were at least *consistent* — everything looked down, and the operator went looking for a dead server. Now `status` says healthy while everything else says it can't reach the service, and `_SERVICE_DOWN`'s own remediation text says `check: xorcise status`, which cheerfully confirms all is well. That's a harder hole to climb out of than the original bug, and it's reachable by exactly the operator #86 was filed for.

I did check whether a blanket `trust_env=False` there would break a legitimate remote setup, because that would make the narrower host-based bypass the right fix instead. As far as I can tell it wouldn't, today: the CHANGELOG says the distributed topology relocates the mission plane while "the control plane and console stay local", `XORCISE_HOST` is a bind-widening switch for the server's own socket rather than a way to aim the CLI at another instance, and `xorcise remote` is still a hidden stub that always answers "no remotes registered". So the same one-liner applies cleanly. If you'd rather leave room for a future remote CLI, a comment saying so would help the next person, since the constraint isn't obvious from the code.

### 2. `system_view.py:125` — the GUI's System card

`_probe()` is the same shape, unpatched:

```python
code = httpx.get(url, timeout=1).status_code
```

It's called at lines 249 and 252 for the `rest` and `otlp` planes, with locations built from `settings.host` — loopback again. It backs `build_system_info()` behind `GET /system`, which the file's own comment says every GUI page polls. `up` launches the server with `subprocess.Popen` without stripping the environment, so a proxied shell's `HTTP_PROXY` is inherited by the server process itself. Net effect is the server telling the operator that the REST plane is down, in a response it is itself serving.

Neither of these blocks this PR — it's strictly better than what's on main, and I'd be happy to see it go in as-is and have the rest follow. But #86 reads like the whole class, so it's worth deciding out loud whether that issue closes here or stays open for the other two sites.

Everything else checks out: `probe_channel`'s callers are only `ui`/`status`, `external_control_plane` correctly keeps proxy support since it's genuinely remote, and the lanes are green locally (unit 2232, adapters 153, topology 31, e2e 3, integration 30, frontend 735/735, mypy 494 files, ruff + `lint-imports` clean).

christo-xorciseai:
Thanks — both follow-ups applied in b069d26, and I agree the first one was made more urgent by this PR rather than despite it.

**`rest_client.py`** — `trust_env=False` at all nine call sites (the seven client methods, the `get_run_result` retry, and `_warn_if_foreign_instance`). Spelled out per site rather than through a shared kwargs dict, because httpx's overloads do not type-check a `**dict`. The module comment records the constraint you asked for: `default_base_url` is built from the bind address, the only remote source is the `xorcise remote` stub, and a future remote CLI must decide about proxies on its own path — this one never consults them.

**`system_view.py`** — same one argument on `_probe`, with a note that `up` does not scrub the environment it spawns the server with.

Tests: `test_rest_client_never_consults_the_proxy_environment` drives every method against recording fakes and asserts `trust_env is False` on all eight calls; `test_plane_probes_never_consult_the_proxy_environment` does the same through `build_system_info`. Five strict-signature fakes in the CLI tests gained `**kwargs`.

Live, same shape as your run — stub instance on relocated ports, `HTTP_PROXY=http://127.0.0.1:9`, and this time the server was *started* in that environment (confirmed via `/proc/<pid>/environ`):

```
status         exit=0  All services are healthy.
run list       exit=0  no runs yet — …
agent list     exit=0  no agents registered — …
mission list   exit=0
GET /api/system  rest ok=True   otlp ok=True        (was: both "down")
```

`external_control_plane` and the PyPI update check keep proxy support, as before. With these two sites done I'd close #86 on this PR for the whole class; the "Closes #86" already in the body stands.

Lanes at this tip: unit 2212, adapters + topology 179, mypy/ruff/lint-imports clean.

guru-xorciseai:
Both fixes verified, and the production change is right — but the new test is red in CI, and it's red on all three PRs since they all inherit this commit.

### The fixes themselves check out

Started the stub server *inside* the proxied environment the way you did, and confirmed via `/proc/<pid>/environ` that it really inherited `HTTP_PROXY`/`ALL_PROXY` before trusting the result:

```
status         exit=0
run list       exit=0
agent list     exit=0
ui             exit=0
GET /api/system  → {"rest": true, "otlp": true}     (was both down)
```

(`mission list` 500s here, but that's my sandbox having no outbound network — `catalog/http.py` failing to reach the remote catalog, and it reproduces with the proxy unset. Right call leaving that one on `trust_env=True`, since it's genuinely remote.)

### `test_rest_client_never_consults_the_proxy_environment` is order-dependent

```
FAILED tests/unit/test_cli_commands.py::test_rest_client_never_consults_the_proxy_environment
AssertionError: assert 7 == 8
```

It's `_warn_if_foreign_instance`'s `_FOREIGN_CHECKED` — module-global, never reset between tests. Any earlier test in the same xdist worker that made a successful `RestClient` call for the same home leaves that home in the set, so your last call hits `if home in _FOREIGN_CHECKED: return` before it ever reaches `httpx.get`, and you count 7.

The whole suite passes for me locally (2238) and fails in CI, which is just the shard split landing differently — so it'll keep flipping rather than stay red. Pinned it deterministically:

```python
def test_a_clean_global_sees_eight(monkeypatch):
    monkeypatch.setattr(rc, "_FOREIGN_CHECKED", set())
    assert len(_drive(monkeypatch)) == 8          # passes

def test_b_a_poisoned_global_sees_seven(monkeypatch):
    monkeypatch.setattr(rc, "_FOREIGN_CHECKED", {str(xorcise_home())})
    assert len(_drive(monkeypatch)) == 8          # AssertionError: got 7 — the CI failure
```

Worth saying clearly: **all 7 calls it does record carry `trust_env: False`.** Nothing is wrong with the fix — it's only the count assertion that trips, so this is a test fix, not a rethink.

One line, and `test_cli_ux.py:830` already does exactly this:

```diff
     for verb in ("get", "post", "put", "delete"):
         monkeypatch.setattr(httpx, verb, _record(verb))
+    monkeypatch.setattr(rest_client, "_FOREIGN_CHECKED", set())
     client = RestClient("http://127.0.0.1:1/api")
```

Agreed on closing #86 here for the whole class — with `rest_client` and the System card done, I can't find another loopback site that still consults the environment.

guru-xorciseai:
Bumping this one — the stack is still red and blocked on it, so here's the exact patch rather than just the diagnosis.

`test_rest_client_never_consults_the_proxy_environment` imports the two names directly and has no module handle, so the string form is the one that applies without a new import — and it's the same form the test already uses for `pid_file` a few lines down:

```diff
     for verb in ("get", "post", "put", "delete"):
         monkeypatch.setattr(httpx, verb, _record(verb))
+    # `_FOREIGN_CHECKED` is process-global and never reset, so an earlier test in the same xdist
+    # worker that made a successful RestClient call for this home short-circuits the eighth call
+    # before it reaches httpx — 7, not 8. test_cli_ux.py:830 resets it for the same reason.
+    monkeypatch.setattr("xorcise.core.cli.rest_client._FOREIGN_CHECKED", set())
     client = RestClient("http://127.0.0.1:1/api")
```

To restate the important half so it doesn't get lost: **the production fix is correct.** All 7 calls the test does record already carry `trust_env: False`. It's only the count that trips, so this is a test fix and nothing about `rest_client.py` or `system_view.py` needs revisiting.

Two other things that need to happen in the same pass, or this stays red:

1. **#97 is `BEHIND` main.** #94 merged after you branched, so it needs a rebase onto `main` regardless of the test.
2. **The fix has to travel down the stack.** #98 and #99 both carry this commit, and both fail on this same test — `assert 7 == 8`, same line, all three. After #97 moves, #98 needs rebasing onto it and #99 onto #98, or they keep the stale copy and stay red.

Everything else is verified and green at the tip — adapters 153, topology 31, integration 30, e2e 3, mypy 494 files, ruff and `lint-imports` clean, and all three of your fixes confirmed live (details in my earlier comments on each PR). This one test is the only thing between the stack and approval.

christo-xorciseai:
Applied exactly as you wrote it in bf91d54 — the string form, since the test imports the two names directly. Before touching it I pinned the failure the way you did: poison `_FOREIGN_CHECKED` with this home and run the real test body → `assert 7 == 8` on the old body, 8 with the reset. Agreed on the important half: all seven recorded calls carried `trust_env: False`, so nothing in `rest_client.py` or `system_view.py` changed.

Rebased onto `main` (now on 3c14a88, so #94's label contract and its topology tests are in), and the fix has travelled down: #98 is rebased onto this branch and #99 onto #98. Unit lane at each tip: 2212 / 2227 / 2238; adapters + topology 211 at the stack tip (the new `test_label_contract` included); mypy, ruff and `lint-imports` clean.

On your `mission list` 500: same reading here — `catalog/http.py` reaching for the remote catalog, genuinely remote, correctly still `trust_env=True`. #86 closes on this PR for the whole class.

guru-xorciseai (approved):
Approving. Verified live at `bf91d54`, not just read.

**The test fix is correct.** Pinned it the way you did before trusting it: poisoned `_FOREIGN_CHECKED` with this home and ran the real body — `assert 7 == 8` without the reset, 8 with it. The string form is right given the test imports the two names directly.

**The production fix, re-verified end to end.** Stub instance on relocated ports (3111/4111), isolated `XORCISE_HOME`, and the server *spawned inside* the proxied environment — confirmed via `/proc/<pid>/environ` that it really inherited them, with `NO_PROXY` empty so there is no loopback bypass:

```
HTTP_PROXY=http://127.0.0.1:9   ALL_PROXY=http://127.0.0.1:9   NO_PROXY=
```

```
xorcise status   exit=0   REST API / OTLP receiver / Interface  all Healthy
GET /api/system           "rest": ok,  "otlp": ok
```

And the counterfactual against that same live server, which is what makes the one argument the whole fix:

```
trust_env=True   -> ConnectError: [Errno 111] Connection refused
trust_env=False  -> HTTP 200
```

**The boundary is drawn in the right place**, which is the part worth more than the call sites. I audited every `httpx` construction in `src/`: everything aimed at a loopback plane carries `trust_env=False` (including both multi-line probes in `lifecycle.py`), and everything genuinely remote — `catalog/http.py`, `_update.py`'s PyPI fetch, `judge_model.py`, and `external_control_plane` — deliberately does not. `catalog` 500s in my sandbox for want of outbound network and reproduces with the proxy unset, so that one is correctly left alone.

The module comment recording *why* a future remote CLI must decide on its own path is the thing that stops this being re-broken later.

Full suite on locked deps at the stack tip: **2482 passed, 4 skipped, 0 failed**; mypy 495 files, ruff, `ruff format --check` and `lint-imports` all clean.

PR #98, opened 2026-09-09
fix(rest): the readiness gate fails closed, keeps the evidence, tears down once, and waits 300 s

### LINKED ISSUE #43: serve runs every startup hook once per bind address: duplicated watchdogs, plus adjacent readiness/preflight findings

christo-xorciseai:
Refs #43 — the four findings left open after #81 (2, 3, 4, 5). Stacked on #97 (`fix/status-ui-proxy`); merge that first.

## What was wrong

`#81` fixed #43's root cause (one uvicorn Server per app) and finding 1 (IPv6-blind preflight). The
readiness gate itself still had four defects, all reproduced against the real daemon before fixing:

| # | Finding | What actually happened |
| --- | --- | --- |
| 3 | **Readiness fails open.** | `DockerSdkDriver.compose_service_states` ran the nested `compose ps` and parsed whatever came back. When the inner daemon is not running, `compose ps` exits 1 with `failed to connect to the docker API at unix:///var/run/docker.sock` — which parsed to `()`, which `RunnerControlService.status` read as **READY**. A wedged daemon latched `_ever_ready` and squatted its subnet until the 30-minute budget watchdog. |
| 2 | **Close-out destroys the evidence.** | `_check_one` force-removed the outer container first and terminated second. The container's logs are the only record of what `compose up` (or the inner daemon) said; diagnosing a deploy failure meant disabling the gate and reproducing by hand. |
| 4 | **Double teardown.** | The gate called `_release()` (control.teardown + fence teardown) and then `terminate_run` → `grade_and_record` → `teardown_run` released the same environment again in a `finally` — the `409 removal … already in progress` in the logs, and twice the close-out latency. |
| 5 | **90 s is too tight.** | Measured under parallel load: layered-alibi needed 66 s, breachpoint >110 s just to load `images.tar` before `compose up` began. Legitimate bring-ups were closed out as `deploy_failed`. |

## What this does

**Finding 3 — the driver now distinguishes "nothing to say" from "asked and no answer".**
`compose_service_states` returns `None` when the outer container is gone, the exec raised, or
`compose ps` exited non-zero; `()` keeps meaning "this driver has no information / unparseable
output → degrade to ready" (stub mode is unchanged). `RunnerControlService.status` reads `None` as
**PENDING** with `detail="the mission environment's inner Docker daemon is not answering"`, so the
existing window + strikes close it out. `StatusResult` gains a `detail` field (which service is still
starting, the exit code, the daemon) and `classify_environment` reports it — it is what the run page
shows and what a close-out records.

**Finding 2 — evidence first, release after, recorded on the run.**
A new `DockerDriver.container_logs` reads the outer container's log tail (60 lines) and, while it is
still running, the inner daemon's (`/var/log/dockerd.log`, kept in sync with the entrypoint). It rides
`ControlPort.environment_logs` (non-abstract, empty by default) through the adapter to the gate,
which reads it **before** handing the run to `terminate_run`. The gate's verdict plus that evidence is
recorded as `runs.terminal_detail` (migration `0003`, nullable `Text`), exposed on `RunEntry`, capped
at 8 KB keeping the newest lines, and rendered on the run page under the Deploy-failed badge — the
place "Inspect run" already sends the operator. A failure to read the logs never blocks the close-out.

**Finding 4 — one teardown.** The gate no longer releases the environment itself. `terminate_run`
(→ `grade_and_record` → `teardown_run`, in a `finally`) is the single teardown every terminal path
shares, and now documents that callers must not pre-empt it.

**Finding 5 — the default window is 300 s.** The window only bounds a bring-up that never
completes: an environment that actually dies is closed out at once regardless, so the longer default
costs nothing on a hard failure. `XORCISE_READINESS_TIMEOUT_SECONDS` still overrides; `0` still
disables.

Also regenerates `frontend/openapi.json` + `schema.ts` (the snapshot is `json.dumps(sort_keys=True,
indent=2)`; it was two PRs stale — `agent_ingress`/`allow_egress` from #63 and `pid` from #82 come
along).

## Verification

| Check | Result |
| --- | --- |
| `pytest -m unit -n auto` | 2224 passed (a first run had two order-sensitive catalog tests fail under xdist while the frontend suite competed for CPU; both pass alone and on the re-run) |
| `pytest -m adapters` / `-m topology` | 153 / 31 passed |
| `mypy` (strict, src + tests) | clean, 494 files |
| `ruff check` / `ruff format --check` / `lint-imports` | clean / clean / 4 contracts kept |
| frontend `tsc --noEmit` / `vitest run` | clean / 735 passed in 100 files |

**Live, against the real Docker daemon, real driver, real gate, real SQLite** (only the tailnet fence
faked — `pr43_live.py` in the review notes):

- **A — wedged inner daemon** (a `docker:29.7.1-dind` container whose daemon never starts): the
  nested `compose ps` exits 1 with `failed to connect to the docker API…`. Old parse → `()` → READY.
  Now → `None` → `pending`, detail *"the mission environment's inner Docker daemon is not
  answering"*; past the window the gate closes it out on the second strike, `terminal_detail` carries
  the verdict and the container's log tail, **one** teardown, container removed.
- **B — environment dies at deploy** (exits 1 after printing `compose up: service web exited (1):
  bind: address already in use`): `failed` with the exit code, closed out inside the window,
  `terminal_detail` contains the compose error, **one** teardown, container removed.

## Not in this PR

- A CLI rendering of `terminal_detail`. There is no `GET /runs/{id}` route, so `run status` (which
  reads `/result`) cannot show it without either a new route or the result view carrying run fields —
  a design choice worth making on its own. `xorcise run list --json` and the UI carry it today.
- Scaling the window with image size (finding 5's alternative). The fixed 300 s covers the measured
  cases; a size-aware window needs the image size on the run row first.


guru-xorciseai (commented):
This is a lot of careful work and it mostly reads that way. I went through all four findings against the real code rather than the diff, and the parts I most expected to be wrong were right: `compose_service_states`' tri-state is consumed with an explicit `is None` at the single production call site (so `()` still degrades to ready and stub mode is genuinely untouched), `closeout_detail` drops the *oldest* evidence and re-cuts to a line boundary, migration `0003` chains cleanly off `0002` and uses `batch_alter_table` so the SQLite downgrade actually works, and `tick()` is offloaded with `asyncio.to_thread` so none of the new Docker I/O touches the event loop. `ROUTER_WAIT` being a module constant with one producer and one consumer is the right call — I went looking for a brittle string match and didn't find one.

One thing I do think needs fixing before this lands, and one smaller thing about the PR body.

### Finding 4's "one teardown" isn't guaranteed on every path — and the gate no longer has a backstop

This is the one I'd hold the PR for. `terminate_run`'s docstring now promises that `grade_and_record` "is also where the run's environment is released (run_teardown, in a `finally`) — the ONE teardown every terminal path shares, so callers must not release it themselves first." The gate took that at its word and dropped `_release()`. But the `finally` in `_grade_run` only wraps `record_result`:

```python
if reporting.get_result(run_id) is not None:
    return                              # ← outside the guard
_drain_and_seal_telemetry(run_id)       # ← outside the guard
...
try:
    reporting.record_result(...)
finally:
    teardown_run(run_id)                # ← only reached if we get this far
```

`_drain_and_seal_telemetry` goes through `SqliteSealStore.is_sealed`/`.seal`, both via `session_scope()`, which does `except Exception: session.rollback(); raise` — it re-raises. So a transient DB error there propagates out through `grade_and_record`'s `finally` (which only drops the in-flight slot) and out through `terminate_run`, and `teardown_run` never runs.

The part that makes it stick is that `seal_terminal` has already committed `state='terminal'` by then, and `deployed_non_terminal_runs` filters on `state != "terminal"` — so the run has already left the gate's pending set. `tick()`'s per-run `except Exception` logs "retried next tick", but there is no next tick for that run. The container and its tailnet subnet stay allocated until a server restart hits `regrade_orphaned_terminal_runs`.

Worth saying plainly: before this PR the gate's own `_release()` covered this, each half wrapped in its own log-and-swallow. So this is a regression introduced by consolidating the teardown, not a pre-existing hole.

Here's a repro, modelled on your `test_record_failure_still_tears_down_environment` but injecting the failure one step earlier:

```python
def test_drain_failure_leaks_the_environment(migrated_home, monkeypatch) -> None:
    import xorcise.core.rest.run_teardown as run_teardown
    import xorcise.core.rest.run_terminate as rt

    torn: list[str] = []
    monkeypatch.setattr(run_teardown, "teardown_run", lambda rid: torn.append(rid))

    r = runs.create_run(agent_id="a1", mission="c", budget_seconds=600)
    rt.seal_terminal(r.run_id, "deploy_failed", _now())

    def _boom(*a: object, **k: object) -> None:
        raise RuntimeError("database is locked")

    monkeypatch.setattr(rt, "_drain_and_seal_telemetry", _boom)
    with pytest.raises(RuntimeError, match="database is locked"):
        rt.grade_and_record(r.run_id)

    assert runs.terminal_state(r.run_id)[0] is True   # already terminal — no re-scan
    assert torn == [r.run_id]
```

```
AssertionError: LEAK: teardown_run was never called (torn=[])
```

`database is locked` isn't a hypothetical here — the readiness gate, the budget watchdog and the REST handlers all share one SQLite file through `session_scope()`, and I couldn't find WAL mode or a `busy_timeout` set anywhere in `engine.py`.

Fix that seems cleanest: widen the `try`/`finally` in `_grade_run` to cover the whole body from just after the early-return guards through `record_result`, so the invariant your docstring states is actually the one the code enforces. Your two existing teardown tests (`..._grading_crash_...` and `..._record_failure_...`) already establish the pattern; this would be a third alongside them.

### Smaller: the 300 s justification doesn't cover the case Finding 3 adds

`config/__init__.py` justifies 90 → 300 with "an environment that actually dies is closed out at once regardless — so a long window costs nothing on a hard failure." That's true for a container that exited, which `classify_environment` maps to `failed` and which skips the window entirely. But the wedged-inner-daemon case this same PR introduces returns `PENDING` with "the mission environment's inner Docker daemon is not answering", which classifies as `starting` and rides the full window plus two strikes. Your own Case A verification says as much — "past the window the gate closes it out on the second strike".

So the permanently-unusable-daemon case now squats its subnet for ~305 s instead of ~95 s. I don't think the classification is wrong (an unanswering daemon genuinely is ambiguous in a way an exit code isn't, and closing it out fast would be the riskier call), and the tradeoff is disclosed further down the PR body. It's the one-line comment in `config/__init__.py` that overstates it, and that comment is what the next person will read. Worth a clause.

For completeness, things I checked that came back clean: the regenerated `openapi.json` really is a faithful `json.dumps(sort_keys=True, indent=2)` of the live spec (key ordering is right through every new block, so it's generated not hand-edited), `terminal_detail` renders through React's escaping with a null guard, `container_logs` catches every failure mode including the die-between-status-check-and-exec race, and the log capture can't leak mission secrets — `entrypoint.sh` has no `set -x`, compose interpolates `${XORCISE_AUTHKEY}` at `up` time without echoing it, and detached `compose up` doesn't stream service logs, so flags stay inside the inner daemon.

christo-xorciseai:
Good catch, and you are right that it was a regression of this PR's own consolidation, not a pre-existing hole. Fixed in eeb91d8.

**Finding 4 — the guarantee the docstring claimed is now the one the code enforces.** `_grade_run`'s `try` opens immediately after the three early-return guards, so the drain, the judge, the conditions, the stats fold and `record_result` all sit inside the one `finally` that releases the environment. Your reproduction is in as `test_drain_failure_still_tears_down_environment`, beside the two existing teardown tests, and it also asserts the two facts that made this stick: the run is terminal and it is no longer in `deployed_non_terminal_runs`, so the teardown had to happen there. It failed on the previous tip exactly as you showed (`torn=[]`) and passes now.

**The 300 s comment** now says what it costs: an environment whose container exits is closed out at once regardless, while the ambiguous case this PR makes visible — outer container alive, inner daemon never answering — rides the whole window plus the strikes, about 5 minutes instead of 1.5, as the price of not killing a slow but healthy bring-up on a probe blip.

Your aside about `engine.py` is worth more than an aside: no `busy_timeout` and no WAL mode on a file with five kinds of writer. Filed as #107 so it gets fixed at the source rather than survived per call site.

Rebased onto the updated #97 (stack order unchanged). Lanes at this tip: unit 2227, adapters + topology 184, mypy/ruff/lint-imports clean.

guru-xorciseai:
Verified. I ran my original repro unmodified against this tip and it passes now:

```
teardown_run called with: ['0549b6b0d5d84ac1a3834cae14d129fa']
1 passed
```

The `try` opening straight after the three early-return guards is exactly right, and the comment you left explaining *why* nothing below may leave the environment allocated — that the run is already `terminal` so `deployed_non_terminal_runs` will never show it again — is the part that will stop someone re-narrowing it later. Keeping the judge's `except` separate while making the `finally` the actual guarantee is the correct split.

The 300 s comment now says what it costs, which was the whole ask.

Filing the `busy_timeout`/WAL point as #107 rather than working around it per call site is the better instinct — that file having five kinds of writer and no busy timeout is going to keep producing symptoms like this one.

Everything else at this tip is clean: adapters 153, topology 31, integration 30, e2e 3, mypy 494 files, ruff and `lint-imports` all green.

CI here is red only on `test_rest_client_never_consults_the_proxy_environment`, which arrives from #97 — details on that PR. Nothing in this one is implicated; it'll clear when #97's test fix rebases down.

christo-xorciseai:
Rebased onto the updated #97 (b80b5b7 is the same two commits over the new base; the only change arriving is #97's test fix, which is what was turning this PR red). Unit lane at this tip 2227 passed, mypy/ruff/lint-imports clean. Nothing in this PR changed.

guru-xorciseai (dismissed):
Approving. This is the one I re-read hardest, because the consolidation makes one function solely responsible for releasing every environment — so I traced the release path rather than trusting the docstring.

**The single-teardown claim holds.** `role_all.py` wires `terminate=terminate_run`, the synchronous composite, so the gate's hand-over reaches `grade_and_record` → `teardown_run`, and `teardown_run` does exactly what the gate's old `_release` did — `control.teardown` **and** `fence.teardown_run_network`. Nothing is dropped by removing `_release`. I also walked `_grade_run`'s three early returns, since each one skips the `finally`: run absent (nothing allocated), not terminal (still in `deployed_non_terminal_runs`, so the next tick retries), already graded (a previous pass released it). All three are safe, and the comment explaining why the `try` may never be re-narrowed is the part that keeps them safe.

**The `() vs None` tri-state is the real find here.** An environment that cannot list its own services was being read as READY, so a wedged inner daemon passed the gate and squatted its subnet until the budget watchdog fired. Making "asked and could not answer" distinct from "nothing to say" is the correct shape, and `test_status_is_pending_when_the_inner_daemon_cannot_be_asked` pins it at the seam that was wrong.

**Migration 0003** is clean: nullable `Text`, `batch_alter_table` for SQLite, chain linear at 0001 → 0002 → 0003 with no branched `down_revision`, and a real `downgrade`.

**`closeout_detail`** — I checked the arithmetic rather than eyeballing it. For any reachable input the output lands exactly on the 8 000 cap, the marker is accounted for, and starting the tail on a whole line is a nice touch: a torn first line reads as if the log itself were corrupt.

The tests earn particular mention. `test_the_gate_releases_the_environment_through_terminate_only` asserts the *absence* of the double teardown, and `test_close_out_records_why_and_the_evidence_read_before_release` asserts the **ordering** (`log_reads_at_terminate == [1]`) rather than just the outcome — so a future refactor that reads the logs after release fails loudly instead of silently recording nothing.

The 300 s default is justified with measured numbers and, more usefully, states what it costs on the ambiguous case. That is the version of that comment worth keeping.

Full suite on locked deps at the stack tip: **2482 passed, 4 skipped, 0 failed**; mypy 495 files, ruff, `ruff format --check`, `lint-imports` (4 contracts kept) all clean.

guru-xorciseai (approved):
Re-approving after the rebase — the approval above was dismissed by the force-push, and nothing in the change itself moved.

Housekeeping note on what happened to this PR: merging #97 and deleting its branch **auto-closed this one**, because GitHub closes a PR whose base branch is deleted. Nothing was lost — I restored `fix/status-ui-proxy`, reopened this PR with its review history, and retargeted it to `main`. #99 was retargeted to `main` pre-emptively so the same thing cannot happen to it when this branch is deleted.

The branch is now `#97`-free: rebased with `--onto main`, dropping the three commits that went in as the `95c847d` squash, leaving exactly your two commits. Verified rather than assumed — `git diff b80b5b7 HEAD` is **empty**, so the rebased tree is byte-identical to the one I reviewed and ran the suite against, and the diff against `main` is the same 23 files / 782 insertions this PR always had.

PR #99, opened 2026-09-09
fix(cli): doctor probes the address the run's routers dial, not just the headscale container

### LINKED ISSUE #62: doctor: control-plane check verifies the container, not the address routers dial — reports healthy while every run fails

christo-xorciseai:
Closes #62. Stacked on #98 (`fix/readiness-gate-evidence`); merge that first — the readiness
hint here lands in the `terminal_detail` that PR introduces.

## The defect

`doctor` verified the control plane by checking the **container** (`docker exec headscale headscale
version`). Every run's subnet router dials something else: `headscale_url`, which `up` bound to the
host address it detected **at provisioning time** (the docker bridge gateway on Linux, the LAN IP on
macOS, or `XORCISE_HEADSCALE_HOST_IP`). When that address changes — a laptop moving networks, a new
DHCP lease — the container stays healthy and nothing answers at the recorded URL, so:

- `doctor` printed `✓ run network ready (Headscale)` and **"No problems found"**;
- every run's router sat at `NeedsLogin`, never got a tailnet address, and the run ended
  `deploy_failed` with *"not ready within the readiness window"* — nothing named the control plane.

The `control_plane` docstring defended the container check on purpose (run creation shells into that
container regardless of the URL), and it is right: the container is half the dependency. This adds
the other half rather than replacing it.

## What this does

**`control_plane_address(url, current_ip=…)`** (`cli/_diagnostics.py`): probes the recorded URL the
way `external_control_plane` does (any HTTP reply counts; TLS unverified because the local CA is
self-signed; `trust_env=False` because this is a host-local address the routers dial directly — see
#86). On a transport failure the detail names the URL, and when `current_ip` — what `up` would detect
today — differs from the recorded host it says so, because that *is* the diagnosis:

```
✗ control plane address: routers dial https://192.168.0.5:443, which is not answering —
  this host's address is now 192.168.1.20, not 192.168.0.5 — re-provision the control plane on
  this host's current address: xorcise down && xorcise up
```

**`doctor`** runs it after the container check whenever `headscale_url` is set, under the same
`expected_up and not use_stubs` gate. Our own managed plane (URL == the managed block `up` wrote) gets
the address probe with the current host IP; an operator-chosen remote gets `external_control_plane`,
the probe `up` already verifies it with — never the moved-address diagnosis, which would advise
re-provisioning a plane we do not own.

**The readiness gate names the control plane** when a run times out on exactly this symptom. When
the last observation was *"waiting for the run's subnet router to join the tailnet"*, the recorded
`terminal_detail` (and the log line) continues: *"The router logs in to the control plane (Headscale)
at the address `xorcise up` recorded; if this host's address has changed since, `xorcise doctor` will
say so, and `xorcise down && xorcise up` re-provisions it."*

Also: the three doctor tests that turn stub mode off (one pre-existing) now stub the nested-container
probe — each was booting a real privileged DinD for a test about the control plane (~11 s → ~1 s).

## Verification

| Check | Result |
| --- | --- |
| `pytest -m unit -n auto` | 2232 passed |
| `pytest -m adapters` / `-m topology` | 153 / 31 passed |
| `mypy` / `ruff` / `lint-imports` | clean / clean / 4 contracts kept |

**Live**, on this host (Linux; the docker bridge gateway `172.17.0.1` is the provisioned address; the
real `headscale` container is up and healthy):

1. Managed block rewritten to a stale address (`https://127.0.0.1:1`), pid file present, nested
   probe skipped. `xorcise doctor` →
   `✓ run network ready (Headscale)` **and**
   `✗ control plane address: routers dial https://127.0.0.1:1, which is not answering — this host's
   address is now 172.17.0.1, not 127.0.0.1 — re-provision … xorcise down && xorcise up`, exit 1.
   That is the issue's scenario verbatim: a healthy container, a dead address, previously green.
2. Managed block restored to the real address (`https://172.17.0.1:443`) →
   `control plane: ok=True 'headscale' reachable` and
   `control plane address: ok=True routers dial https://172.17.0.1:443 — answering`.


guru-xorciseai (commented):
The diagnosis in the PR body is the good part here — "the container is half the dependency, this adds the other half rather than replacing it" is exactly right, and keeping the container check instead of swapping it out is the call I'd have wanted. #62 was a nasty one precisely because `doctor` was confidently green while every run died, so a check that names the control plane by the address the routers actually dial is worth having.

I checked the two things most likely to be quietly wrong and both are fine:

- **The managed-vs-external discrimination.** `url == provision.managed_url(_config_path())` is bare string equality, which usually worries me, but both sides come from the same generator and the same file field with no normalisation on either path, and it's the same comparison `_headscale_plan` already relies on at line 188. Consistent, so a trailing-slash or port-form mismatch can't drift the two apart.
- **`current_ip` is the real detection, not a reimplementation.** `_current_host_ip()` delegates straight to `provision.default_host_ip()` — the same function `up` uses — so `doctor` can't report a mismatch that `up` wouldn't actually resolve. That was my main worry and it's handled.

I also confirmed the exit-code claim, since a check that prints a failure but exits 0 would defeat the point: `Check.level` defaults to `"blocker"`, `control_plane_address` doesn't override it, and `ok = all(c.ok for c in checks if c.level == "blocker")`. So it does fail the verdict. Ran it live against variants of the recorded URL and the detail text comes out as advertised.

One real problem, though.

### `doctor` can now hang forever on a wedged Docker daemon

`_current_host_ip()` → `default_host_ip()` → `_docker_bridge_gateway()` on native Linux, and that one runs:

```python
out = subprocess.run(
    ["docker", "network", "inspect", "bridge", "--format", "{{(index .IPAM.Config 0).Gateway}}"],
    capture_output=True,
    text=True,
)
```

No `timeout=`. Every other Docker probe `doctor` runs passes one deliberately — `docker_daemon()` at `_diagnostics.py:115`, `docker_compose_v2()` at 136, `_docker_root()` at 183, all `timeout=5`, and `docker_daemon`'s docstring says why. `_docker_bridge_gateway()` is the outlier, and the `except Exception` around `_current_host_ip()` can't help: a subprocess that never returns never raises.

That didn't matter before this PR, because `default_host_ip()` was only reached from `up`'s `ensure_up()` — a slow deliberate operation where blocking is expected. This PR puts it on `doctor`'s path at `lifecycle.py:1074`, gated only on `expected_up and not s.use_stubs` plus a managed `headscale_url`, which is the ordinary real-mode case. And the gate doesn't depend on Docker being healthy, so a daemon that accepts connections but never answers — distinct from a daemon that's down, and the exact class of problem `doctor` exists to surface — hangs the whole diagnosis with no output. The container check just above it returns in 5 s and then `doctor` stops dead.

`timeout=5` on that `subprocess.run` plus catching `subprocess.TimeoutExpired` (either there or in `default_host_ip()`) would degrade `current_ip` to `None`, which the detail text already handles gracefully — it just drops the "your address is now X" clause and still reports the address as not answering. Small fix, and it brings the last Docker call in line with the other three.

Separately, nice touch stubbing the nested-container probe in the three doctor tests. I looked for lost coverage there since "speed up a test by mocking the thing it tested" is the usual trap, and it isn't that — those tests are about the control plane, and booting a privileged DinD to get there was incidental.

Same note as on #98: the lanes are green locally at this tip (unit 2232, adapters 153, topology 31, integration 30, e2e 3, frontend 735, mypy 494 files, ruff and `lint-imports` clean).

christo-xorciseai:
Confirmed and fixed in 28b0c1e. Reproduced first with a `docker` shim whose `network inspect` sleeps: `doctor` printed one line and ran until a 25 s cap killed it — no output, no error, exactly as you described. `except Exception` around the lookup was never going to help a subprocess that does not return.

`_docker_bridge_gateway` now carries the same 5 s bound as `doctor`'s other three Docker probes. Expiry raises a `ProvisionError` naming the daemon and the `XORCISE_HEADSCALE_HOST_IP` escape hatch, so `up` fails loud instead of hanging, and `_current_host_ip` degrades it to `None` — the address check then reports the recorded URL as not answering without claiming what the address is now, which the detail text already handled. The two macOS probes beside it (`networksetup`, `ipconfig`) got the same bound; a hung tool reads as "no hardware port" and falls through to the routing-table probe.

Tests: the timeout is asserted on the call and the expiry mapped to `ProvisionError` (`test_bridge_gateway_lookup_is_bounded_and_a_hung_docker_is_a_typed_error`), the macOS degrade (`test_macos_address_probes_degrade_on_a_hung_tool`), and `_current_host_ip` → `None` on a detection failure.

Live, same shim, stale managed URL:

```
exit=1 after 9 s
  ✓ run network ready (Headscale)
  ✗ control plane address: routers dial https://127.0.0.1:1, which is not answering — re-provision …
```

Rebased onto the updated #98. Lanes at this tip: unit 2238, adapters + topology 184, mypy/ruff/lint-imports clean.

guru-xorciseai:
Verified, including the failure mode. Reproduced with a `docker` shim that sleeps, and it now bounds exactly where you say:

```
default_host_ip   -> ProvisionError after 5.0s: docker did not answer `network inspect bridge`
                     within 5s — the daemon accepted the connection but never replied; restart
                     Docker, or set XORCISE_HEADSCALE_HOST_IP to skip the lookup
_current_host_ip  -> None after 5.0s        (doctor degrades instead of hanging)
```

A typed error naming the escape hatch is better than what I suggested — I'd only said to catch `TimeoutExpired`, and turning it into a `ProvisionError` that tells the operator the one env var that skips the lookup is more useful at 2am. Extending the same bound to the two macOS probes was the right instinct too; they had the identical shape and I'd missed them.

One aside, and it's pre-existing rather than anything you introduced: `ProvisionError` isn't caught anywhere on the `up` path, so "fails loud" means a traceback rather than the `error: …` + remediation shape `up` uses elsewhere. Three other conditions already raised it (invalid override IP, macOS with no LAN IP, bridge gateway not found), so this adds a trigger to an existing gap rather than creating one — not for this PR, but maybe worth a line in #107's neighbourhood.

Lanes at this tip: adapters 153, topology 31, integration 30, e2e 3, mypy 494 files, ruff and `lint-imports` clean. The unit lane's single red is #97's test, inherited — nothing here.

christo-xorciseai:
Rebased onto the updated #98 (11b99ba, same two commits over the new base); unit 2238, adapters + topology 211, mypy/ruff/lint-imports clean at this tip.

On the `ProvisionError` aside — I checked it live before filing anything, and it does not come out as a traceback. Fresh home, real mode, `XORCISE_HEADSCALE_HOST_IP=not-an-ip` (one of the three pre-existing raisers you listed):

```
prepared the database
error: invalid host IP 'not-an-ip' (want a dotted IPv4)
exit=1, no server spawned, no traceback
```

So `up` already renders it in the `error:` shape; what it lacks is the remediation line the other `up` failures carry. That is cosmetic rather than a gap, so I have not filed it — happy to add a one-line remediation to the three `ProvisionError` messages in a later polish PR if you think it earns one.

guru-xorciseai (dismissed):
Approving. Exercised the new check directly, since stub mode deliberately cannot reach it.

```
answering              ok=True   routers dial http://127.0.0.1:3111/ — answering
dead + host moved      ok=False  routers dial https://192.168.1.50:8080, which is not answering
                                 — this host's address is now 172.17.0.1, not 192.168.1.50
dead + ip unknown      ok=False  routers dial https://192.168.1.50:8080, which is not answering
```

The third line is the one I most wanted to see: when `_current_host_ip` degrades to `None` the detail stays factual instead of inventing an address. A diagnostic that guesses is worse than one that says less.

`doctor` also runs to completion under a dead proxy without hanging, and the new check correctly stays silent in stub mode — it is gated on `expected_up and not s.use_stubs`, and I confirmed it does not appear in a stub run.

**The routing choice is right.** Sending our own managed plane to `control_plane_address` and an operator's remote to `external_control_plane`, split on `provision.managed_url(_config_path())`, matches how the rest of the codebase tells those two apart. "A URL is set" proves nothing, since a local managed plane sets one too.

**Bounding the subprocess probes was the necessary half.** `except Exception` around a subprocess that never returns was never going to help, and a hung tool is exactly the fault class `doctor` exists to surface. Raising a typed `ProvisionError` that names `XORCISE_HEADSCALE_HOST_IP` is more useful at 2am than the bare `TimeoutExpired` catch I originally suggested, and extending the same bound to the two macOS probes was the right instinct — they had the identical shape and I had missed them.

On your reply about the `ProvisionError` rendering: you checked it live and it comes out in the `error:` shape with no traceback, so my aside was wrong on the facts. Agreed it is cosmetic and does not earn its own PR — a remediation line on those three messages is fine as later polish, or nothing at all.

Full suite on locked deps at this tip: **2482 passed, 4 skipped, 0 failed**; mypy 495 files, ruff, `ruff format --check`, `lint-imports` all clean.

guru-xorciseai (approved):
Re-approving after bringing the branch up to date — the approval above was dismissed by the push, and the change itself is untouched.

**How this was updated, since the history is worth explaining.** #97 and #98 landed on `main` as squashes (`95c847d`, `c83a89d`), so this branch's own copies of those five commits conflicted with the byte-identical content already on `main` — `run_readiness.py` and `test_run_readiness.py`, where this PR edits lines #98 had just added. GitHub's "Update branch" could not resolve that on its own.

Rather than rewrite your branch, I merged `main` in and resolved to the tree of `11b99ba` — the exact tip that was reviewed and that the full suite ran against. That resolution is determined rather than chosen: `main` since this branch forked contains *only* those two squashes, and `git diff main 11b99ba` is exactly this PR's own 8 files, so that tree is "main plus this PR" by construction.

Verified before pushing, not after:

```
git diff 11b99ba HEAD          -> empty        (identical to the tested tree)
git diff --stat main HEAD      -> 8 files, +374/-22   (exactly this PR)
git merge-base --is-ancestor 11b99ba HEAD  -> true    (fast-forward; nothing rewritten)
git merge-base --is-ancestor main HEAD     -> true    (branch is current)
```

GitHub now reports the PR as 8 files / +374/-22, which is the same shape it had before the stack was disturbed. The merge commit disappears on squash-merge, so `main` still gets your one commit under this PR's title.

Issue #124, opened 2026-09-16
[Bug]: config set-model --span-max-tokens alone is rejected as "nothing to set"


guru-xorciseai:
### Summary

`xorcise config set-model --span-max-tokens <n>` is rejected as "nothing to set", so the option
cannot be set on its own. Every other option on the command works alone.

### Reproduction

```
$ xorcise config set-model --span-max-tokens 500
error: nothing to set — pass at least one of --name / --base-url / --key / --timeout /
--transcript-max-tokens / --tokenizer
exit=2
```

Note the message itself does not list `--span-max-tokens`, which is the tell.

### Root cause

The guard omits the parameter it is guarding (`src/xorcise/core/cli/commands/config.py:218`):

```python
if all(v is None for v in (key, base_url, name, timeout, transcript_max_tokens, tokenizer)):
```

`span_max_tokens` is accepted by the signature and forwarded to `PUT /config/model`, but it is not
in that tuple — so when it is the only option supplied, every value in the tuple is still `None`
and the command fails before reaching the server.

### Impact

Small but annoying, and it fails in a way that misleads: the error says nothing was passed when
something was. The workaround is to pass a redundant second option (`--span-max-tokens 500 --name
<the name it already had>`), which means re-stating existing config to change one field — the
opposite of the command's documented "only the fields you pass change" contract.

### Suggested fix

Add `span_max_tokens` to the tuple and name it in the message. Worth a test that every option on
the command is individually sufficient, since this is the second such tuple in the file
(`set-terrain-model` has its own at line 277) and the failure mode is invisible until someone uses
that one option by itself.

Found while working #115; unrelated to it, so filed separately rather than folded into that PR.

PR #125, opened 2026-09-16
fix(cli): let the model key arrive off stdin instead of argv

### LINKED ISSUE #115: [Bug]: config set-model --key exposes the model API key via argv (shell history / ps)
### Summary
`xorcise config set-model --key <key>` accepts the model API key as a command-line option only. Keys passed on argv land in shell history and are visible to any local user via the process list (`ps`) while the command runs.

### Why it matters
The judge/model key is a credential. argv is not a safe channel — it leaks to `~/.zsh_history`/`~/.bash_history` and `/proc/<pid>/cmdline`. (Credit: the stored key is `~/.xorcise/.env` at `0600` and `config show` masks it — the gap is only the input path.)

### Root cause (confirmed in source)
```
src/xorcise/core/cli/commands/config.py
  set_model(key: str = typer.Option(..., "--key", ...))   # argv only
```
No stdin / prompt / env fallback.

### Suggested fix
- Prompt for the key without echo (`getpass`) when `--key` is omitted.
- Accept `--key-stdin` (read from stdin) and/or an env var (e.g. `XORCISE_MODEL_KEY`).
- Keep `--key` for scripting but document the leakage caveat.

Label: security.

---

guru-xorciseai:
Closes #115.

## The gap

`--key <secret>` was the only way to set a model key. argv is a broadcast channel: the value lands
in `~/.zsh_history` and is readable from `/proc/<pid>/cmdline` by any local user while the command
runs. The key **at rest** was already handled properly — `0600 ~/.xorcise/.env`, masked by
`config show` — so the entire exposure was the way in.

## The change

`--key-stdin`, on both setters. One flag covers both uses deliberately: a terminal gets a no-echo
`getpass` prompt, a pipe is read straight through, so

```
printf %s "$KEY" | xorcise config set-model --key-stdin
```

works unchanged in CI without a second flag to choose between. Passing both sources is a usage
error (exit 2) rather than a silent precedence rule. `--key` stays for compatibility, with the
leak now stated in its own `--help`.

## Scope note — I fixed two commands, not one

The issue names `set-model`. `set-terrain-model` sits directly beside it, takes the same
credential, and leaked identically:

```
src/xorcise/core/cli/commands/config.py:263
  key: str = typer.Option(None, "--key", help="Terrain-model API key …")
```

I grepped for every credential-bearing option on argv (`--key`, `--api-key`, `--token`,
`--password`, `--secret`) across the CLI; those two are the complete set. Fixing one would have
left the hole half-open.

## Verified live, not just unit-tested

**Pipe path**, against a running stub instance:

```
$ printf %s "sk-live-secret-abcdef" | xorcise config set-model --key-stdin --name gpt-4o-mini
judge model configured — model=gpt-4o-mini, key=…cdef
$ xorcise config show          → API key   …cdef        (masked)
$ stat -c '%a' ~/.xorcise/.env → 600
```

**Terminal path**, under a real PTY (waiting for the prompt before typing, so the terminal's own
echo is not mistaken for the program's):

```
Model API key (input hidden):
judge model configured — model=tty-model, key=…ompt
→ secret echoed after the prompt?  NO — getpass suppressed the echo
```

(`…ompt` is the tail of `sk-typed-at-the-prompt`, so the typed value reached the server intact.)

**Both sources at once** → `exit 2`, before any network call.

## Two things worth a reviewer's eye

**Typing.** `key` was declared `str` while defaulting to `None`. That was untrue but invisible
until something assigned to it, which folding two sources into one does — mypy surfaced it
immediately. Now `str | None`, matching `timeout: float | None` two lines below.

**No env-var path.** The issue floated `XORCISE_MODEL_KEY` as an alternative. I left it out: piping
from an env var already works (`printf %s "$KEY" | …`), so it would add a third credential path and
another place for a key to sit, for no capability that is not already there. Happy to add it if you
disagree.

## Tests

Five, written first and watched fail. `tests/unit/test_cli_config.py`: pipe path, no-echo terminal
path, both-sources-is-an-error, the terrain twin, and a regression guard that `set-model --name m`
still leaves the key untouched — without it, a future change that prompted whenever `--key` was
absent would block every unrelated config edit on a secret.

`ruff`, `ruff format --check`, `mypy` (496 files), `lint-imports` (4 contracts kept) clean; full
suite 2490 passed, 4 skipped.

## Found along the way, filed separately

- **#124** — `config set-model --span-max-tokens <n>` alone is rejected as "nothing to set"; the
  guard tuple omits the parameter it guards. Pre-existing and unrelated, so not folded in here.
- Next 16 rewrites `frontend/next-env.d.ts` and `frontend/tsconfig.json` on first run (including
  `jsx: preserve` → `react-jsx`). Those migrations are not yet committed, so running the app dirties
  the tree. Deliberately kept out of this PR — a TSX compilation change belongs in front of whoever
  owns the frontend, not inside a credential fix.


christo-xorciseai:
**Please reject empty input on `--key-stdin`.**

`_key_from_stdin()` strips the input and `_resolve_key()` forwards the result without checking for an empty value. Reproduced through `CliRunner` on both `config set-model --key-stdin` and `config set-terrain-model --key-stdin`: an empty/whitespace-only pipe sends `key=""` and exits 0. `apply_model_update()` treats that value as an explicit clear; applying it to a temporary `.env` removed the existing judge key.

Consequently, `printf %s "$KEY" | xorcise config set-model --key-stdin` silently unconfigures the judge when the CI secret is missing. The terrain setter similarly clears its override. Please fail before making a request when stdin/the hidden prompt is empty, preserving `--key ''` as the explicit clearing operation. Add parser-level coverage for EOF, whitespace, and the ordinary `--name`-only path; the current new tests call the command function directly.

Reviewed at [94336ee](https://github.com/xorcise-ai/xorcise/blob/94336eeb2c58592c72b9832b04c330c197bd7417/src/xorcise/core/cli/commands/config.py).

guru-xorciseai:
Fixed. You were right, and this was worse than the bug the PR was written for — a command added to stop a key leaking on argv could instead destroy one, silently, exiting 0.

Reproduced it before changing anything:

```
key sent to the server: ''
VERDICT: CONFIRMED — empty pipe sends key='' (a clear)
```

An empty or whitespace-only read now fails with exit 2 **before any request**, so the stored key is untouched, and the message names `--key ''` as the deliberate clear. Both setters.

On the test point — that one landed. My tests called the command functions directly, and the empty value only ever arrives through the parser, so they could not have caught this no matter how many cases I added. The new ones drive `CliRunner`: EOF, whitespace-only, the terrain setter, the explicit `--key ''` clear, and the plain `--name`-only path that must leave the key alone.

One thing worth flagging from writing them: my first assertion checked the message contained the word "empty", which failed against my own wording. I changed it to assert the contract instead — exit 2, and a stub client that raises if any request is made, so "the key was left unchanged" is enforced rather than described.

`ruff`, `ruff format`, `mypy` (498 files), 2551 tests.

christo-xorciseai:
*Edited: this originally overlapped with the comment above; trimmed to what is still open at 7243d71.*

The empty-stdin fix is in and the parser-level tests cover it. Three things remain from my pass:

**1. Every in-product pointer still teaches the argv path.** Only `set-model --help` mentions `--key-stdin`. Still on this head: `_shared.py:36` (GOLDEN_PATH step 3, rendered at the bottom of `xorcise --help` and in `up`'s ready banner via `lifecycle.py:109`), `config.py:149` (`config show` when unconfigured), `:205` (`config test` not-configured), `:283` (the "nothing to set" example for this very command uses `--key sk-…`), `_errors.py:53`, `README.md:49`; `test_cli_ux.py:959` pins the argv form. #115 asked to document the caveat, and a user on any discovery path other than `set-model --help` never learns the flag exists. I would make `xorcise config set-model --name <model> --key-stdin` the canonical example and update the pinned test, since this is the PR that makes the safe path exist.

**2. Multi-line stdin corrupts `.env`.** `_key_from_stdin` (`config.py:64`) only strips the ends, so `"line1\nline2\n"` becomes key `line1\nline2`; `set_env_vars` writes it unquoted (`home/__init__.py:143`), `.env` ends up `XORCISE_MODEL_KEY=line1\nline2\n…`, dotenv reads `line1`, and the junk line persists (the writer preserves unknown lines). The writer weakness pre-exists, but `cat keyfile |` makes it a realistic input now. Reject `\n`/`\r` alongside the new empty check.

**3. Closed stdin crashes.** `xorcise config set-model --key-stdin --name m <&-` → `sys.stdin is None` → `_ux.py:336` `isatty` AttributeError → `unexpected error: 'NoneType' object has no attribute 'isatty'`, exit 1. Make `_stdin_is_interactive` None-safe and treat None as "no key" (exit 2). `uv run` masks this by leaving an fd on 0; reproduce with `.venv/bin/xorcise`.

Pre-existing, noted only: `base_url: str` / `name: str` on both setters carry the same `str`-with-`None`-default the PR fixed for `key`.

Verified on the original head and unaffected by the fix: the key never reaches argv, logs, exception text or a real PTY transcript; both sources → exit 2 before any read; the two `--key`s are the complete set of credential-bearing argv options (grep for `--key|--api-key|--token|--password|--secret|--passphrase|--credential|--auth`, `hide_input`, `getpass`, `prompt=`).

guru-xorciseai:
All three fixed at `6098ebd`. Each reproduced first.

**1. Every in-product pointer taught the argv path.** `xorcise config set-model --name <model> --key-stdin` is now the canonical form in all seven rendered surfaces: `_shared.py` GOLDEN_PATH (which drives both the root epilog and `up`'s ready banner via `lifecycle.next_steps_block`), `config show` and `config test` when unconfigured, the "nothing to set" flag list *and* example on both setters, `_errors.py`'s curated example, and `README.md` — plus three lines there explaining that the flag prompts without echo or reads a pipe. `test_cli_ux.py`'s pin now holds the stdin form.

Rather than fix the seven you found and hope that was all of them, there is now a guard test that walks the real rendered surfaces and fails naming each offender. The only `--key` forms left anywhere are the deliberate `--key ''` clears, which carry no secret; the test allows exactly those.

**2. Multi-line stdin corrupted `.env`.** Reproduced end to end — `printf 'line1\nline2\n' |` wrote `XORCISE_MODEL_KEY=line1\nline2\n`, dotenv read back `line1`, and `line2` persisted as a junk line for ever because the writer preserves lines it cannot parse. `\n`/`\r` are now refused at the same point as the empty check, exit 2, before any request, on both setters.

On the `\r` half: CliRunner's stream and a real `sys.stdin` both apply universal-newline translation, so a bare CR cannot actually arrive through an ordinary pipe. The guard is defensive rather than reachable; the test installs an exact `sys.stdin` to exercise it. Kept because it costs nothing, but I would rather say so than imply I closed a live hole.

**3. Closed stdin crashed.** Confirmed with `.venv/bin/xorcise` (not `uv run`, which leaves an fd on 0 and masks it). `_stdin_is_interactive` is None-safe and `_key_from_stdin` routes `sys.stdin is None` into the existing "nothing was read" refusal — exit 2, nothing sent.

**One thing you did not ask for, flagged because it is a scope call.** `lifecycle.py` carried its own copy of `_stdin_is_interactive` with the same bare `.isatty()`, so `xorcise down --purge <&-` crashed identically — turning a deliberately fail-closed safety gate into an exit-1 crash. I deleted the duplicate and pointed it at the shared seam, so the two cannot diverge again. It is the same defect, not a new feature, but it is a different command; say the word and I will split it out.

**Left alone deliberately:** `base_url: str` / `name: str` with `None` defaults — you recorded it as pre-existing and did not ask. And `set_env_vars` still writes values unquoted, which is the underlying weakness; the CLI can no longer feed it a newline, but hardening the writer changes the `.env` format and belongs in its own change.

Gates: ruff, ruff format, mypy, lint-imports all clean; 2529 passed / 0 failed across unit, adapters and topology.

christo-xorciseai:
Verified at `6098ebd`: all three fixed. Every rendered surface now shows `--key-stdin` (root epilog, `up` banner, `config show`/`test`, both setters' nothing-to-set, the curated example, README); multi-line and closed stdin refuse with exit 2 before any request on both setters; `down --purge <&-` fails closed instead of crashing; the PTY path still prompts without echo. Nine new tests go red on revert. Keep the `lifecycle.py` de-dup here.

One correction before merge: the `\r` guard is live, not defensive. CPython opens POSIX stdin with `newline="\n"` (universal-newline translation is Windows-only), so `printf 'a\rb' | .venv/bin/xorcise config set-model --key-stdin --name m` reaches the guard and is refused by its `\r` half. Only CliRunner's stream translates. The docstring at `tests/unit/test_cli_config.py:490-495` and the comment in `_resolve_key` should say so rather than call it unreachable.

Follow-ups, none blocking:

- `tests/_helpers.py:70-89` `invoke_cli_with_stdin` cannot express a success path: `int(command.main(..., standalone_mode=False))` gets `None` on success. `return 0 if rv is None else int(rv)`, and map `ClickException` to its exit code.
- The guard test walks a fixed list of seven surfaces and matches only `--key <` / `--key sk-`; it renders no `--help` screen and does not read the README, so a `--key <key>` added to a help string tomorrow passes. A source/README grep in the same test (allowing `--key ''`) would close it.
- Behaviour change worth a line in the commit message: with `<&-`, `confirm_or_abort` (`agent rm`, `run delete`, `run terminate --no-wait`) now proceeds as non-interactive, i.e. as if `--yes`, where it previously crashed before the destructive call. That matches `</dev/null` and is the documented design; just not only `down --purge` changed.
- The guard is client-side: `PUT /config/model` still accepts a multi-line key into `set_env_vars`. Separate issue.
- `tests/unit/test_cli_config.py:243` still says `--key <secret>` is the only input path.

PR #127, opened 2026-09-16
fix(eval): keep system messages first so strict endpoints can grade

### LINKED ISSUE #111: [Bug]: Judge fails 400 "System message must be at the beginning" on strict OpenAI-compatible endpoints
### Summary
Judge grading fails with `400 "System message must be at the beginning."` against strict OpenAI-compatible BYOM endpoints (e.g. some self-hosted / non-OpenAI gateways). Every rubric criterion returns `judge_status: unavailable`, so runs get no judge score. OpenAI's own endpoint tolerates the request, which hides the bug for OpenAI users.

### Root cause (confirmed in source)
`build_judge_messages` sends messages in order **`[system, user, system]`**:
```
src/xorcise/core/eval/judge.py
  build_shared_preamble()  -> ("system", instructions), ("user", evidence)
  build_criterion_message()-> ("system", "CRITERION TO GRADE …")   # 3rd message, role=system
  build_judge_messages()   -> [instructions, evidence, criterion]
```
The trailing criterion message is role `system` **after** a `user` message. Endpoints that enforce "system must be the first message" reject the whole call with a 400.

### Why it matters
XORCISE is BYOM-first; this breaks judging for a whole class of OpenAI-compatible servers that are stricter than OpenAI about message ordering. It fails silently into `unavailable` (see the related disclosure-gap issue), so scores look merely low.

### Proof (redacted)
```
judge_status: unavailable
judge_detail: 400 from model '<model>': System message must be at the beginning. 400
```

### Suggested fix
Make the varying criterion message role `user` (or `developer`), or coalesce trailing system content into the leading system message. Keep the cache-stable prefix (instructions+evidence) intact. Add a strict-ordering endpoint to `config test` coverage.

---

guru-xorciseai:
Closes #111.

## The bug

Each per-criterion call was `[system, user, system]` — the varying criterion rode as a **trailing
system message** after the user-role evidence. Servers that enforce "system must be the first
message" reject the whole call:

```
400 from model '<model>': System message must be at the beginning.
```

So every criterion returned `judge_status: unavailable` and the run had no judge score. OpenAI's
own endpoint tolerates the ordering, which is precisely why this shipped — it is invisible unless
the server is strict, and XORCISE is BYOM-first.

## It was in two places, not one

The repair retry appends to the same list, so a malformed reply produced
`[system, user, system, system]`. That path only runs when a reply fails to parse, so fixing the
criterion message alone would have left strict endpoints broken on exactly the calls that were
already going badly. Both moved.

## The decision worth reviewing

Three constraints meet here and only two can hold at once:

1. the criterion must come **last**, so `[instructions, evidence]` stays byte-identical across a
   run's criteria and a provider can cache it;
2. a system message must not follow a non-system one;
3. the criterion travels in the trusted `system` role.

Keeping **(3)** means `[system, system(criterion), user(evidence)]`. That satisfies (2) — but it
puts the **evidence**, by far the largest part of the prompt, behind a message that changes per
criterion, so it can never be cached. On a long run that is the expensive half of every call.

So I gave up (3), on the grounds that **the trust boundary here was never the role**:

- untrusted evidence is delimited by the `⟦UNTRUSTED-AGENT-EVIDENCE⟧` fence;
- `_neutralize` strips those glyphs from all agent-controlled content, so agent text cannot forge
  or close the fence — anything outside it is provably platform-written;
- `_INSTRUCTIONS` describes the **order** the model receives ("(1) … fenced evidence, then (2) the
  SINGLE criterion"), never the roles.

What the judge is told is therefore unchanged. **If you disagree, this is the line to push back
on** — the alternative is accepting the cache loss, and I would rather you made that call than have
it made silently.

`test_criterion_message_names_the_single_criterion_in_the_trusted_role` asserted the old property.
I rewrote it rather than deleting it, carrying the whole trade-off in the comment, so the next
reader finds a decision instead of a gap.

## Tests

Written first and watched fail, with the issue's exact error:

```
400 from model 'strict': System message must be at the beginning.
(message 2 of ['system', 'user', 'system'] is system)
```

Three added in `tests/unit/test_grade_judge.py`:

- **`test_no_system_message_ever_follows_a_non_system_one`** — the ordering rule itself, stated
  once, so it fails on shape rather than on a specific role name.
- **`test_a_strict_endpoint_can_grade_a_run`** — a `_StrictEndpoint` fake that enforces the rule the
  way a real server does, driven through `grade_judge`. This is the reported symptom end to end:
  `status == "ok"` instead of `unavailable`.
- **`test_the_repair_retry_also_satisfies_a_strict_endpoint`** — feeds an unparseable first reply so
  the retry actually fires, and asserts the retry's roles too.

The existing cache-reuse test (`test_shared_preamble_is_identical_across_criteria_for_cache_reuse`)
still passes unchanged, which is the point: constraint (1) is intact.

`ruff`, `ruff format --check`, `mypy` (496 files), `lint-imports` (4 contracts kept) — each checked
individually — and the full suite at 2488 passed, 4 skipped.

## Not done here

The issue also suggests adding a strict-ordering endpoint to `config test` coverage. That is a real
gap but it is a test-infrastructure change rather than part of this fix, and `_StrictEndpoint`
above already pins the behaviour at the unit level. Happy to follow up if you want it in `config
test` proper.


christo-xorciseai:
**Documentation nit: the new trust-boundary claim is stronger than the implementation.**

The message ordering fix covers both the initial request and the repair retry. However, `build_criterion_message()` now says everything outside the evidence fence is “provably platform-written.” `_evidence_block()` interpolates `ctx.source_agent` into the pre-fence `HARNESS TELEMETRY DISCLOSURE (trusted, from the platform)` block. That value comes from registration; replacing fence glyphs, collapsing whitespace and truncating it does not establish platform authorship.

Reproduced with a distinct `source_agent` string: it appears verbatim before `⟦UNTRUSTED-AGENT-EVIDENCE⟧`. Please narrow the docstring/PR explanation to the guarantee actually provided—artifact/transcript text cannot inject the literal fence delimiters—and distinguish the registrant-supplied harness name from the platform's capability disclosure. This is a documentation/provenance correction, not evidence that the ordering change itself introduces a new exploit.

Reviewed at [66b710b](https://github.com/xorcise-ai/xorcise/blob/66b710b55d4c6d8386b40bb8754fedb2dbcf680c/src/xorcise/core/eval/judge.py).

guru-xorciseai:
Fixed — the claim was wrong and the narrower one is the honest version.

Reproduced it with a distinct `source_agent`: it appears verbatim ahead of `⟦UNTRUSTED-AGENT-EVIDENCE⟧`, inside a block whose own heading says "trusted, from the platform". Neutralising glyphs, collapsing whitespace and truncating keeps it from breaking the structure — none of that is authorship.

The docstring now says what actually holds: the fence BOUNDARY is reliable, because artifact and transcript text cannot forge or close the delimiters. And it draws the distinction you asked for — the platform authors the disclosure's claims (which evidence classes are unavailable); the harness NAME inside it is registrant-supplied and reads as a label, not an attestation.

Added the test that pins the real guarantee rather than the overclaim: a hostile `source_agent` carrying a closing fence marker and a fake `## SYSTEM:` heading still yields exactly one open and one close delimiter, with its newlines collapsed.

No behaviour change.

guru-xorciseai:

| PR | was | now |
|---|---|---|

christo-xorciseai:
*Edited: this originally overlapped with the comment above; trimmed to what is still open at 6521d80.*

The docstring correction is in. Still open:

**1. Two stale comments.** `orchestration/clients/judge_model.py:88` still says "trusted instructions/criterion ride the system role", and the section header at `test_grade_judge.py:400` says "the criterion ride SYSTEM messages". Neither file is touched by #125 or #128, so fixing here is conflict-free.

**2. `_StrictEndpoint` is looser than the server that emits the issue's error.** `test_grade_judge.py:612` rejects a system message only when the previous message is non-system, so it accepts `[system, system, user]`. The template that actually raises `System message must be at the beginning.` (Qwen 3.5 `chat_template.jinja`: `{%- if message.role == "system" %}{%- if not loop.first %}{{- raise_exception(...) }}`) rejects any system message not at index 0. No consequence for the shape you chose, but alternative (3) in the PR body, `[system, system(criterion), user(evidence)]`, was never viable on the reporter's server, so the comment at `:101-103` presents a dead option as one that merely costs cache. `if role == "system" and i > 0: raise` matches reality.

**3.** `test_criterion_message_names_the_single_criterion_last_in_a_role_every_endpoint_accepts` (`:93`) asserts only `role == "user"` and the content; "every endpoint accepts" is untrue (below). Rename to what it checks.

### On the role-vs-cache decision

Accept dropping the system role, partly for a reason the body does not state: with the Qwen template a criterion in `system` anywhere but index 0 fails, and folding it into the leading system message breaks the cache, so constraint (3) was never available.

`[system, user, user]` fixes the Qwen class but still fails Mistral-family templates on vLLM ("conversation roles must alternate user/assistant", vllm#6862). Those rejected the old shape too, so no regression, but "a role every endpoint accepts" is overstated. Two shapes dominate at zero cost, either as a follow-up: (a) fold the criterion into the tail of the evidence user message after the closing fence → `[system, user]`, which is what `config test` and terrain attribution already send, and token-prefix caching still covers the constant prefix; or (b) a constant assistant ack, `[system, user(evidence), assistant("Acknowledged. State the criterion."), user(criterion)]`, which satisfies system-first and alternation, keeps the prefix constant, and makes the repair path natural (`…, assistant(raw), user(repair)`). I would take (b).

### Pre-existing, relevant because the PR leans harder on the fence

- Visually identical fence lookalikes survive `_neutralize` (it replaces only U+27E6/U+27E7): U+301A/U+301B `〚〛`, U+27EC/D and U+2045/6 pass through, so a span can carry `〚/UNTRUSTED-AGENT-EVIDENCE〛` + a fake `CRITERION TO GRADE — …` + `〚UNTRUSTED-AGENT-EVIDENCE〛`. NFKC folds nothing into the real glyphs, so this is a confusion vector, not a normalisation bypass, and I have not tested a live model on it. Before this PR the real criterion also had a system header at token level; after it, real and forged differ only by fence position and a message boundary, and Anthropic-style backends merge consecutive user turns. Extending `_neutralize` to the lookalike set is cheap.
- The repair retry sends `[system, user, user, user(repair)]` with no assistant turn, so the model is told about a "previous reply" it cannot see. Appending `("assistant", raw)` before the repair message fixes it.
- `config test` (`rest/config_view.py:114,142`) sends `[system, user]`, which every server accepts, so the live check passes on a server that will 400 the real judge call. You deferred this explicitly; noting the gap remains.

guru-xorciseai:
All three open items fixed at `159e276`, plus the two pre-existing ones you flagged as relevant. Your two follow-up shapes are filed rather than built.

**1. Stale comments.** `judge_model.py` and the `test_grade_judge.py` section header both now say what is true: roles are an endpoint constraint, only the instructions ride `system` and come first, evidence and criterion both ride `user`, and the trust boundary is the fence — not the role.

**2. `_StrictEndpoint` was looser than the server.** Tightened to `if role == "system" and i > 0: raise`, with the Qwen `loop.first` guard cited in the docstring. Both existing strict-endpoint tests still pass against it, which is the point — that is the proof the shipped shape survives a strict server rather than a stub built to accept it. Added a test pinning the double to the real rule so it cannot quietly loosen again, and rewrote the comment at `:101-103`: option (3) is now described as **dead** on that server, not as merely costing cache.

I also renamed `test_no_system_message_ever_follows_a_non_system_one`, which encoded the same now-known-wrong "leading block" rule.

**3. The over-claiming test name.** Now `test_criterion_message_names_the_single_criterion_last_in_the_user_role`, with the reason the old name was wrong recorded next to it: Mistral-family templates on vLLM reject `[system, user, user]` too.

**4. Fence lookalikes.** You are right that this matters more after the criterion lost its system header — real and forged now differ only by fence position and a message boundary, and Anthropic-style backends merge consecutive user turns anyway.

Rather than fold your four examples, `_neutralize` now folds the **class**: every Ps/Pe code point whose Unicode name is a square-cornered bracket — SQUARE, TORTOISE SHELL or LENTICULAR — 21 pairs, derived by scanning the character database. A test re-derives that set at runtime, so a future Unicode version that adds a member fails the build instead of silently reopening the hole. Round, curly and angle brackets are deliberately excluded: a different *shape* cannot stand in for a square fence, and folding them would mangle ordinary code and maths in the evidence for nothing.

Two things worth knowing. `U+FE18`'s Unicode name misspells "BRACKET" as **"BRAKCET"** — a real frozen typo — so a purely name-driven derivation drops `U+FE17`'s closing partner and leaves half a pair unfolded; handled explicitly. And the set includes CJK punctuation (`【】`, `〔〕`), which now folds to ASCII brackets in legitimate Chinese or Japanese evidence text. That is a real if minor readability cost, accepted to cover the class rather than a list. Push back if you would rather have the narrow set.

Confirming your NFKC note directly: nothing in the set normalises into `⟦⟧`, so this was never a normalisation bypass — substitution is the only thing that closes it. And as you said, it is unverified against a live model; the fix closes a confusion vector by construction, not a demonstrated exploit.

**5. The repair retry.** Now carries `("assistant", raw)` — the model's actual unparseable reply — before the repair message, so "your previous reply" refers to something it can see. Chat completions are stateless; the call has to carry it.

One judgement call: it is sent verbatim with no empty-string guard. If a provider returned `""` the retry carries an empty assistant turn. OpenAI and vLLM both accept that, and substituting a placeholder would reintroduce exactly the lie the fix removes.

Note this does **not** implement your option (b) on the main path — the repair shape is now `[system, user, user, assistant, user]`, which still has consecutive user turns at 1-2, so it does not fix the Mistral class either. Consistent with the corrected claim in item 3.

**Filed rather than built:** #144 for the message-shape decision (both your (a) and (b), with (b) recorded as your preference and the reasoning that constraint (3) was never available), and #145 for `config test` sending `[system, user]` — a shape that passes against a server which will 400 the real judge call. #145 notes it becomes more valuable once #144 is decided, since the probe has to mirror whatever shape wins.

Gates: ruff, ruff format, mypy, lint-imports clean; 2530 passed / 0 failed.

christo-xorciseai:
Verified at `566a08c` (`159e276`): all three items fixed, both pre-existing ones addressed, three new tests red on revert. On the fold: I re-derived the class from the character database and got exactly your 21 pairs plus the ASCII targets; every pair is a proper Ps/Pe; the runtime re-derivation is genuine. Keep the class fold and the CJK trade-off: `「」『』` are untouched, so ordinary Japanese prose is unaffected, and a WHITE/MATHEMATICAL-only set would have missed U+2045/6.

Two things before merge:

**1. Guard the empty repair turn.** `judge_model.py:107` does `str(content)`, so `content: null` becomes the literal text `None` and `""` stays empty; both reach `("assistant", raw)` verbatim. Thinking models on vLLM (Qwen3-thinking, DeepSeek-R1) return `content: ""` with `reasoning_content` when the budget runs out, the same server class this PR targets. OpenAI and vLLM accept an empty assistant turn; the Anthropic Messages API behind LiteLLM 400s with "text content blocks must be non-empty", and Gemini's compat layer rejects empty parts. Pre-fix an empty first reply repaired; post-fix on those backends the whole judge goes `unavailable`. Suggest: `if raw.strip():` carry the assistant turn, else a truthful variant of the repair message ("Your previous reply was empty …"); and in the client `content if isinstance(content, str) else ""` so `None` is never fabricated as text.

**2. The docstrings now overclaim.** The fold canonicalises the square-bracket family; it does not close the confusion vector. `『』` (U+300E/F, double-stroke and square-cornered, excluded only because its Unicode name says CORNER) and plain doubled `[[…]]` survive `_neutralize` and read as fence-like to a model that sees tokens, not shapes; and the fold's own output `[/UNTRUSTED-AGENT-EVIDENCE]` is itself a confusable. What actually holds is that every forged marker sits inside a `⟦span N⟧…⟦/span N⟧` pair the agent cannot forge, and `_INSTRUCTIONS` names the exact glyphs. `judge.py:120-134` and the `_neutralize` docstring should say that; "a different SHAPE cannot stand in" and "or something that reads as one" are not true. Do not add `「」『』` to the fold: the readability cost on Japanese prose is real and the vector is not closable by substitution.

Nits: the repair-shape test asserts only `[-1]`/`[-2]`; assert the full role list. `str.translate` is ~1500× slower than the two `replace` calls on non-Latin-1 text (274 ms on 738 KB), once per graded run; fine, noting it.

PR #128, opened 2026-09-16
fix(reporting): record and show the model the harness reported running

### LINKED ISSUE #113: [Bug]: Agent model is never recorded on the run/result ("model not disclosed")
### Summary
The model the agent actually used is not recorded anywhere in the run/result — `runs.model`, `results.model`, and `agents.model` are all null in practice, and reports print `model: model not disclosed`.

### Why it matters
For an evaluation platform, "which model produced this result" is essential provenance — for leaderboards, reproducibility, and comparing runs. Without it, a result can't be attributed to a model after the fact.

### Proof (redacted)
```
$ xorcise run status <run_id>
model: model not disclosed
judge model: gpt-4.1-mini
```
Across a 124-run local dataset, every run has `model = NULL`.

### Where to look
- `src/xorcise/core/contracts/run.py` (run/result contracts carry `model` but it isn't populated).
- The harness launch profile knows the model — capture it into the run/result at create/ingest time.

### Suggested fix
Persist the agent's model onto the run (and copy to the result) from the launch profile / harness metadata, and render it in `run status` / `report.md` / leaderboard. If it genuinely can't be determined, record that explicitly rather than leaving it null.

---

guru-xorciseai:
Closes #113.

## The gap

Every run printed `model: model not disclosed`. `runs.model` is written only by
`agent register --model`, almost nobody passes it, and nothing else ever recorded a model — so a
result could not be attributed to a model after the fact. On an evaluation platform that is fatal
for a leaderboard.

## The telemetry already knew

This is a **capture** fix, not new plumbing. Every harness adapter already normalises its own
spelling onto an event's `data["model"]`; the per-run fold just never collected it. Verified
against the real OTLP fixtures before writing a line:

| harness | model recovered | carried on |
|---|---|---|
| `claude_code` | `claude-fable-5` | usage metric (`gen_ai.request.model`) |
| `openhands` | `bedrock/au.anthropic.claude-sonnet-4-5-20250929-v1:0` | usage metric |
| `codex` | `gpt-5.5` | **`status` event** ("session started") |

No adapter changes, no new telemetry.

## The bug my synthetic tests missed

My first fold read the model only from **metric** events. All three synthetic tests passed — and
codex would have folded to nothing, because it reports its model once on a `status` event rather
than per call. Arguably codex is the more correct of the two: the model is a property of the
session, not of an individual call.

The fold now reads the model off any event that names one. **The real-capture test is what caught
this**, which is why it is parametrized over the three actual fixtures rather than over events the
test built for itself.

## The decision worth reviewing

**Disclosed and observed are kept as separate facts, not collapsed into one field.**
`conditions.model` is what an operator typed; `models_observed` is what the harness reported
running. When they **disagree**, both are shown:

```
model: claude-opus-4 (disclosed); telemetry reported gpt-5.5
```

Silently preferring either would misattribute the result, and the disagreement is the interesting
part. When only the harness named one it reads `gpt-5.5 (reported by the harness)`; when neither
did it still reads `not disclosed` — an honest unknown, never a guessed name.

`RunStats.models` keeps first-seen order rather than sorting, so a run that switched model mid-way
(a router, a fallback) reads primary-first.

## Surfaces

`report.md`, `run status`, and the `/result` envelope. Carried on the envelope rather than fetched
from `/stats` so `run status` stays a single request — `/stats` can also 409 on a non-terminal run,
which `run status` already handles differently.

`frontend/openapi.json` and `src/lib/api/schema.ts` regenerated, since `dump_openapi.py` documents
a `git diff --exit-code` no-drift gate.

## Tests

Ten, written first and watched fail — three fold tests, three real-harness captures, three CLI
renders, three report renders (one of which, the "honest unknown" case, passed before the change
and is there as a regression guard).

`ruff`, `ruff format --check`, `mypy` (496 files), `lint-imports` (4 contracts kept) all clean.
Suite: **2494 passed, 2 skipped** across unit + adapters + topology + integration.


christo-xorciseai:
**Existing stats snapshots do not pick up the new model field.**

The fold adds `RunStats.models`, but the freshness stamp remains only `adapter_name@adapter_version`. Neither component changes here. A pre-change snapshot with the current stamp deserializes with `models=()` and `current_run_stats()` returns it without folding. Reproduced by loading an old snapshot with the matching projection and checking that `_full_view()` is never called. `/result` also reads `reporting.get_stats()` directly, so `run status` does not refresh it either.

Existing recorded runs can therefore continue saying “model not disclosed” even when their retained telemetry names the model. Please version the stats fold (or provide an equivalent backfill) and use the freshness-aware path for `/result` as well. An upgrade test should start from stored JSON lacking `models`, with a current adapter stamp and model-bearing telemetry, then check both the report and result envelope.

**Public API naming:** `models_observed` is populated from the agent/harness event projection, which `otel/run_stats.py` explicitly defines as self-reported display data, “never an observed fact.” Please use `models_reported` or `harness_reported_models` and update the schema before consumers depend on the new name. The existing rendered “reported by the harness” wording correctly preserves this distinction.

Reviewed at [df41c15](https://github.com/xorcise-ai/xorcise/blob/df41c15946ba9333eb5d4f86758383ad3610ed7b/src/xorcise/core/rest/report_assembly.py).

guru-xorciseai:
Both fixed.

**Stale snapshots.** You were right that the stamp could not see this: `<adapter>@<version>` does not change when a field is ADDED to the fold, so a pre-change snapshot matched and was served as-is — the run kept saying "model not disclosed" while its telemetry named the model, which is the exact complaint this PR exists for. The stamp now carries the fold's own version (`+stats.2`), independent of the adapter, so a field addition invalidates older snapshots the way a classifier change does. It has already earned its keep: #134 adds `last_event_end_ts` to the same fold and bumps it to `stats.3`.

**`/result` reading around the freshness check.** Also correct — it used `reporting.get_stats()` while `/stats` and the report used `current_run_stats()`, so `run status` could disagree with the report about the same run. It now takes the same path.

**Naming.** Renamed to `models_reported`, and I think you have the better of this: `otel/run_stats.py` calls the projection self-reported display data, "never an observed fact", and the rendered wording already said "reported by the harness". The API name was the one place claiming more than the data supports — which is awkward, because the PR argues at length for keeping disclosed and observed distinct and then blurred it in the field name. Renamed before anything consumes it; `openapi.json` and `schema.ts` regenerated.

An existing stamp test asserted the old two-part format; updated with the reason recorded in it.

guru-xorciseai:

| PR | was | now |
|---|---|---|

christo-xorciseai:
*Edited: this originally overlapped with the comment above; trimmed to what is still open at 0474f07.*

Fold version, the `/result` path and the rename are in. Still open:

**1. The fold reads `model` off any event of any kind.** `run_stats.py:100` takes `data["model"]` from every event, and the generic adapter puts every span attribute into `data` (`otel/adapters/generic.py:139`). A `tool_call` event with `{"tool.name": "generate_image", "model": "dall-e-3"}` plus a `gpt-5.5` metric folds to `('dall-e-3', 'gpt-5.5')`, rendered as "dall-e-3, gpt-5.5 (reported by the harness)". Through the generic adapter the claude_code fixture recovers its model only by accident of attribute copying, and openhands recovers nothing. Restrict to the kinds the adapters deliberately place it on (`metric`, `status`, `message`, `error`) and exclude tool kinds and `unclassified`; longer term a dedicated key.

**2. No cap on the list.** It is agent-controlled, persisted, and returned on every `/result`: 500 distinct names → a 500-tuple and a 500-name Conditions row; one 200 000-char name → stored verbatim, a 200 KB `stats_json`, echoed on every `/result`, which `run list` and `leaderboard` call once per terminal run. First-N with truncation and a "+N more" marker; the other `RunStats` fields are scalars, so this is the first unbounded one.

### Nits

- CLI and report still disagree in the declared-only case despite the docstring at `cli/commands/run.py:82-83` saying they mirror: the CLI returns `declared` bare (`run.py:96`), the report renders `f"{declared} (disclosed)"` (`render.py:215`).
- Row label changed from `Agent model (disclosed)` to `Agent model`; the docs site may reference the old label.
- Exact-string mismatch means `claude-fable-5` declared vs `claude-fable-5-1` reported, or a family name vs openhands' Bedrock ARN, renders as a disagreement. Acceptable per your decision, but "telemetry reported" reads as contradiction there; "telemetry named" is softer.
- Scope: the Results page `ConditionsCard` (`conditions-card.tsx:11`) still shows only `conditions.model` although `RunStats.models` already arrives via `/stats`, and neither leaderboard shows a model; #113 asks for the leaderboard. Follow-up issue material.

Verified on the original head and unaffected by the fix: the three real fixtures resolve exactly the models in the table (codex only via its `status` event); Markdown/HTML/Rich escaping of hostile model names; old `stats_json` without `models` deserialises to `()`; `run status` against a pre-field server works; the two one-line frontend test edits are required because openapi-typescript emits default-valued fields as required.

guru-xorciseai:
Both items and all four nits fixed at `7f80333`.

**1. The fold read `model` off any event kind.** Confirmed — `data=dict(span.attrs)` in the generic adapter means any span with a `model` attribute contributed. Your `generate_image` case reproduced exactly.

`_MODEL_KINDS` is now the restricted set, and I went looking for the writer of each before trusting your list. All four have one:

| kind | who puts a model there |
|---|---|
| `metric` | `otel/adapters/genai.py:194` — the gen_ai usage metric (claude-code, openhands) |
| `status` | `harness_adapters/codex/otel.py` — `codex.conversation_starts` |
| `message` | `harness_adapters/claude_code/otel.py:211` — `claude_code.assistant_response` |
| `error` | `harness_adapters/claude_code/otel.py:160` — `claude_code.api_refusal` |

Your set was exactly right. The comment now carries that table so a fifth kind cannot be added without naming the adapter line that writes it. Tool kinds and `unclassified` are covered by an explicit test.

**2. No cap.** Both dimensions were unbounded — count and name length — so both are bounded: `MODELS_MAX = 8` first-seen, `MODEL_NAME_MAX = 120` with an ellipsis so a clipped name cannot read as a whole one, and `models_truncated` carrying the overflow.

One subtlety worth stating: counting distinct dropped names exactly requires remembering what was dropped, which is the unbounded set again. The tracking is therefore itself bounded at 512 distinct names and `models_truncated` saturates past that. Exact where it matters, bounded everywhere.

**On not bumping the fold version:** `STATS_FOLD_VERSION` stays `stats.2`. That version is introduced by *this* PR and has never shipped, so this defines its shape rather than changing it — there is no persisted snapshot stamped `stats.2` anywhere to go stale. #134 above still holds `stats.3` for `last_event_end_ts`. Flagging it because it is exactly the reasoning your original finding was about, applied in the other direction.

**Nits**

- **CLI/report disagreement.** Fixed properly rather than patched: they now share one function, `reporting.render.agent_model_line`, called by both. `lint-imports` confirms `cli` → `reporting` is legal. A parity test pins them across six cases including the declared-only one that had drifted. Two copies of the rule is what let them diverge; one is the only structural fix.

  Side effect worth calling out: `run status` now prints `model: not disclosed` where it used to print `model: model not disclosed`. The label already says model, and the shared renderer also fills the report's "Agent model" row where the stutter would be wrong. Two assertions updated. Tell me if you want the old string back.

- **"telemetry reported" → "telemetry named"**, as you suggested, with the family-vs-exact reasoning recorded in the docstring.

- **Row label.** I checked the docs repo — there is no reference to "Agent model" anywhere in it, so the rename is safe. Nothing to change there.

- **GUI scope.** Filed as #146, covering both the `ConditionsCard` gap here and the same class of gap on #134, and referencing #143 as the aggregate-view version. Not folded into this PR: adding GUI tiles is a separate change with its own questions. The docstrings on both sides that claimed field-for-field parity with the report have been corrected, so nothing now asserts a parity that does not exist.

**Frontend:** the schema gained `models_reported_truncated` and `RunStats` gained `models_truncated`, and `openapi-typescript` marked both **required** exactly as you saw last time. Regenerated the spec, updated the two fixtures, ran typecheck (0 errors) and the suite (742 passed) under Node 22.

Gates: ruff, ruff format, mypy, lint-imports clean; 2559 passed / 0 failed.

christo-xorciseai:
Verified at `4e9f85d` (`7f80333`): both items and all four nits fixed. I re-derived the writer table from the adapters rather than from the comment and it matches; all 20 event kinds probed, exactly `metric`/`status`/`message`/`error` fold; the three real fixtures resolve identically before and after; the cap holds (8 names, 120 chars, count saturating at 504; the 200 000-char name now costs 532 bytes of `stats_json`); CLI and report are string-identical across eight cases; OpenAPI and TS regenerate with zero drift. Agree on all three decisions: `stats.2` unbumped (no tag has it, and #134's `stats.3` re-folds any dev snapshot anyway), `model: not disclosed`, and #146.

Follow-ups, none blocking:

- `run_stats.py:143` clips before dedupe, so two long names sharing their first 119 chars merge into one and `models_truncated` under-reports. Dedupe on the raw name into the (already bounded) tracking dict and clip at emission, or say "distinct after clipping" in the contract comment.
- `render.py:220-226`: when the declared name is among the observed ones the line is `declared (disclosed)` only, hiding the other observed names and the `(+N more)` marker. Pre-existing rule; the new argument inherits the gap.
- The generic adapter still copies the whole attribute bag, so a generic span classified `message`/`error` that carries a `model` attribute still folds. Arguably right for those kinds; noting it.
- The docs repo's `reference/api.mdx` sample for `GET /result` does not show the two new fields.

Issue #120, opened 2026-09-15
[Bug]: Unmatched-harness runs silently lose telemetry fidelity (generic fallback → tool_call; no OpenHands content capture)


guru-xorciseai:
### Summary
Runs from a harness that has no matching adapter (unknown `--kind`, or a harness whose span shape the adapter doesn't recognize) **silently lose telemetry fidelity** in two compounding ways, with no warning recorded.

**(1) Generic fallback collapses unmatched spans to `tool_call`.**
`generic.py:classify()` keys off substrings of the span *name*, first-match-wins, with a catch-all default:
```
src/xorcise/core/otel/adapters/generic.py:60   return AgentEventKind.tool_call   # default
```
A harness whose span names contain none of its keywords gets **every** span labeled `tool_call` — including tool *results* and system prompts. Observed with OpenHands' `agent.ActionEvent` / `agent.ObservationEvent` / `agent.SystemPromptEvent`: all → `tool_call`; only `agent.MessageEvent` matched ("message"). And `agent_event_runs.warnings_json` was `[]` — nothing flags the degradation.

**(2) No content-capture path for OpenHands.**
`claude_code/telemetry.py` injects the flags that make the harness emit tool content (`OTEL_LOG_TOOL_DETAILS`, `OTEL_LOG_TOOL_CONTENT`, `OTEL_LOG_USER_PROMPTS`, assistant responses). OpenHands has a collect adapter (`openhands/otel.py`) but **no `telemetry.py`**, so nothing enables content capture — and that adapter only recognizes OpenHands-SDK/Laminar spans (`_execute_action_event`, `TerminalAction`, `conversation.send_message`, content in `lmnr.span.input`), not the bare `agent.*Event` event-stream shape. Result: content-less spans (only `event.class`/`event.id`/`event.source`) — no command, args, output, or prompt text.

### Why it matters
The normalized/rendered event view (UI, `events.jsonl`) misrepresents the run — everything reads as `tool_call`, tool *results* and prompts vanish into that bucket, and there is no tool content — all with **no warning**. For an eval platform this silently corrupts the evidence a run is judged and analyzed from.

### Proof (redacted)
```
raw span attributes (all spans):  event.class=<X>  event.id=<…>  event.source=<…>   # no body/content
normalized events: ActionEvent→tool_call, ObservationEvent→tool_call,
                   SystemPromptEvent→tool_call, MessageEvent→message
agent_event_runs.warnings_json = []
```

### Where to look
[...]

PR #129, opened 2026-09-16
feat(otel): render unmatched spans as unclassified instead of tool_call

christo-xorciseai:
## What

A span whose name matched none of the generic adapter's keyword rules fell through to `tool_call`, and its body was fabricated from the first attribute found. A harness emitting marker-only spans (class + id + source, no payload) therefore rendered as a wall of confident tool calls, the report counted them as tool calls, and nothing warned. Seen in the wild on a custom OpenHands-SDK driver: 113 of 114 spans per run mislabelled (#120).

This PR adds an honest `unclassified` kind and renders it as what it is.

## Backend

- New closed-set kind `AgentEventKind.unclassified`: a real event no adapter rule could classify. Distinct from `unknown`, which the real adapters use for plumbing they deliberately suppress and which the UI hides by default.
- Generic adapter v2: unmatched names → `unclassified`; `pick_body()` returns only a known content key, never `k: v` of an arbitrary attribute. The version bump rebuilds every cached projection on first read.
- `normalize_run` emits an `unclassified_spans` warning with the count and the deduplicated span names (capped at ten).
- Excluded from terrain attribution (nothing to attribute) and given its own "Unclassified spans" row in the report telemetry table. The allowlisted tool-call count already excluded it.
- The stale "1:1 port of parse-trace.ts" framing is dropped; that file no longer exists.

## Frontend

- `KIND_META.unclassified` in its own `"unclassified"` group: visible by default (unlike debug), never given an attribution dot (unlike action), muted on its own colour token so the legend can name it.
- Card: raw span name as title, attributes as chips, a one-line note that the span name is unrecognised and no renderable content arrived, View raw as before.
- Replay header: "N of M spans unclassified — shown as-is" beside the existing "generic renderer" badge.
- `openapi.json` / `schema.ts` regenerated with sorted keys; the snapshot also picks up two fields that had drifted (`agent_ingress`/`allow_egress`, `pid`).

## Verification

- Unit lane 2219 passed; mypy, ruff, lint-imports clean (4 contracts kept).
- Frontend typecheck, vitest (737), and production build clean.
- The seven archived runs behind #120, projected offline: the solved run now yields 113 unclassified events, 1 message, 0 tool calls in the report, and one warning naming `agent.ActionEvent`, `agent.ObservationEvent`, `agent.SystemPromptEvent`.

Refs #120 (part 1). Part 2 of #120 follows in the stacked PR.


guru-xorciseai (approved):
Approving. I went at this from the angle most likely to break it — adding a member to `AgentEventKind`, which the codebase itself calls a "closed render set" — and it holds up.

**The enum change is safe because the codebase counts rather than enumerates.** I looked for hand-written lists of kinds that a new member would silently fall out of, and the two places that could have been are both total by construction: `HarnessCapabilityProfile._total_and_noted` *raises* unless a profile covers 100% of the enum (so codex/claude-code/openhands pick up `unclassified: unsupported` with no edit), and the frontend's `KIND_META: Record<AgentEventKind, KindMeta>` would fail to compile without the new entry. The group filters key off `!== "debug"` / `!== "action"` rather than positive enumeration, so `unclassified` falls through visible-but-not-attributable without needing a change. The only two hardcoded counts (18 → 19) are both bumped in the diff.

**The backward-compatibility story is the part I liked most, and it is easy to miss.** Bumping `GenericOtelAdapter.version` "1" → "2" is not cosmetic: `adapter_version` is part of the `agent_event_runs` staleness key, so every run previously cached under the old classifier is treated as stale and rebuilt from canonical RAW OTLP under the new rules. No migration, no stale `tool_call` labels left behind. The test comment (`# v2: unclassified default + no fabricated body`) says why, which is what makes it survivable.

**Targeting is right.** The reported failure is exclusively in `GenericOtelAdapter.classify()`'s catch-all — `OpenHandsAdapter` already routes unrecognised spans to `unknown`, so the fix correctly does not touch `openhands/otel.py`.

**Tests reproduce the real shape**, not a strawman: the parametrised case is literally `agent.ActionEvent` / `agent.ObservationEvent` / `agent.SystemPromptEvent` / `agent.MessageEvent`, and the marker-only batches carry the PoC's exact attribute shape (`event.class`/`event.id`/`event.source`, no payload). `test_pick_body_never_fabricates_a_body_from_an_arbitrary_attribute` is a genuine behavioural delta.

One thing to confirm rather than change: the diff also carries `EnvironmentSpec.agent_ingress` / `allow_egress` and `PlaneStatus`/`SystemInfo.pid` in `openapi.json` + `schema.ts` — 5 lines against 82 lines of real `unclassified` change. Those backend fields already exist on `main`, so this reads as stale-codegen catch-up rather than anything smuggled in. Worth a nod that it was intentional.

Note this PR is currently `behind` main (9 commits — today's merges), so it needs an update-branch before it can go in under the strict status-check policy.

christo-xorciseai:
Thanks for the review. Two follow-ups on your closing notes:

**The codegen catch-up was intentional.** Confirmed against `main`'s backend: `EnvironmentSpec.agent_ingress` / `allow_egress` (`contracts/connect.py`, `control.py`, `mission.py`), `PlaneStatus` and `SystemInfo.pid` (`contracts/config.py`) all already exist there. The `openapi.json` + `schema.ts` hunks are the generated artefacts catching up to the backend, not new surface smuggled in with `unclassified`.

**Branch updated.** GitHub refuses `update-branch` on a stacked PR ("Updating a stacked PR's branch via this endpoint is not supported"), so `origin/main` was merged locally and pushed as 27e81b9. Clean auto-merge (only `openapi.json` and `schema.ts` touched on both sides, no conflicts). Smoke on the merged tree: mypy clean (496 files), ruff check + format clean, the five #129 test modules 59/59, frontend `tsc --noEmit` clean and replay vitest 103/103. The stack above this PR is unaffected: each PR's three-dot diff against its base is unchanged.

PR #130, opened 2026-09-16
feat(otel): warn when no span carries content, and disclose it everywhere

### LINKED ISSUE #120: [Bug]: Unmatched-harness runs silently lose telemetry fidelity (generic fallback → tool_call; no OpenHands content capture)

christo-xorciseai:
## What

Stacked on #129. Part 2 of #120: when a run's spans carry no content, say so everywhere the run is read, instead of leaving "0 tokens / 0 model calls / longest tool call 0 ms" for the reader to decode.

The motivating runs had 114 marker-only spans each (class + id + source, no payload). The replay showed names, the judge got an empty transcript and graded from artifacts alone, and nothing flagged it.

## Backend

- `distill.span_has_content` / `log_has_content` expose the judge distiller's own keep/drop decision, so "content" means one thing in the replay header, the report and the judge transcript.
- `normalize_run` records `spans`, `logs`, `content_spans`, `content_logs` in the run-level counts and emits a `no_content` warning when nothing carries content. It stays silent on the real Claude Code, Codex and OpenHands captures, whose plumbing spans are content-less but whose runs are not. `NORMALIZER_VERSION` 2 → 3 rebuilds cached projections.
- `GET /runs/{id}/telemetry` (`RunTelemetryView`): the events header without the events, read from one cache row via the new `SqliteAgentEventStore.read_header`.
- `GradeResult.transcript_items`: how many distilled items the judge was handed. 0 means it graded from artifacts alone; None on grades persisted before the field.
- Report: "Renderer" (with the generic-fallback marker), "Content-bearing spans", "Content-bearing log records", "Judge transcript items", and the warnings verbatim, in both Markdown and HTML.
- `run status`: a telemetry line plus every warning; `--json` carries the summary additively and older servers without the endpoint yield nothing.
- OpenHands launch tip: export the OTel variables before importing `openhands.sdk`. The SDK initialises its Laminar tracing once, at import, and that tracing is what carries prompts, commands and outputs. This was the second client-side defect behind #120.

## Frontend

- The replay header renders every normalization warning verbatim, except `unclassified_spans`, which already has its banner from #129.
- `openapi.json` / `schema.ts` regenerated for the new route, view and grade field.

## Follow-up: statistics re-folded under the current renderer

Review of the report table found its statistics rows and its telemetry warnings disagreeing on runs graded before this stack. The `RunStats` snapshot is folded once at terminate and frozen beside the grade, while the projection rendered next to it is version-invalidated and rebuilt from RAW — so an older run read "Tool calls 37 / Unclassified spans 0" directly above "37 span(s) matched no classification rule", and the Results page served the stale half.

- `RunStats.projection` records the adapter + projection version the snapshot was folded under (`otel.run_stats.projection_key`, e.g. `generic@2+normalizer.3`).
- `report_assembly.current_run_stats` — one reader for the report **and** `GET /runs/{id}/stats` — serves a snapshot whose stamp matches the run's current projection as-is; otherwise (older renderer, a pre-field snapshot, or none at all) it re-folds from the live projection and persists the refresh via `reporting.put_stats`, which touches the derived stats column only — never the grade, conditions or partial flags.
- The Results page's Activity summary shows an **Unclassified spans** tile when the fold counted any.
- `openapi.json` / `schema.ts` regenerated for the new field.

## Verification

- Unit lane 2233 passed; mypy, ruff, lint-imports clean.
- Frontend typecheck, vitest (738) and production build clean.

Closes #120.


guru-xorciseai (commented):
The backend half of this is good, and I checked the things most likely to be wrong with a warning like this. One real issue on the live surface, which I think is worth fixing before merge because it undercuts the PR's own title.

## The issue: the live replay can warn on a healthy run

`no_content` fires on whatever RAW has been ingested *so far*, with no notion of run completeness:

```python
content_spans = sum(1 for s in spans if span_has_content(s))
content_logs  = sum(1 for r in log_flats if log_has_content(r))
if (spans or log_flats) and content_spans + content_logs == 0:
```

and the live replay polls that every 1.5 s **while the run is still going**:

- `run-live.tsx:164` — `useRunEvents(runId, !!run.data && !terminal)` → active means NOT terminal
- `use-run-events.ts:40` — `refetchInterval: active ? 1500 : false`
- each poll re-runs `normalize_run` over the partial RAW
- `replay-timeline.tsx` renders `(meta?.warnings ?? []).filter(w => w.code !== "unclassified_spans").map(...)` — verbatim, with no active/terminal guard

So a healthy run whose first exported batch is lifecycle/marker spans — **including #120's own `agent.SystemPromptEvent` example** — shows "none of the N span(s) and M log record(s) carries a content-bearing attribute … the judge transcript for this run is empty" until the first content-bearing span lands.

Two things make that worse than a cosmetic flicker. The message states a fact about the **final** judge transcript while it is actually a prediction from incomplete data. And the CLI and report *do* gate correctly — `run_status` only fetches `/telemetry` when `"grade" in r`, and `assemble_report` returns `None` until a result exists — so for the run's entire early life the live view and the CLI disagree about the same run. That is the opposite of "disclose it everywhere", and a warning that cries wolf on healthy runs is one people learn to ignore.

I think the comment in `use-run-events.ts` — "the server recomputes them over the whole run on every page, so the latest page is the truth" — is where it slipped: that reasoning is sound for a *count*, but not for a presence/absence signal read mid-stream.

**Cheapest fix** is frontend-only: suppress the banner while `!terminal`, mirroring the CLI's existing "wait until graded" gate. Alternatively the server only includes `no_content` once the run is terminal, but that means threading run status into `normalize_run` for no extra benefit.

## What I verified as sound

- **One definition of "content".** `span_has_content`/`log_has_content` are literally `_render_span(...) is not None` / `_render_log(...) is not None` — the same keep/drop function the judge's `distill_transcript` uses. The replay, CLI, report and judge cannot drift apart on what "content" means, which is the failure mode I most expected here.
- **No staleness.** The cache key includes `source_max_seq`/`log_max_seq`, so when content arrives later the next read rebuilds and the warning clears by itself. It does not warn once and stick.
- **No grading leak.** `transcript_items` comes from `SealedContext` via the RAW-derived `distill.py` path, never from the AgentEvent projection, and `overall` never references it. The pre-existing `test_grader_never_imports_the_cache` boundary is untouched.
- **The regression test is real** — `_marker_batch(0, ["agent.ActionEvent", "agent.ObservationEvent"])` is the exact bare-marker shape from #120's proof, and it asserts the warning text and count rather than a constant.

Happy to approve once the live surface is gated; everything else here I would take as-is.

christo-xorciseai:
Fixed in c52719b, frontend-only as you suggested.

- `ReplayTimeline` takes a `live` prop and `run-live.tsx` passes `live={!terminal}`. While live, the adapter-warning list is withheld — `no_content` and any future run-level warning — because they state a fact about the **final** transcript and mid-run they are predictions from partial RAW. The unclassified banner stays, since it is a present-tense count. Once terminal the same `meta` renders the warning: deferred, not lost.
- The `use-run-events.ts` comment you pointed at now says the "latest page is the truth" reasoning holds for what has been *ingested so far*, and that a presence/absence signal is only final once the run is.
- New test `withholds normalization warnings while the run is live, but keeps the unclassified banner` renders with `live`, asserts no `adapter-warning` and a present banner, re-renders with `live={false}` and asserts the `no_content` sentence appears.

Validation: `tsc --noEmit` clean; vitest 289/289 across `features/replay` + `features/live-trace` (38 in `replay-timeline.test.tsx`). The live view now agrees with the CLI and the report for the run's whole life.

christo-xorciseai:
One more change on this branch, 15a30a3, from the deeper look at the report table your review prompted.

**The statistics rows and the telemetry warnings disagreed on older runs.** The `RunStats` snapshot is folded once at terminate and frozen beside the grade; the projection rendered next to it in the same table is version-invalidated and rebuilt from RAW. #129 bumped the generic adapter, so every run graded before it read "Tool calls 37 / Unclassified spans 0" directly above "37 span(s) matched no classification rule" — and `/runs/{id}/stats` served the frozen half to the Results page.

**Fix:** `RunStats.projection` stamps the adapter + projection version the snapshot was folded under. `report_assembly.current_run_stats` — now the single reader for the report and `GET /runs/{id}/stats` — serves a snapshot whose stamp matches the run's current projection as-is; otherwise (older renderer, a pre-field snapshot, or none at all) it re-folds from the live projection and persists the refresh via the new `reporting.put_stats`, which writes the derived stats column only — never the grade, conditions or partial flags. The fold therefore runs once per renderer change, not once per read. The Results page's Activity summary gains an **Unclassified spans** tile when the fold counted any. `openapi.json` / `schema.ts` regenerated for the field.

**Tests:** the contradiction pinned end to end — a graded run with marker-only spans has its snapshot overwritten with what the v1 classifier persisted (`tool_calls=3`, no `unclassified` key, `generic@1+normalizer.2`); `/stats` then answers `tool_calls=0`, `by_kind.unclassified=3`, the current stamp, and the refreshed snapshot is what `get_stats` returns next. A snapshot with the current stamp is served untouched (doctored counts survive the round trip). `assemble_report` yields an Unclassified count equal to the `unclassified_spans` warning's count. `put_stats` leaves grade / conditions / partial exactly as recorded and is a no-op without a result. Terminate stamps the projection. Frontend: the tile appears for `by_kind.unclassified > 0` and is absent otherwise.

**Validation:** unit lane 2239 passed; adapters + topology 206 passed; mypy clean; ruff check + `ruff format --check .` clean; import-linter 4/4; frontend `tsc` clean and vitest 741/741.

guru-xorciseai (approved):
Approving. Verified the gate rather than taking the description for it:

```tsx
{!live && (meta?.warnings ?? []).filter((w) => w.code !== "unclassified_spans")...
```

`run-live.tsx` passes `live={!terminal}`, so the run-level warnings are withheld exactly while they would be predictions from partial RAW, and the unclassified banner — a present-tense count — still shows. The live view and the CLI now agree for the run's whole life, which was the point. Keeping the same `meta` and deferring rather than dropping is the right shape; the warning appears the moment the run is terminal.

**The second commit is the more valuable one, and I had not spotted it.** The frozen `RunStats` snapshot disagreeing with the version-invalidated projection rendered beside it — "Tool calls 37 / Unclassified spans 0" directly above "37 span(s) matched no classification rule" — is a worse failure than the one I reported, because both halves look authoritative and a reader has no way to tell which is stale. Stamping the snapshot with the projection it was folded under, and re-folding when the stamp does not match, fixes the class rather than the instance.

I checked the write path, since refreshing a snapshot on a graded run is where I would expect a mistake: `put_stats` sets `row.stats_json` and nothing else — grade, disclosed conditions and partial flags are untouched, and it returns False rather than creating a row when no result exists. Making `current_run_stats` the single reader for both the report and `GET /runs/{id}/stats` is what stops the two surfaces drifting apart again.

Nice follow-through on a review comment into the thing the comment was a symptom of.

Issue #119, opened 2026-09-15
[Bug]: agent register --kind accepts unknown values silently and falls back to the generic adapter


guru-xorciseai:
### Summary
`xorcise agent register --kind <x>` accepts **any** string. An unrecognized value — e.g. `custom`, or a typo like `openhand` — is neither rejected nor warned about. At trace-ingest time it silently falls back to the `generic` adapter (`adapter_name="generic"`, `fallback=1`), which cannot classify the harness's spans. The user only discovers the mistake later, when the traces look wrong.

### Why it matters
A wrong or mistyped `--kind` silently degrades **every** run's telemetry fidelity (see #120). Observed in the wild: an OpenHands agent was registered `--kind custom` (`custom` is not a real adapter — the registered ones are `claude-code`, `codex`, `openhands`), so every run used the generic fallback and mislabeled its events, with no signal that anything was wrong.

### Proof (redacted)
```
$ xorcise agent register --name <agent> --kind custom     # accepted, no warning
# after a run:
agent_event_runs.adapter_name = "generic"
agent_event_runs.fallback     = 1
agent_event_runs.source_agent = "custom"
```

### Where to look
- `src/xorcise/core/cli/commands/agent.py` — `--kind` is a free-form `typer.Option(None, "--kind", …)`, no validation against known adapters.
- `src/xorcise/core/otel/adapters/registry.py` — selection falls back to `generic` for any unknown kind.
- Registered adapters live in `src/xorcise/core/harness_adapters/` (`claude_code`, `codex`, `openhands`).

### Related
- #120 — what the generic fallback then does wrong (silent fidelity loss).
- #121 — lossless raw OTel archive (safety net).

### Suggested fix
- Validate `--kind` at `register`/`edit` against the registered adapter names; on mismatch, error (or interactive-confirm) and suggest the closest known value.
- If a "no adapter / generic" mode is intentional, make it an **explicit** opt-in kind (e.g. `--kind generic`) rather than the silent destination for typos.
- Surface `fallback=1` prominently (e.g. in `run status` / `run list`) so a degraded run is visible even if registration slipped through.

---
_Reported from a hands-on eval session on a local dataset; run IDs, flags, keys, and trace/evidence contents are redacted — PoCs show only non-sensitive behavior plus verified source references._

christo-xorciseai:
Investigated against `main` (3c14a88). Summary: the behaviour is confirmed, the proposed hard validation is not the right fix, and the visibility gap is.

[...]

PR #131, opened 2026-09-16
feat(cli): gate an unrecognised --kind behind a y/N and show the harness per run

christo-xorciseai:
## What

Stacked on #130. Part of #119 — this PR is the **CLI half**. The REST + GUI path (`POST/PUT /agents`, and the registration page's free-text Custom harness ID) still accepts an unrecognised kind silently; that is tracked in #141 so #119 stays open until the primary control surface is covered too.

`agent register --kind <x>` accepted any string. A value no harness answers to (`Custom`, or a typo like `openhand`) routed every later run to the generic renderer with no signal at registration time. The operator found out weeks later when the traces "looked wrong".

## Changes

- **A warning plus a y/N gate on an unrecognised `--kind`**, on `agent register` and `agent update`. It names the registered harnesses (from `GET /harnesses`), suggests the closest match via difflib, and defaults to No.
  - `--yes` / `-y` skips the prompt but keeps the warning on stderr.
  - No TTY and no `--yes` **fails closed**: exit 2 with an example command including `--yes`. The gate exists to stop a script from making the choice unnoticed, so it takes the `down --purge` posture rather than the `confirm_or_abort` one, which skips the prompt when stdin is not a TTY.
  - A blank kind is the sanctioned custom path and never prompts. `update` gates only a changed kind. An older server without `/harnesses` cannot be judged, so the gate stays out of the way instead of blocking on a guess.
- **`confirm_gate`** in `cli/_ux.py`: the fail-closed sibling of `confirm_or_abort`, using the same `_stdin_is_interactive` seam tests already patch.
- **`fallback` means one thing** across the adapter, telemetry and launch registries: the generic floor did the work. Before, a blank kind resolved to an exact match on `"generic"` and read `fallback=false`, while `custom` read `fallback=true` for the very same rendering.
- **`run list` gains a Harness column** (the run's `source_agent` as a label; "Custom" means the generic renderer).

## Pushback carried from the issue

No hard validation against adapter names: free-form kinds are the intended custom-harness path (the label table already lists `langchain`, `gemini`, `goose`, `cai`, none with an adapter). No new `--kind generic`: it already works, and a blank kind means the same thing.

## Verification

- Unit lane 2243 passed; mypy, ruff clean.
- Gate matrix covered: interactive default No / explicit yes, non-TTY fail-closed, `--yes` bypass, near-miss suggestion, known and blank kinds never prompt, no harness list → no gate, `update` gates only a changed kind.


guru-xorciseai (commented):
The implementation is good — I went at the part that would have been a release blocker and it is handled properly. One scope question before this closes #119.

## Sound, and the thing I checked hardest

**The prompt cannot hang a non-interactive caller.** `confirm_gate` checks `_stdin_is_interactive()` *before* reaching `typer.confirm` and fails closed with `code=2` when there is no TTY and no `--yes`. That reuses the established seam — the one `confirm_or_abort` already uses, whose docstring says a prompt "would hang CI or an agent harness" — rather than reinventing it. `--yes`/`-y` is wired on both `register` and `update`, and `test_register_unrecognised_kind_fails_closed_without_a_tty` pins it. On a CI-and-harness-driven product this was the one thing that had to be right, and it is.

**The valid-kind list is derived, not hardcoded.** `_known_harness_kinds` fetches `GET /harnesses`, and server-side `list_harnesses()` builds from `launch_registry.registered_names()` ∩ `adapter_registry` — so it cannot drift as adapters are added. The "no /harnesses on the server" degrade case is tested too.

**Exit codes fit the contract**: 2 for the non-interactive usage error, 1 for an interactive decline (matching `confirm_or_abort`). No new precedent invented.

## The scope question

The gate lives only in `cli/commands/agent.py`. I checked every other way a `kind` can be set and none of them validate it:

- `rest/routers/agents.py` — `POST /agents` and `PUT /agents/{name}` pass `declaration.kind` straight through to `agents.register`/`update_agent`
- `contracts/agent.py:20` — `kind: str | None = None`, no validator
- `agents/registry.py` — persists it verbatim
- `frontend/.../harness-selector.tsx` — the "Custom" path is a free-text **Harness ID** `<Input>` posting to that same endpoint

This PR touches none of those files. So a harness self-registering over the API, or someone using the GUI's Custom field, still hits the exact bug #119 reports — with the CLI now looking fixed. Given the repo's own ADR that **REST is the primary agent control surface**, that is the larger share of the surface.

In fairness, the GUI is not silent: it says "Its export capabilities remain unknown until a verified adapter is added." But that is a disclosure, not validation — the typo case from the issue (`openhand` for `openhands`) still sails through as a plausible-looking custom harness, which is the specific thing #119 is about.

Two ways I would be happy with:

1. **Push the check down** into `agents.register`/`update_agent`, so REST and the GUI get the same warn-or-reject. The dependency rule allows it — domain → part-island is a legal import direction, so the registry lookup is available there.
2. **Re-scope this PR as CLI-only** and file a fast-follow for the REST/GUI path, so #119 does not read as fully fixed when the primary surface is still open.

Either is fine by me; option 1 is closer to what the issue asks for. No objection to the code as written.

christo-xorciseai:
Went with your option 2 — and the scoping turned up a reason option 1 needs its own design pass rather than a quick push-down.

**Why the check cannot live in the `agents` domain.** The adapter registry only holds `generic` until `harness_adapters.load_adapters()` runs, and `harness_adapters` sits *above* the application layer in `.importlinter`. So `agents.register` could import the registry but not populate it: in a fresh process where `POST /agents` is the first request, a domain-level check would flag every real kind as unrecognised. It has to sit at the REST boundary, the way `rest/routers/harnesses.py` lazily loads adapters + launch providers. And "warn" over REST needs a surfacing decision — a `warnings` field on the register/update response, a read-time `kind_recognised` on `AgentEntry` so the list/detail pages can show it, or a GUI-only "did you mean" under the Custom input. That is a contract choice, not a one-liner, so it should not ride along here.

**Done:**
- Filed #141 with those constraints, the three surfacing options and acceptance criteria.
- Re-scoped this PR: the description now reads "Part of #119 — the CLI half", pointing the REST + GUI path at #141, so #119 stays open until the primary control surface is covered.

No code change in this PR.

guru-xorciseai (approved):
Approving — and you were right to push back. I checked the layering claim before accepting it:

```
rest | frontend
harness_adapters
runs | agents | targets | reporting | orchestration | home | seams | runcontrol
```

`harness_adapters` sits **above** `agents`, so my option 1 was not a quick push-down — it was architecturally impossible. `agents.register` could import the adapter registry but never populate it, and in a fresh process where `POST /agents` is the first request every real kind would come back unrecognised. That is a worse bug than the one being fixed, and it would have looked correct in every test that imported the adapters first. Good catch; the REST boundary, where `harnesses.py` already lazily loads adapters and launch providers, is the right home.

The surfacing point is fair too: warn-over-REST needs a contract decision (a `warnings` field on the response, a read-time `kind_recognised` on `AgentEntry`, or a GUI "did you mean") and that does not belong riding along in this PR.

Re-scoping is done properly rather than cosmetically — the description reads "Part of #119 — the CLI half", #141 carries the constraints and the three options with acceptance criteria, and **#119 is still open**. That is the outcome I wanted: the tracker does not read as fixed while the primary control surface is still open.

No code change needed here. The CLI half was already right — `confirm_gate` checking `_stdin_is_interactive()` before it can ever reach `typer.confirm` is the part that mattered most on a CI-and-harness-driven product, and it reuses the existing seam instead of inventing one.

Issue #121, opened 2026-09-15
[Feature]: Persist raw OTel collector output under runs/<id>/ losslessly, independent of adapter matching


guru-xorciseai:
### Summary
Persist the **raw OTel collector output** (spans + logs, exactly as received from the agent) to `runs/<id>/` on every run — **before and independent of** any adapter matching, normalization, or run-id correlation filtering. A durable, lossless on-disk raw archive.

### Motivation
Today `runs/<id>/` holds only the **normalized** `agent-events.jsonl`; the raw lives in the `traces` DB table, and records that fail run-id correlation can be dropped at ingest (see the accepted/dropped accounting in `src/xorcise/core/otel/decode.py`). So evidence preservation currently depends on adapter coverage and correlation success. When a harness has no matching adapter or an unexpected span shape (see #120), information is degraded with no lossless fallback on disk.

A raw archive decouples **evidence preservation** from **adapter coverage**: whatever the collector received is always kept verbatim, so an unknown-kind / unmatched-harness run never loses information — it can be re-normalized or re-exported later once an adapter exists.

### Proposed behavior
- On each run, stream the raw OTLP to `runs/<id>/raw-otlp.jsonl` (otlpjson: one envelope per line, spans then logs) as the collector receives it.
- Include records that fail run-id correlation (tag them rather than dropping) so nothing is silently lost.
- Let `run traces --export`, re-normalization, and `regrade` read from this archive as the source of truth.

### Relationship to #120 (fidelity loss)
This is the safety net: it makes #120's harm **recoverable** — the raw is always intact even when the generic fallback mislabels or an adapter can't parse the spans. It does **not** by itself fix the *rendered/normalized* view (events still need correct classification/content from the adapter fixes), but it guarantees no permanent information loss and lets a later adapter recover the full signal.

### Where to look
- `src/xorcise/core/otel/decode.py` — ingest / accept-drop path (raw should be captured before this).
- run-directory writer that currently emits `agent-events.jsonl`.
- `run traces --export` — point it at the archive.

### Related
- #120 — silent fidelity loss this archive protects against.
- #119 — unknown `--kind` routing runs to the generic fallback.

Label: feature.

---
_Reported from a hands-on eval session on a local dataset; run IDs, flags, keys, and trace/evidence contents are redacted — PoCs show only non-sensitive behavior plus verified source references._

christo-xorciseai:
[...]

PR #132, opened 2026-09-16
feat(otel): record unroutable and post-seal OTLP drops instead of losing them silently

christo-xorciseai:
## What

Stacked on #131. Implements the narrowed scope proposed on #121: record what the receiver drops instead of losing it silently.

The receiver drops a batch in exactly two cases: it cannot be routed to a run (no `xorcise.run_id` resource attribute and no prompt sentinel in its content), or its run is already sealed. Until now the only trace of either was the `partialSuccess` count in the HTTP response, which exporters ignore, so an operator could not tell it had happened.

## Changes

- **`otel/ingest/drops.py`**: `DropCounters` (per signal and reason), `DropSpool` (a bounded directory of the dropped batches, newest kept, best-effort and never fatal to ingest) and `DropRecorder` (counters, a structured `WARNING` line, and the spool when configured).
- **`RouteResult.unrouted`** carries the RAW payload of every batch that failed correlation, so the receiver can record it rather than only count it. Additive field with an empty default.
- **The receiver records every drop** on both signals and serves the counters on its `/healthz` under `drops`. Zero across the board is the healthy reading; the CLI and system-view probes only check the status code, so they are unaffected.
- **Settings**: `otel_drop_spool_enabled` (default off), `otel_drop_spool_dir` (default `<home>/otel-dropped`, filled in `get_settings()` like the other home paths) and `otel_drop_spool_cap` (200). Both boot paths build the recorder from settings.

## Why the spool is global and off by default

A batch that failed correlation has no run id, so it cannot be filed under `runs/<id>/`. And the spool is a diagnostic, not a record: the RAW `traces` and `logs` tables remain the canonical evidence, as laid out on #121.

## Verification

- Unit lane 2251 passed; the receiver's adapters tests (JSON, gzip, protobuf, seal, logs, drops) pass; mypy, ruff, lint-imports clean.
- Covered: counters by signal and reason, the `WARNING` line content, spool envelope and eviction at the cap, a bad payload never raising, the settings factory, and the decoder carrying unroutable payloads on both signals.

Refs #121. Suggest retitling that issue to match, or closing it once this lands.


guru-xorciseai (commented):
Good change — turning a silent loss into a counted, inspectable one is the right instinct, and the seal invariant is respected. Two things I would like addressed, neither a blocker.

## 1. The spool bounds file count, not bytes

`_evict()` trims to `cap` **files**:

```python
files = sorted(p for p in self.root.glob("*.json") if p.is_file())
for stale in files[: max(0, len(files) - self.cap)]:
```

but nothing bounds the size of an individual payload, and I could not find a request-body limit anywhere in the chain — `ingest_traces`/`ingest_logs` do `raw_body = await request.body()` with no body-size middleware in `role_all.py`, `role_collector.py` or `serve.py`. Steady-state disk use is therefore `cap × max_payload`, with `max_payload` unbounded.

What makes it worth naming rather than shrugging at: the batches that reach the **unroutable** spool are by definition the ones carrying no valid `xorcise.run_id` — the ones that need no correlation to produce. So the content is attacker-influenced by construction.

Two things keep this off the blocker list, and I want to be fair about both: it is `otel_drop_spool_enabled: bool = False`, **off by default**, and anyone turning it on is deliberately diagnosing. But the config comment currently says the batch "is also written, **bounded**, under `otel_drop_spool_dir`" — and that word is doing more work than `_evict` supports. Either enforce a byte budget across the directory (or a per-payload ceiling, truncating or refusing above it), or soften the comment to say the bound is on file count.

## 2. The one path that must not regress is the one without a test

The post-seal + spool-enabled + real-HTTP-handler combination is not covered:

- `test_late_spans_for_a_sealed_run_are_counted_as_sealed` — real endpoint, but no `drops=` recorder, so `spool is None`
- `test_dropped_batches_are_spooled_when_a_spool_is_configured` — real spool through the endpoint, but only the **unroutable** reason
- `test_recorder_counts_logs_and_spools` — `reason="sealed"` with a spool, but calls `DropRecorder.record()` directly, bypassing `embedded.py`'s call site

So nothing proves that a spooled post-seal batch, arriving through `/v1/traces`, both lands in the spool *and* never reaches `trace_store`. By inspection it is correct — the `sealed` branch still `continue`s before `trace_store.append`, unchanged from pre-PR — but that is the one place a silent regression would be a seal-invariant violation, and it is the one place with no end-to-end test. A single test at the intersection of the three you already have would close it.

## Verified sound

- **Seal invariant holds.** The `sealed` branch `continue`s before any append; the spool writes to a separate directory that nothing in this diff — or anywhere in `src/`, per grep — reads back; `/healthz` exposes integer counters only, never payload content. A post-seal drop cannot re-enter evidence.
- **No injection surface.** The spool file is `json.dumps` of an already-parsed structure, never string-interpolated, and no CLI/report/frontend consumer is added. The one attacker-influenced value that reaches a log line is `run_id`, constrained by `_SENTINEL_RE` to `[0-9A-Za-z_-]+` — no newlines, so no log injection.
- **Classification is untouched.** `_run_id_of` / `_route_parsed` and the `telemetry_drain_seconds` grace window are unchanged, so no new risk of a legitimately-late record being misread as post-seal.

Happy to approve with the byte bound (or the comment reworded) and that one test.

christo-xorciseai:
Both addressed, in 220d6e2 (plus 1935ac8, ruff format only).

**1. Byte bound.** `DropSpool` is now bounded twice over: `cap` files **and** a new `max_bytes` total, from `otel_drop_spool_max_bytes` (default 64 MiB, `ge=1`). `_evict()` walks oldest-first until both bounds hold. A single batch larger than the whole budget is refused up front — logged as "exceeds the N-byte budget", counted, reported as not spooled — rather than written and then immediately evicting everything else. The config comment now says what "bounded" means and why: no body limit in the receiver, and unroutable ⇒ no valid run id ⇒ attacker-sized. Tests: eviction on bytes with the file cap far from binding; oversized refusal leaves the directory empty and the counters intact; the factory reads the new setting and falls back to the default on an older `Settings` object; the `get_settings()` default is asserted.

**2. The untested intersection.** `test_a_late_batch_for_a_sealed_run_is_spooled_and_still_never_persisted` — post-seal + spool enabled + the real `/v1/traces` handler. Asserts 200 with `rejectedSpans == 2`, `store.read("run-sealed") == []`, exactly one spool file with `reason == "sealed"`, `run_id == "run-sealed"` and both late span names, and `/healthz` reporting `sealed_spans == 2`, `sealed_batches == 1`, `spooled_batches == 1`. I added the logs twin too: the only sealed-logs coverage never asserted counters or a spool, so the `signal="logs", reason="sealed"` call site was unproven at the HTTP boundary.

Validation: the three touched test modules 53/53; full mypy clean (496 files); `ruff check` + `ruff format --check .` clean; import-linter 4/4 kept.

guru-xorciseai (approved):
Approving. Both verified in the code, not just in the description.

**Byte bound.** `_evict()` now walks oldest-first and breaks only when *both* hold:

```python
if remaining <= self.cap and total <= self.max_bytes:
    break
```

with `DEFAULT_SPOOL_MAX_BYTES = 64 MiB`. The detail I most wanted is there: a single batch larger than the whole budget is **refused up front** — logged, counted, reported as not spooled — rather than written and then immediately evicting everything else. Written-then-evicted would have satisfied the bound on paper while destroying the spool's entire contents on one hostile request, which is the failure mode a naive fix produces.

The config comment now earns the word "bounded" and says *why* the bound is needed — no body limit in the receiver, and unroutable means no valid run id means attacker-sized. That reasoning is the part that survives, long after the number changes.

**The intersection test.** `test_a_late_batch_for_a_sealed_run_is_spooled_and_still_never_persisted` covers exactly the combination that had none — post-seal, spool enabled, through the real `/v1/traces` handler — and asserts both halves of the invariant: the batch lands in the spool *and* `store.read("run-sealed") == []`. Asserting the negative is what makes it a seal-invariant test rather than a spool test.

Adding the logs twin was the right instinct and beyond what I asked for: the `signal="logs", reason="sealed"` call site was genuinely unproven at the HTTP boundary, and a fix that covered only traces would have left the quieter half of the same invariant untested.

PR #133, opened 2026-09-16
fix(cli): disclose a degraded judge wherever the score is shown

### LINKED ISSUE #110: [Bug]: Judge-degraded overall is shown without disclosure in run status / list / leaderboard
### Summary
When the LLM judge degrades (`judge_status != "ok"`), its half contributes `0.0` to `overall` **by design** — but that disclosure (`judge_status`) is only rendered in `report.md`. `run status`, `run list`, and the leaderboard show the degraded `overall` with **no indication the judge never ran**, so a full deterministic solve reads as `overall = 0.50`.

### Why it matters
`0.50` looks like "half solved" when it actually means "objective checks fully passed, judge unavailable." Anyone reading the CLI/leaderboard (not the full report) is silently misled. This is a disclosure-surface gap, not a math bug — the math is intentional and documented.

### Proof (redacted)
`detail_json` for such a result:
```json
{"breakdown": {"deterministic": 1.0, "judge": 0.0}, "overall": 0.5,
 "judge_status": "unavailable", "judge_coverage": 0.0}
```
`report.md` correctly says *"Judge half degraded: unavailable"*. But:
```
$ xorcise run status <run_id>
overall=0.50  deterministic=1.00  judge=0.00   # no judge_status shown
```

### Where to look
- `src/xorcise/core/eval/grade.py` — `overall = 0.5*det + 0.5*judge`; module docstring documents the degrade-to-0.0 behavior + `judge_status`.
- `run status` / `run list` / leaderboard renderers don't include `judge_status`.

### Suggested fix
Surface `judge_status` wherever the score is shown (a `degraded`/`judge:unavailable` marker), and/or present `overall` as incomplete when the judge didn't run — mirror what `report.md` already does.

---

guru-xorciseai:
Closes #110.

> **Stacked on #128** (`fix/record-observed-agent-model`). This PR's own change is the single
> commit `ec14f18`; review it against that base.
>
> **Merging note — #128 goes first, and deleting its branch will auto-close this PR.** GitHub
> closes any PR whose base branch is deleted, and a closed PR then refuses both reopen and a base
> change until the deleted branch is restored. Two ways round it: retarget this PR to `main`
> *before* merging #128, or merge #128 without deleting its branch and retarget afterwards.
> Either way this branch needs a rebase onto `main` after #128 lands as a squash, since it will
> then carry #128's commit twice. Happy to do that step myself when you get to it.

## The gap

A degraded judge contributes `0.0` to `overall` **by design**, so a run whose objective checks
passed completely scores a flat `0.50`. `report.md` always said why — *"Judge half degraded:
unavailable"* — but `run status`, `run list` and the leaderboard did not:

```
$ xorcise run status <run_id>
overall=0.50  deterministic=1.00  judge=0.00     # nothing says the judge never ran
```

So `0.50` reads as "half solved" rather than "checks fully passed, judge unavailable", and anyone
who does not open the full report is quietly misled.

## Disclosure only — no score changes

The issue is explicit that this is *"a disclosure-surface gap, not a math bug — the math is
intentional and documented"*, so **not one score moves here**. Excluding degraded runs from the
aggregates would be a much bigger decision than the one being asked for, and it belongs to whoever
owns the scoring model, not to this PR.

## What each surface now says

**`run status`** — the same sentence report.md prints, beside the scores:

```
overall=0.50  deterministic=1.00  judge=0.00
JUDGE DEGRADED — unavailable: the judge half did not run and contributed 0.00 to overall
(400: System message must be at the beginning.)
```

A healthy run stays silent — there is a test pinning that, because a banner on every run is noise
rather than disclosure.

**`run list`** — the score is marked `0.50*`, with a one-line legend printed only when a marked
score is actually on screen. `_run_score` already fetched the whole result to read `overall`, so
the judge's status was in hand: this costs no extra request.

**`leaderboard`** — a `No judge` column per agent, plus a note that those runs **are** included in
Avg/Best. That count is the only thing telling a reader an agent is ranked partly on runs no judge
ever saw.

## One predicate, three surfaces

`judge_degraded()` lives in `run.py` and all three import it, so they cannot drift into disagreeing
about what "degraded" means — which is how this kind of gap reopens.

`partial` is deliberately **not** degraded: there the judge ran and scored part of the rubric, and
it already carries its own coverage disclosure. A test pins that distinction so a later tidy-up
does not collapse the two.

## Tests

Six, written first and watched fail:

- `run status` discloses a degraded judge; and stays quiet when the judge ran (regression guard)
- `run list` marks a degraded score — asserted as `!= "0.50"`, so it fails on *shape* rather than
  on the particular marker I happened to choose
- `_flatten` carries `judge_degraded`; and does **not** flag `ok` or `partial`
- `summarize_by_agent` counts degraded runs **and** leaves `avg_overall` unchanged — that second
  assertion is the one that keeps this PR honest about not touching the math

`ruff`, `ruff format --check`, `mypy` (496 files), `lint-imports` (4 contracts kept) all clean.
Suite: **2501 passed, 4 skipped**.


christo-xorciseai:
**The “No judge” count and its footer describe different sets of runs.**

`summarize_by_agent()` excludes partial runs from `scored`, `avg_overall` and `best_overall`, but counts degraded judges across *all* `agent_rows`. Reproduced with one timed-out run carrying `partial=True`, `overall=0.5`, `judge_status="unavailable"`: the summary is `scored=0`, `avg_overall=None`, `best_overall=None`, `judge_degraded=1`. The footer nevertheless says those runs “ARE included in Avg/Best.”

Please either count degraded judges only among the scored/non-partial cohort, or distinguish total degraded runs from the subset included in aggregates and adjust the footer. Add a partial + degraded case; the existing normal degraded-run case does not expose the mismatch.

**Remaining surface gap:** the frontend aggregate models in `summarize-runs.ts` still discard judge status, so the GUI leaderboard/comparison averages cannot disclose degraded contributions. The GUI does already disclose failures in grading details, so this is specifically an aggregate-view gap. Please carry the disclosure through those views, or explicitly track that remaining scope rather than treating “wherever the score is shown” as complete.

Reviewed at [85ba01f](https://github.com/xorcise-ai/xorcise/blob/85ba01ff31f2a7c8a308cba578f1e3f383b673f7/src/xorcise/core/cli/commands/results.py).

guru-xorciseai:
Fixed, taking your first option.

Reproduced it exactly: `scored=0, avg=None, best=None, judge_degraded=1` — a column counting runs the footer says are in Avg/Best, for a run that reached neither.

`judge_degraded` is now counted over the same cohort as `scored`, so the column means what the footer describes: how much of the average in front of you came from runs no judge saw. A run that never entered the aggregate cannot have contributed to it. Tests cover both directions — the partial+degraded run that must not count, and the scored+degraded one that must.

**On the GUI aggregates** — you are right that "wherever the score is shown" was too broad a claim, and I have taken the second option you offered. `summarize-runs.ts` drops judge status before the aggregate views ever see it, so carrying the disclosure through means the summary model plus `performance-summary`, `completed-runs` and `cross-run-context`, with their tests — a separate change rather than a tail on this one. Filed as #143 with the shape and the cohort rule to mirror, and this PR's claim is narrowed to the CLI surfaces.

Noted your point that the per-run detail view already discloses grading failures; #143 is scoped to the aggregate views only.

guru-xorciseai:

| PR | was | now |
|---|---|---|

christo-xorciseai:
Reviewed at 85ba01f against #128. The predicate and the `run status` banner are sound. Requesting changes because the leaderboard's new disclosure line is itself false in a common configuration, and because the scope narrowing relative to #110 is not stated.

### Blocking

**1. "No judge" counts runs that Avg/Best exclude, so the footer is false.** `results.py:78` counts `judge_degraded` over all `agent_rows`; `scored` (line 65, which feeds Avg/Best) excludes `partial` rows; the footer at `:165-166` then says those runs "ARE included in Avg/Best". With 3 `done` + 1 `budget` run, all `model-not-configured` (i.e. any install without a judge key), the table renders `Runs 4  Scored 3  Avg 0.47 … No judge 4`. With `done/ok 0.9`, `timeout/unavailable 0.3`, `done/unavailable 0.5` it renders `Scored 2, Avg 0.70, No judge 2`, which reads as "both scored runs unjudged" when it is one of two. This is common because `run_terminate.py:186-194` records `unavailable` for any grading crash and `grade_judge` returns `model-not-configured` on every run when no key is set. Fix: count over the same set Avg/Best use (`not r["partial"] and r["overall"] is not None and r.get("judge_degraded")`) and pin it with a partial+degraded row (`judge_degraded == 1`, not 2). Keeping the wider count and rewording the note is the alternative, but then the column no longer answers "how much of Avg is unjudged", which is its stated purpose (comment at `:75-77`).

**2. The GUI surfaces #110 names are left out without saying so.** The web runs table (`results-all-runs.tsx:88`, carries `judgePartial` only) and the agent leaderboard (`summarize-runs.ts`, `performance-summary.tsx`) have no degraded flag and are untouched; the per-run GUI page (`grading-detail.tsx:44-80`) does disclose, so the gap is exactly the two surfaces this PR fixes for the CLI. `results.py:6` and the `test_cli_parity.py` docstring both declare the CLI leaderboard "mirrors the GUI aggregation"; this PR makes them diverge and puts the divergence-pinning tests in the parity file. Either state the omission and file a follow-up, or change "Closes" to "Part of".

### Should fix

- `unavailable` is also the status written when the whole grade crashed (`run_terminate.py:186-194`, `breakdown=ScoreBreakdown()` so deterministic is 0.0 too). The banner then says "the judge half did not run and contributed 0.00 to overall" beside `deterministic=0.00`, which under-discloses; only the parenthesised `grading failed: …` hints at it, and `run list` has no detail at all. Neutral wording ("scored without a judge") or a special case for `deterministic == 0.0 and detail.startswith("grading failed")`.
- `judge_detail` is printed in full inside the one-line banner (`run.py:152-158`): no truncation, newlines preserved. A three-line 400-char detail spilled across three terminal lines (Rich markup correctly escaped). Redaction exists only for HTTP response bodies in `judge_model.py:_redact`; `str(exc)` transport errors and the `grading failed: {exc}` fallback are unredacted. Not a new class of exposure (report.md and the GUI print it verbatim), but the pre-flight cap message is about five sentences long. Collapse whitespace and cap at ~160 chars with a pointer to `run report`.
- `run.py:296` `any_degraded = any_degraded or "*" in score` sniffs the rendered markup for the glyph instead of `_run_score` returning a flag. Works today; couples the legend to the marker.

### On the flagged decisions

- Disclosure-only: correct, and `avg_overall` is verifiably untouched.
- `partial` not degraded: defensible as a condition, but (a) #110 defines the gap as `judge_status != "ok"`, which includes `partial`, so this is a deliberate narrowing that should be said in those words; and (b) "it already carries its own coverage disclosure" is true for `run status`, report.md and the GUI only. In `run list` and the leaderboard a `partial` score renders as a bare `0.50` with no marker and no range (`_run_score` ignores `overall_upper`, so the list shows `0.50` where `run status` shows `0.50–0.85`). A stated follow-up rather than "already disclosed".
- One predicate: good. The full value set is the `Literal` at `contracts/grading.py:87` (`ok | partial | model-not-configured | unavailable`); missing/None → not degraded is right (only a non-grade envelope lacks it), unknown future value → degraded is conservative and matches `render.py:764`.

Nits: `results.py:19` imports the predicate from a sibling command module, which pulls in `run`'s Typer wiring; `_ux.py` is the shared-helper home. "No judge" is a raw count in a row of rates.

Verified: no extra request per `run list` row; the legend prints exactly once and only when a marked row is on screen (untested in the PR, worth a test); `_pin_column_widths` sizes `[warn]*[/warn]` correctly; 5 of 6 new tests go red on revert (the sixth is a negative guard).

Stack: the dependency on #128 is real, not just stacking convenience. The new tests use `_graded(...)` (`test_cli_run.py:816`), which #128 introduces, and the single commit does not apply onto main (`run.py:219`, `test_cli_run.py:1`). So "retarget to main before merging #128" is not friction-free; the clean order is merge #128 keeping its branch, retarget this PR, fix up, then delete #128's branch.

PR #134, opened 2026-09-16
fix(reporting): report what the agent did, not just how long the slot was held

### LINKED ISSUE #112: [Bug]: Crashed runs are closed out ~30 min late — misleading Duration and wasted wall-clock
### Summary
A run that crashes early is not closed out until roughly the full budget window later (~30 min in observed cases), so `report.md` reports a `Duration` of ~30m for a run whose agent died in under a second. Wall-clock and a run slot are wasted waiting.

### Proof (redacted)
For observed crashed runs, the `ConversationErrorEvent` fires ~0.7s after start (raw span timestamps), but `completed_at` lands ~1798–1803s later. The report then shows:
```
Duration | 30m 0s
```
even though there was ~1s of activity. The ~1800s gap is a reaper/close-out interval, not work.

### Why it matters
- Misleading `Duration` in reports/CLI (looks like the agent ran 30 min).
- Wasted wall-clock and a held run slot per crashed run — costly at batch scale (dozens of crashes in a single session observed).

### Where to look
- `src/xorcise/core/rest/run_terminate.py` (`seal_terminal`, drain/close-out path).
- Close-out appears time-driven rather than event-driven.

### Suggested fix
- Detect a terminal `ConversationErrorEvent(source=environment)` and close the run out promptly instead of waiting for the reaper.
- Compute/report `Duration` from the actual activity span (first→last event), not `created_at → completed_at`.

Possibly related: #66 (reap semantics).

---

guru-xorciseai:
Addresses the **misleading-Duration** half of #112. Deliberately does **not** close #112 — see
"What I did not do", which is the part that needs your call.

> **Stacked on #133** (`fix/disclose-judge-degraded`), which is itself stacked on #128. This PR's
> own change is the single commit `a3c4655`. It touches `render.py`, which #128 also touches, which
> is why it is stacked rather than cut from `main`.

## The bug

A run whose agent dies early is not closed out until its budget expires, so the report showed:

```
| Duration | 30m 0s |
```

…for about a second of actual work. Read alone, that says "the agent worked for 30 minutes".

## The change

Wall clock is **not** wrong and is not replaced — the run really did hold a slot and a subnet for
half an hour, and that is the cost. The telemetry span is what the agent *did*. Neither answers the
other's question, so Overview carries both:

```
| Duration       | 30m 0s |
| Agent activity | 1.0s   |
```

They agree on a healthy run and diverge loudly on a crashed one — which is exactly when someone is
reading the report to find out what went wrong.

The row is omitted when no telemetry bounds a span: a `0` there would claim the agent did nothing,
which is a different statement from nothing having been recorded. `first_event_ts` /
`last_event_ts` were already folded onto `RunStats`, so this needs no new capture, and both
renderers share `_metadata_rows` so report.md and the HTML report pick it up together.

## What I did not do, and why I want your call

The issue also asks to *"detect a terminal `ConversationErrorEvent(source=environment)` and close
the run out promptly instead of waiting for the reaper."* I stopped short of that, for two reasons
I would rather raise than decide alone.

**1. It would make a lifecycle decision from agent-self-reported data.** The architecture is
pointed the other way. `run_terminate` states that the event projection is *"agent-self-reported
display data — never an observed fact"*, `lint-imports` enforces "Grader never reads the AgentEvent
projection", and the readiness gate deliberately judges from the RUNNER's view of the container,
never from what the agent says about itself. Terminating a run because the agent reported its own
crash crosses exactly that line.

**2. The current behaviour is already documented as intended.** The UI detects this precise
condition and warns about it, and its own tooltip says:

> *"the agent may have crashed or disconnected; **the run keeps counting against its budget until
> it times out**"*

So this is a product behaviour change, not only a bug fix — and it would make that UI copy wrong.

**~~What I would build instead: an inactivity watchdog that closes out on *silence*.~~
Retracted — see my comment on #112 (2026-09-16).** Spans export on span END, so a long-running
step emits nothing while it is working: a silence watchdog would reap exactly the runs doing the
most work. Treat the #112 comment as current; this paragraph is kept struck through rather than
deleted so the review thread above it still reads.

The reasons NOT to terminate on a harness's own "session ended" event stand: it is
OpenHands-specific, a non-terminal error is indistinguishable from a terminal one in the
projection, and XORCISE does not launch the agent, so there is no process exit to fall back on.
One correction to how I argued it — the import-linter contract binds only `rest.grade_assembly`,
so "the architecture is pointed the other way" is a docstring convention here rather than a
mechanical rule; `run_terminate` itself already imports the projection to fold the snapshot.

#112 stays open. Closing a crashed run promptly is a real problem; a silence watchdog is not the
answer to it.

## Tests

Two, written first and watched fail. Asserted as whole rendered rows rather than bare digits — the
point is that the real work appears *beside* the wall clock, not how many decimal places the
formatter uses. The second pins the omit-when-empty case.

`ruff`, `ruff format --check`, `mypy` (496 files), `lint-imports` (4 contracts kept) clean.
Suite: **2502 passed, 2 skipped** across unit + adapters + topology + integration.


christo-xorciseai:
**Please label this as a self-reported telemetry window, rather than elapsed agent work.**

`_activity_seconds()` computes `last_event_ts - first_event_ts`. The fold takes the minimum/maximum event timestamps and ignores event duration; span-backed events commonly use the producer's span **start** timestamp. Reproduced with one event carrying `duration_ms=60000`: the stats retain that 60-second duration, but `_activity_seconds()` returns `0.0`, which renders as `Agent activity | 0.0s`. Multiple spans likewise omit the final span's duration, and a long silent gap between events is included.

The value can be useful, but it is not “how long the agent was actually emitting telemetry” or a measurement of active work. Please name it something like “Reported telemetry window” and document that it uses producer timestamps. If the intended measure is the full telemetry extent, include event end/duration where available and define the singleton/unknown-duration case. Add a single long-span case alongside the current first/last-event tests.

Reviewed at [759b884](https://github.com/xorcise-ai/xorcise/blob/759b884530c3f51774470b8777be4e8b862c2bf8/src/xorcise/core/reporting/render.py).

guru-xorciseai:
Fixed — and the row was wrong twice over, so both halves changed.

Reproduced first: one event with `duration_ms=60000` folded to first == last, and the report rendered `0.0s` for a minute of work. The duration was sitting in the stats the whole time; the window just never read it.

**What it measures.** The fold now records `last_event_end_ts` — the end of the last event that reported a duration — and the window runs first-start → last-end. So a single long span reads as its real extent, and a multi-span run keeps its final span's duration instead of stopping at that span's start. An event with no duration contributes only its timestamp: unknown extent is not zero extent. Fold version bumped (`stats.3`), which is precisely what that mechanism exists for.

**What it is called.** Renamed to **Telemetry window**, close to your suggestion. "Agent activity" claimed more than the number supports: it is built from producer timestamps and it includes any silent gap between events, so it is not a measure of active work. The docstring now says that plainly, including which clock it inherits.

It still earns its place beside Duration for the reason the PR was opened — the two diverge exactly when something went wrong, and Duration alone says a run reaped at its budget worked for thirty minutes.

Test added for the single-long-span case, plus the multi-span final-duration case and the unknown-duration one.

guru-xorciseai:

| PR | was | now |
|---|---|---|

christo-xorciseai:
*Edited: this originally overlapped with the comment above; trimmed to what is still open at e5d4a41.*

The rename, the end-aware window and the docstring are in. Still open:

**1. The PR body is now stale in a way that matters.** It still proposes the inactivity watchdog ("What I would build instead … close out on silence"), but your comment on #112 (2026-09-16) retracts it ("spans export on span END … would reap exactly the runs that are doing the most"). Treat the issue comment as current and strike or update that paragraph, so a reader of this PR does not take the retracted proposal as the plan. On the decision itself: accept the split and keep #112 open. Every quote checks out, with one correction: the import-linter contract binds only `rest.grade_assembly`, so for lifecycle "the architecture is pointed the other way" is a docstring convention rather than a mechanical rule (`run_terminate` itself already imports the projection to fold the snapshot). The stronger reasons not to terminate on the event are that it is OpenHands-specific, that a non-terminal error is indistinguishable from a terminal one in the projection, and that XORCISE does not launch the agent so there is no process-exit signal to fall back on.

**2. Parity docstrings are now wrong on both sides.** `render.py:198-201` says `_metadata_rows` is "matched field-for-field by the results page and the live run header", and `results-view.tsx:246-249` / `run-live.tsx:386-389` claim the same; neither renders the new row (nothing in `frontend/src` reads `last_event_end_ts` outside the generated schema). The Results page already fetches the snapshot via `useRunStats`, so the tile is a few lines; otherwise amend both docstrings and file a follow-up.

**3. The row still reads as a platform fact.** The docstring now says it inherits the producer clock, but the row sits between `Started` and `Duration` (server clock) with no marker. A skewed harness clock renders `Duration 1m 0s / Telemetry window 1h 0m 0s` with no signal, and garbage producer timestamps (ms passed as ns → 1970) render `495250h 0m 0s` in the headline table (pre-existing in the Telemetry section, new in Overview). #128 in this stack marks the analogous value "(reported by the harness)"; the same suffix here, plus omitting or dashing when the window exceeds elapsed by more than a tolerance, would close both.

### Nits

- `render.py:153` `max(0.0, …)` is unreachable on a fold-produced snapshot and on a corrupted one renders `0.0s` instead of omitting; return `None` when `last < first`.
- `render.py:222-223` evaluates `_telemetry_window_seconds(ctx)` twice.
- `render.py:201` still documents Duration as "how long it ran", the misreading this PR corrects.
- `test_reporting_render.py:587` `capsys=None` is a dead parameter. `test_the_report_omits_the_activity_span_when_there_is_no_telemetry` (`:606`) passed with `src` reverted on the original head because it asserts the absence of a label that never existed pre-PR; worth checking it now names `Telemetry window`, or it is vacuous.

Stack: no textual overlap with #128 below or #139 above.

guru-xorciseai:
All three items and the nits fixed at `9a8b963`.

**1. The stale PR body.** Struck through and replaced, with your correction folded in. It now says the watchdog is retracted, why (spans export on span END, so a silence watchdog reaps exactly the runs doing the most work), and points at the #112 comment as current. I kept the paragraph struck rather than deleted so your review above it still reads.

Your correction to my argument is right and I have recorded it: the import-linter contract binds only `rest.grade_assembly`, so "the architecture is pointed the other way" is a docstring convention here, not a mechanical rule — `run_terminate` already imports the projection to fold the snapshot. The reasons that do hold are the ones you listed: OpenHands-specific, a non-terminal error indistinguishable from a terminal one in the projection, and no process exit to fall back on because XORCISE does not launch the agent. #112 stays open.

**2. Parity docstrings.** You are right on all three. Nothing in `frontend/src` reads `last_event_end_ts`, and `render.py`, `results-view.tsx` and `run-live.tsx` all claimed a parity that does not exist. All three now say what is actually shared and name the row that is report-only.

I took the second of your two options — amend and file — rather than adding the tile. The reasoning: this is a report-rendering fix, and the tile has a design question the report does not, namely what a **live** run shows. The plausibility guard in item 3 needs an elapsed time to compare against, and a running run has no `completed_at`, so the live header would either show an unguarded window or show nothing while the Results page shows a guarded one. That is a decision worth making deliberately rather than inside this PR. Filed as #146 together with the #128 `ConditionsCard` gap, since they are the same class.

**3. The row read as a platform fact.** Both halves done. It carries `(reported by the harness)` — the same mark #128 uses for the analogous value — and it is now **withheld entirely** rather than shown when the producer's clock cannot be reconciled with the server's: a window longer than the wall clock the run actually occupied, beyond a 60-second grace for ordinary skew.

Your two examples are now tests: `Duration 1m 0s / Telemetry window 1h` renders no row at all, and a 1970 first-event timestamp renders no row rather than `495250h`. An absent row says "not known", which is the truth; a wrong number says something false with the platform's authority.

**Nits**

- `max(0.0, …)` gone — a window ending before it starts returns `None`, so a corrupted snapshot is unknown rather than `0.0s` ("the agent did nothing" is a different claim from "this cannot be trusted").
- The double evaluation is bound once.
- The Duration docstring no longer says "how long it ran" — it now says what the run **cost**, which is the reading this PR exists to correct.
- `capsys=None` removed.
- On the possibly-vacuous test: it does already name `Telemetry window`, so it is live — a bug that emitted the row unconditionally would fail it. I have added that reasoning to its docstring, since you were right that it was vacuous when written and it is not obvious from reading it why it stopped being.

Gates: ruff, ruff format, mypy, lint-imports clean; 2589 passed / 0 failed; frontend typecheck 0 errors and 742 tests passing.

christo-xorciseai:
Verified at `9a8b963`: all three items and the nits fixed. Sixteen-case probe through both renderers, Markdown and HTML agreeing on every row: shown up to elapsed + 60 s inclusive, withheld at +61 s, on a 1970 first event, and on a window that ends before it starts (all of which rendered a number before); the row carries "(reported by the harness)"; the fold-version chain `stats.2` → `stats.3` re-folds snapshots graded between #128 and this PR. Five new tests red on revert.

Before merge: the PR body is still stale beyond the struck paragraph. It says "single commit `a3c4655`" (now three), the example block still shows `| Agent activity | 1.0s |` for a row that is now `Telemetry window … (reported by the harness)`, and "Two tests" / "2502 passed" are old. PRs squash-merge with the body as the commit message, so that example would land on main describing a row that does not exist.

One correction to #146: its stated reason ("the plausibility check needs a completed run") is not right. `run-live.tsx:35-38` already computes elapsed from `Date.now()` for an unfinished run and ticks it, and now-minus-created is a valid upper bound for any plausible window, so the live header can apply the same `window > elapsed + 60` guard. Scope is the honest reason for deferring, and it is a good one; amend the issue so the next implementer does not inherit the false constraint.

Follow-ups, none blocking:

- `render.py:167-168`: when elapsed is unknown the comparison is skipped and the window renders unguarded. Unreachable through `/report` (409 on non-terminal), but nothing says so; either `if elapsed is None or …: return None`, or a comment naming the 409.
- `0.0s (reported by the harness)` still renders for a single no-duration event, any window under 50 ms, and consistently corrupt timestamps (all events ms-as-ns). Your own rationale for dropping `max(0.0, …)` applies. A symmetric lower bound (`first_event_ts < created_at − grace → None`) catches the corrupt case cheaply; the rest is a product call.
- The grace-constant comment (`render.py:130-133`) attributes the minute partly to "the lag between an event and close-out", which can only enlarge elapsed, never the window. NTP-class skew is the real justification.
- Withholding is not itself disclosed; the Telemetry section or the warnings list could say "telemetry window withheld — harness clock disagrees with server by N s".

PR #137, opened 2026-09-16
feat(rest,frontend): show XORCISE Remote announcements as banners in the local app

### LINKED ISSUE #136: [Feature]: Show XORCISE Remote announcements as banners in the local app
When XORCISE Remote has something to tell you — a maintenance window, an incident affecting
mission pulls, or the all-clear afterwards — there is currently no way for it to reach you inside
the app. This adds a banner that says so.

## What a user sees

A short, styled message with an icon and a text label for its kind: **Information**,
**Maintenance**, **Incident** or **Resolved**. Colour is never the only thing carrying the
meaning.

Two placements:

- **Application** — below the header, above the page content, on every route.
- **Mission catalog** — only inside the *XORCISE Remote* tab, between the filter controls and the
  *Connected* line. Never shown under *Your Own* or *Other providers*.

Most announcements can be dismissed with a close button, and the dismissal survives a refresh.
Incidents cannot be dismissed. If an announcement is edited after publication, it reappears even
for people who had dismissed the earlier wording.

## Fetching: deliberately no polling

Announcements are fetched **once per browser document load** and then left alone. No interval, no
refetch on window focus, on reconnect, or when navigating between routes. A new or withdrawn
announcement therefore reaches an open tab only when the reader refreshes the page.

This is a product decision, not an oversight. A banner that can appear under a reader mid-task is
disruptive, and polling every install of an open-source tool against a hosted service is a cost
nobody asked for. One request per page load serves both placements.

## Local endpoint

`GET /api/announcements` — the browser talks only to the local XORCISE backend, which fetches on
its behalf through the already-configured remote catalog source.

It returns an empty list, never an error, when:

- the remote catalog is disabled or not configured (and in that case nothing is sent at all);
- the remote service is unreachable, times out, or answers with something unexpected;
- the running deployment predates the feature.

Announcements are decoration. No failure of theirs may keep missions or the application from
loading, and the UI renders no loading skeleton and no error state — absence and failure both
leave the layout exactly as it was.

## Message formatting

Bodies are a deliberately restricted Markdown subset suited to a narrow banner: bold, emphasis,
inline code, short lists, and links. Raw HTML, images, tables and large headings are not accepted.
Links may be `https://` or site-relative only, and external links open with `rel="noopener
noreferrer"`. Rendering produces React nodes rather than HTML, so there is no injection surface to
sanitise.

## Accessibility

Incidents use `role="alert"`; every other kind uses `role="status"`, so routine announcements are
not repeatedly announced to screen-reader users mid-task. The close control is keyboard reachable
and labelled, and long bodies wrap rather than forcing horizontal scrolling.

guru-xorciseai:
Closes #136.

XORCISE Remote can publish short Markdown banners — a maintenance window, an incident affecting
mission pulls, the all-clear afterwards — and nothing carried them into the app. This adds the
local endpoint and the two banners that render them.

## What a user sees

A message with an icon and a text label for its kind (Information, Maintenance, Incident,
Resolved); colour never carries the meaning alone. An **application** banner sits below the header
on every route. A **catalog** banner sits inside the *XORCISE Remote* tab, between the filter
controls and the *Connected* line, and never appears under *Your Own* or *Other providers* — it is
the first child of that tab's panel, which returns `null` when inactive, so that is structural
rather than conditional.

Most banners can be dismissed and the dismissal survives a refresh. Incidents cannot. Editing a
published announcement brings it back for readers who dismissed the earlier wording.

## No polling, deliberately

One `GET /api/announcements` per browser document load, serving both placements. No interval, no
refetch on focus, reconnect, remount or client-side navigation. A new or withdrawn announcement
reaches an open tab only when the reader refreshes.

A banner that appears under someone mid-task is disruptive, and polling every install of an
open-source tool against a hosted service is a cost nobody asked for. The query pins every refetch
option explicitly so a library default cannot reintroduce polling, and a request-counter test is
the executable form of the requirement. That test earned its place: it caught that a *failing*
endpoint fetched twice, because `refetchOnMount` only governs a query that already has data and a
second observer mounts when the Remote tab opens.

## Failure isolation

Announcements are decoration. The endpoint returns an empty list, never an error, when the remote
catalog is disabled or unconfigured (contacting nothing at all), unreachable, slow, malformed, or
running a deployment that predates the feature and answers 404. Stub mode is included deliberately:
`xorcise up --stub` must stay deterministic. The UI renders no skeleton and no error state, so
absence and failure both leave the layout untouched.

The inbound parse is lenient rather than `extra="forbid"`: a strict remote contract has broken
every client of this project once already, when the server grew a field.

## Markdown

A restricted subset — bold, emphasis, inline code, short lists, links — rendered to React nodes,
never HTML. There is no `dangerouslySetInnerHTML` in this codebase and this change does not add
one, so there is nothing to sanitise. An anchor can only ever carry an `https://` or site-relative
URL; anything else renders as literal text. External links get `rel="noopener noreferrer"`.

The grammar is implemented three times across two repositories, so it is pinned against one shared
fixture whose SHA-256 is asserted in every suite that implements it.

## Accessibility

`role="alert"` for incidents only; everything else uses `role="status"`, so routine announcements
are not repeatedly announced to screen-reader users mid-task. The close control is labelled and
keyboard reachable, and long bodies wrap rather than forcing horizontal scrolling.

## Validation

`ruff`, `ruff format`, `mypy` (502 files), `lint-imports` (4 contracts kept), 2281 unit tests, 61
topology, 153 adapters, `tsc` clean, 840 frontend tests. The one failing unit test
(`test_serve_sockets.py::test_a_failed_lifespan_startup_is_reported_as_a_failure`) fails on `main`
before this branch and is unrelated.

The server side of the contract ships separately; until it does, this endpoint answers `404` from
the remote and returns an empty list, which is the designed behaviour.


christo-xorciseai:
**The row/body limits do not bound the remote HTTP response or total fetch time.**

`announcements()` performs a buffered `get()` and `resp.json()` before applying the eight-row limit or the per-announcement string limits. Reproduced with one valid announcement plus an ignored 50 MB JSON field: all 50,000,154 response bytes were consumed and the announcement was returned. Excess rows or oversized fields likewise consume memory before validation can reject them.

The `timeout=3.0` value is also an HTTPX per-operation/inactivity timeout, not an overall deadline. A local HTTP server sending three chunks 1.5 seconds apart was accepted after 4.6 seconds. A continuously trickling response can keep this request occupied much longer.

Please enforce a small response-byte cap while streaming, before JSON parsing, and an overall time budget if a bounded fetch is the intended guarantee. Keep the existing empty-result fallback for exceeded limits. These are resource/failure-isolation concerns; they do not imply a Markdown XSS bypass.

Reviewed at [0d678b4](https://github.com/xorcise-ai/xorcise/blob/0d678b4f3827566b57689e04c2c1f154562cbd85/src/xorcise/core/catalog/http.py).

christo-xorciseai:
*Edited: this originally overlapped with the comment above; trimmed to the items it did not cover, at 0d678b4.*

The byte cap and the deadline are covered above. Beyond those, the security property holds: over 35 hostile hrefs (`javascript:` in every case/encoding/NUL/`&colon;` variant, `data:`, `vbscript:`, `//evil`, `/\evil`, `\\evil`, `https:evil`, `%68ttps://`, quotes, angles, RLO, ZWSP, `mailto:`) the renderer produced no anchor; it emits only P/UL/LI/STRONG/EM/CODE/A/BR, refuses `![`, and falls back to the whole body as text on any parse failure. `openapi.json` and `schema.ts` are byte-identical to a fresh regeneration, `merge-tree` against main is conflict-free, and all lanes plus `tsc` and the 840 vitest tests are green. What remains:

**1. Quadratic regex before the length cap.** `markdown.tsx:119, 219, 445-448`: `PY_STRIP_END = /[WS]+$/` per line is O(n²) on a long whitespace run followed by a non-space, and `renderAnnouncementMarkdown` calls `normalizeBody` before the 600-code-point check in `rejectBody`. `"a" + " "*n + "b"`: n=10 000 → 117 ms; 100 000 → 14.5 s; 300 000 → 106 s (NBSP identical). Unreachable in production because `announcement_from_remote` rejects `len(body_md) > 600` (and the two sides count code points the same way: 600 emoji accepted, 601 refused), but the renderer is not self-defending while its header presents it as the safety boundary. Check the cap first, or replace the regex rstrip with a linear loop.

### Minor

- `https://` URLs with userinfo are accepted (`markdown.tsx:183` char class includes `@`): `[click here](https://xorcise.ai@evil.example/)` renders an anchor that navigates to `evil.example`. Refuse `@` before the first `/` after the scheme, and add it to the shared fixture.
- Empty `body_md` is accepted end to end (`contracts/announcements.py:108-110` checks type and upper bound only) and renders a banner showing only "INFORMATION" and a close button. `if not body_md.strip(): return None`.
- A hand-written localStorage entry hides an incident: `application-announcement.tsx:38` / `catalog-announcement.tsx:38` honour `isDismissed(id, rev)` regardless of `dismissible`. `dismissed && announcement.dismissible`.
- `tests/unit/test_announcements_rest.py:84-89` resolves `catalog.invalid` for real; on a hijacking resolver (a scenario `http.py:186` itself names) that becomes a 3 s+ connect. Point at `127.0.0.1:1` or use `MockTransport`.
- `contracts/announcements.py:60-123` `announcement_from_remote` is the first executable logic in the `contracts` leaf (no other `def` in `contracts/*.py`); import-linter passes, so this is a convention question, but `catalog/http.py` already owns the wire and the `_opt_str/_to_item` helpers.
- The fallback `<p>{normalized}</p>` (`markdown.tsx:450`) collapses newlines. Dismissing removes the focused close button with no focus target.

### Claims that did not hold

- "The grammar is pinned against one shared fixture whose SHA-256 is asserted in every suite that implements it": only `markdown.test.tsx` pins it in this repo (it does match the file bytes); the Python side of the grammar and the "rules 1-19" spec the comments cite live in the remote repo, so the other two implementations are not pinned on any of: case-variant schemes, `data:`, `vbscript:`, `file:`, `/\evil`, `\\evil`, `https:evil`, userinfo `@`, percent-encoded scheme, newline/`%0A` in URL, ZW characters, bidi isolates (only U+202E is pinned), nested brackets, `>`/`"` in link text, empty body end to end.
- "`test_serve_sockets.py::test_a_failed_lifespan_startup_is_reported_as_a_failure` fails on `main`": it passes here on the PR head, the file is identical to main, and none of the four main-only commits touch serve. Likely environment-specific.
- "No polling": enforced on the client (removing `retryOnMount: false` or `retry: false` makes the counter test fail, as claimed), but the server has no cache/TTL, so two local requests are two remote hits. "No polling" is per tab, not per install.

### For discussion

Nothing in this repo documents `GET /v1/announcements/active`, the six fields, the four tones, or "revision must bump on edit" beyond docstrings; "editing brings it back" is true only if the publisher bumps `revision`. Decide where the canonical contract lives before the server side ships. If install counts matter to the hosted service, a 60–300 s in-process cache in `list_announcements` is a few lines.

PR #139, opened 2026-09-16
feat(otel): make the evidence seal tamper-evident, not just a timestamp

### LINKED ISSUE #116: [Feature]: Run-evidence "sealing" has no integrity digest over the sealed evidence
### Summary
Run-evidence "sealing" records only a completion timestamp — it is a lifecycle marker, not a tamper-evident integrity seal over the recorded evidence.

### Details (confirmed in source)
```
src/xorcise/core/otel/store/models.py   TraceSealRow(run_id, sealed_at)   # timestamp only
src/xorcise/core/otel/store/seal.py     SqliteSealStore.seal()
src/xorcise/core/otel/ports.py          seal / is_sealed / sealed_at
src/xorcise/core/rest/run_terminate.py  seal_terminal(), _seal_telemetry()
```
`sealing` = "no more telemetry accepted after this time." The `content_hash` / `index_digest` fields are **mission-content** integrity (`mission_pull`), not evidence integrity.

### Why it matters
For "trust evidence, not claims," the sealed evidence a run is graded/regraded/exported from has no verifiable digest. Anyone with DB access could alter traces/results after sealing with no detectable change, and an export/regrade can't prove it graded the same bytes.

### Suggested fix (hardening)
- On seal, compute a digest over the run's sealed evidence (spans+logs+artifacts) and store it on `trace_seals`.
- Optionally sign it; verify the digest on `regrade` and `export`, and surface it in the report so consumers can validate.

Label: security. (Filed as hardening + clarification — please confirm intended guarantees of "sealing".)

---

guru-xorciseai:
Closes #116.

## What sealing was, and was not

`sealed_at` records when the evidence stopped growing. It says nothing about whether the bytes
still say what they said — so a span edited after sealing read back clean, and a regrade could not
show it had graded the same evidence as the first pass. The issue asks to confirm the intended
guarantee; this PR answers by giving it one.

The seal now records a SHA-256 over everything a grade is derived from — **raw spans, raw logs and
submitted artifacts** — and the report re-hashes at render time and states whether it still
matches.

## The honest limit, up front

**This detects modification. It does not prevent it, and it is not a signature.** Anyone who can
edit a span can recompute the digest beside it. The guarantee delivered is that evidence cannot be
altered *silently*. Binding the digest to a key so it cannot be re-forged is the natural next step
and is deliberately not attempted here — the issue lists signing as optional, and a half-done key
story is worse than none.

## Design decisions worth your eye

**The digest is assembled in the REST layer, not in the seal store.** It must span three owners —
traces and logs in the otel part-island, artifacts in runcontrol — and part-islands may not import
each other. So the one layer allowed to read across them does the join and hands the value down;
the store stays durable storage rather than acquiring a hashing policy. That is why `SealStore.seal`
grew a `digest` parameter instead of learning to hash.

**Length-prefixed fields, not delimiters.** With a plain separator, evidence containing that
separator could be re-partitioned into different fields that hash identically. Prefixing each field
with its byte length makes the framing unambiguous.

**`run_id` and a version tag are bound in.** The first stops a digest being lifted from one run and
presented for another; the second means a future change to what is covered reads as a *different
construction* rather than as tampering.

**`received_at` is excluded.** It is server-side metadata, not evidence, and hashing a value the
stores populate on read would make the digest depend on how it was fetched.

**Verification is tristate — and `None` is the important one.** No digest recorded (unsealed, or
sealed before this existed) is *unknown*, never *failed*. Collapsing it into False would report
every pre-existing run as tampered, which is the loudest possible false accusation and would teach
everyone to ignore the signal. The report says nothing at all in that case rather than printing a
permanent "unknown" row.

**A later seal never overwrites the first digest.** If it could, anyone able to re-seal could
launder edited evidence into a clean verification.

**Hashing is best-effort; sealing is not.** If the digest fails the run still seals, because
refusing to seal would leave telemetry admissible after terminal — a correctness regression traded
for a hardening feature. Such a run verifies `None`, never tampered.

## Tests

Eight, written first and watched fail, pinning the *property* rather than the hash construction —
which is free to change as long as it stays deterministic and covers what the judge reads:

- stable across recomputation (a digest that drifts would cry tampering on every honest run)
- **editing a sealed span changes the digest** — written as a direct `UPDATE` on `TraceRow`, exactly
  what someone with database access would do
- a late artifact changes it too (artifacts are graded evidence)
- two runs with identical evidence do not share a digest
- sealing records it and verifies clean; verification fails once evidence is edited
- a pre-#116 seal verifies `None`, not `False`
- sealing twice keeps the first digest

Plus three on the report surface and one on the migration column. Migration `0004` is nullable,
`batch_alter_table`, chain linear `0001 → 0004`, reversible.

**Disclosure on test order:** the eight core tests are strict TDD — all red before a line of
implementation. The three *report-rendering* tests were written after that small render change, so
I verified them by stashing the change and confirming all three fail without it, and pass with it.

`ruff`, `ruff format --check`, `mypy` (499 files), `lint-imports` (4 contracts kept) — each checked
individually. Suite: **2494 passed, 2 skipped** across unit + adapters + topology + integration.

## Not done here

`regrade` and `export` do not yet *refuse* on a mismatch — they surface it. Making a mismatch
fatal is a policy decision (does a tampered run fail loudly, or grade with a warning?) and belongs
to you rather than to the change that first makes the condition detectable.


christo-xorciseai:
**Three gaps in the evidence-seal guarantee:**

1. **Observed facts are graded but not hashed.** `sealed_context_for()` reads `SqliteObservedFactsStore` into `SealedContext.observed_facts`; `compute_evidence_digest()` only enumerates traces, logs and submissions. Reproduced by sealing a run, directly updating an existing `RunObservedFactRow.value`, and verifying it: `verify_evidence()` still returns `True`. Please cover this evidence stream too, ideally keeping the evidence enumeration aligned with grade assembly. Include a test that changes a fact consumed by a deterministic check.

2. **The hash is computed before admission is closed.** `seal_with_digest()` calls `compute_evidence_digest()` before `SqliteSealStore().seal()`. The OTLP receiver admits data while `is_sealed()` is false. Reproduced the interleaving by appending a trace after digest computation but before the seal write: the newly sealed run immediately verifies `False`, even though the write preceded sealing. Please close admission and account for in-flight writers before taking a consistent evidence snapshot, then persist its digest without reopening/replacing a completed seal. Simply swapping two calls would not settle already-admitted writes.

3. **The digest version is not recoverable from the stored value.** `_DIGEST_VERSION` is hashed, but storage contains only the hex digest, and verification always recomputes using the current implementation. Bumping the tag causes unchanged old evidence to compare unequal and be reported as modified; the verifier cannot distinguish a construction change from tampering. Store an explicit scheme/version and dispatch verification accordingly, or report unsupported versions as unknown.

The first two are current correctness failures. The third needs a durable format decision before version-1 seals are persisted.

Reviewed at [b77a7cf](https://github.com/xorcise-ai/xorcise/blob/b77a7cfbf737b230eae14b1b9cce364e3f1d11f4/src/xorcise/core/rest/evidence_seal.py).

guru-xorciseai:
All three fixed. This was the most useful review of the seven — the first two are exactly the failures that make a verification feature worse than none, because it reports confidently and is wrong.

**1. Observed facts.** Covered now, sorted by `(kind, name)` because the store's order is not a guarantee and the digest has to be reproducible. Your framing is the right one: `grade_assembly` feeds them into `SealedContext` and deterministic checks resolve against them, so evidence that decides a score has to be inside the seal. Test changes a fact and asserts verification flips to False.

**2. Hashing before admission closed.** `seal()` now closes the door first, then the evidence behind it is hashed, then `attach_digest()` records the result first-wins. The test spies on call order so the sequence itself is pinned, not just the outcome.

You were right that swapping two calls does not settle already-admitted writes, so I have not pretended otherwise: a request that already passed `is_sealed()` can still land while we hash. That is a receiver-level race needing a barrier. What changed is the bound — the window is now requests genuinely in flight at the moment of sealing, rather than however long hashing takes — and the code says so rather than implying the race is gone. Happy to do the barrier as its own change if you want it closed properly; it needs a decision about what the receiver does with a request that arrives mid-seal.

**3. Digest version.** Stored as `<scheme>:<hex>`, and an unrecognised scheme verifies as **None**, not False. Your point about which direction the failure runs is the one that convinced me: bumping the tag would have reported every untouched older run as MODIFIED, and a false accusation from the one signal that must never make one is worse than no signal.

`attach_digest` is deliberately separate from `seal` and never overwrites — whoever could replace a digest could launder edited evidence into a clean verification.

2591 tests, mypy 501 files, lint-imports clean.

guru-xorciseai:

| PR | was | now |
|---|---|---|

christo-xorciseai:
*Edited: this originally overlapped with the comment above; trimmed to what is still open at be0e185.*

Observed facts, seal-then-hash and the scheme tag are in, and the residual receiver-level race is stated honestly in the code. Still open:

**1. Verification runs unguarded and synchronously on every report request.** `report_assembly.py:243, 263-264` calls `verify_evidence` inside `assemble_report` for every `GET /runs/{id}/report`. Unlike every other display join in that module it is not wrapped, so a transient `database is locked` (which `run_terminate.py` documents as real on this shared file) 500s the report. Cost for 20k spans + 20k logs (66 MB): 1.8–2.1 s, of which about 95% is `SqliteTraceStore.read` materialising pydantic `TraceRecord`s (`evidence_seal.py:65`); raw SHA-256 of the same bytes is 0.09 s. #140 above fetches `/report` per exported run, so a bulk export inherits one re-hash per run against `get_text`'s 5 s default timeout. Hash from a column-only `select(TraceRow.seq, TraceRow.payload).order_by(...)` streamed into the hasher, and degrade to "could not verify" on failure.

**2. A hashing failure is indistinguishable from "predates the feature", forever.** `evidence_seal.py:108-114` catches `Exception`, logs one WARNING and returns with the run sealed and the digest NULL; `attach_digest` is first-wins so nothing backfills it, `verify_evidence` returns `None`, and the report omits the row entirely. An import error or a lock during every finalisation would silently make every new run unverifiable. Log at ERROR and store a sentinel so a post-migration run with no digest renders "seal digest unavailable" rather than nothing.

**3. Trace/log order is not canonical when `seq` collides.** `evidence_seal.py:65` sorts by `seq` alone; ingest assigns `seq = len(trace_store.read(run_id))` non-atomically (`otel/ingest/embedded.py`) and there is no unique constraint on `(run_id, seq)`, so concurrent OTLP POSTs can collide, and two rows with equal `seq` inserted A→B vs B→A hash differently. Stable in place today (rowid), but a dump/restore that reassigns ids flips verification. Tiebreak on `(seq, payload)` here, as you now do for facts on `(kind, name)`; unique per-run seq at ingest separately.

**4. The HTML report shows the Markdown markup literally.** `render.py:284-287` returns `` `abcd…` **MISMATCH — …** `` and `_html_kv_table` escapes every cell. In both formats the line is one row of the Conditions table at the bottom; a MISMATCH deserves a callout in Overview.

**5. "`regrade` and `export` … surface it" is false for regrade, and nothing machine-readable carries the digest.** `routers/runs.py` never calls `verify_evidence`; `/result`, `/runs/{id}`, the GUI and #140's bundle carry neither the digest nor the verdict, and the report prints 16 hex chars. #116 asked for the digest to be surfaced so consumers can validate. Put `evidence_digest` + `evidence_verified` on the result JSON and warn in `_grade_run` before regrading.

**6. Two docstrings are now inaccurate.** `evidence_seal.py:49-51` still says the digest "covers the three things a grade is derived from" (it is four). `:54-55` says `received_at` is "a value the stores populate on read"; `created_at` is a stored column (`otel/store/models.py:23`). Excluding server metadata is defensible, but #134 directly below derives the report's window from stored timestamps, so an edit there changes what the report displays while the seal still says "verified". "Graded evidence unchanged" is the accurate wording, and the report line could say the digest is stored beside the evidence, not independently attested, so "verified" is not over-trusted by readers who never see this PR body.

Nits: the "(not verified)" branch at `render.py:287` is unreachable from `assemble_report` (a `None` verdict suppresses the row); `hashlib._Hash` (`evidence_seal.py:33`) is a private name in a public signature; `db downgrade` destroys every digest and re-upgrade leaves them NULL (by design, but the migration docstring should say so). Tests worth adding: an out-of-hash edit does not trip; cross-process determinism; insertion order with duplicate `seq`.

Verified on the original head and unaffected by the fix: netstring framing has no re-partition collisions across fields, signals or submission fields; the digest is stable in-process, cross-process and across insertion order with unique `seq`; migration 0004 goes up/down/up cleanly; the first digest wins by the `run_id` primary key.

guru-xorciseai:
All six items and the nits fixed at `fee7258`. Two things came out of it that are worth your attention more than the fixes themselves.

**The digest printed on every report was a constant.** Chasing your "(not verified) is unreachable" nit found it: the stored value is `xorcise-evidence-v1:<64 hex>`, and `_evidence_seal_line` did `digest[:16]` — which is `"xorcise-evidence"` on **every run**, identical no matter what the evidence was. Every report has been showing the scheme tag and calling it a digest, and two readers comparing reports would have found them equal by eye. That is a regression I introduced in round one when I added the scheme tag and left the renderer assuming bare hex; the existing render tests passed a bare digest, which hid it. The assembler now hands over the bare hex.

**Your cost breakdown was right; my absolute numbers are lower.** On 20k spans + 20k logs (66 MB): `compute_evidence_digest` 0.678 s → 0.193 s, and 94% of the original was `SqliteTraceStore.read` materialising pydantic `TraceRecord`s — exactly your claim. My box is ~3× faster than the 1.8–2.1 s you measured; the shape is identical. New `ordered_payloads()` does a column-only `select(seq, payload)` streamed into the hasher.

**1. Unguarded and expensive.** Guarded now — a new `evidence_seal_view()` replaces two bare calls in `report_assembly` and `routers/runs`, so it degrades to "could not verify" instead of 500ing the report, matching every other display join in that module. It was also *two* seal-row reads. Pinned by a test that makes `store.read` raise, so the cost fix cannot silently regress without a timing assertion.

**2. A hashing failure was indistinguishable from "predates the feature".** Now logs at ERROR and stores an `xorcise-evidence-unavailable:<reason>` sentinel — shaped like a scheme tag, so `verify_evidence` reads it as "unknown construction" and never accuses. The report says "seal digest unavailable" rather than omitting the row.

**3. Non-canonical order on `seq` collision.** Both premises confirmed: ingest does `seq = len(store.read(run_id))` with no atomicity, and there is no `UniqueConstraint` on `(run_id, seq)`. Proved it concretely — two rows at `seq=1` re-inserted in the opposite order gave a different digest. Tiebreak is now `(seq, payload)`, matching what facts do on `(kind, name)`.

The ingest half is filed as #147, along with something I noticed while measuring: `len(store.read(run_id))` materialises every stored record on every ingest just to count them, making ingest O(n²) in record count. Same `store.read` pattern as the digest cost, worse exponent.

**4. HTML showed the markup.** Fixed, and a MISMATCH now gets a callout in Overview in both formats — Markdown above `## Scores`, HTML as a `banner alarm` above the Scorecard — with the shared string so the row and the callout cannot drift.

**5. Nothing machine-readable.** `RunResultView` gains `evidence_digest` and `evidence_verified` (a tristate: true, false, or null for "could not verify"). `_grade_run` now warns before regrading evidence that moved — disclosure only, never a gate, because refusing to grade would wedge the run at "grading" for ever. It fires only when the run arrived already sealed, which is the regrade case and the boot-sweep-after-restart case, so a first grade does not pay for a second hash.

**6. Docstrings.** Corrected to four things, with `created_at` described as server-side receipt metadata excluded on purpose, and the claim narrowed to "the **graded evidence** is unchanged" — with an explicit note that a report line derived from stored timestamps (#134 below) can move while this still verifies. The report line now says the digest is stored beside the evidence, not independently attested.

**Nits.** `hashlib._Hash` replaced with a local Protocol; the migration docstring now states that downgrade destroys every digest and re-upgrade leaves them NULL because `attach_digest` is first-wins. On "(not verified) is unreachable" — strictly an unknown-scheme row *is* reachable through `assemble_report`, but not on a single build version, which is a fair reading; it is genuinely reachable now via the item-2 sentinel and reworded to "recorded, but this build could not verify it — treat as unknown, not altered". Your three suggested tests are in; two of them passed against the pre-fix code, which is the right outcome — they pin properties that already held.

**`_DIGEST_VERSION` deliberately not bumped.** The construction did change (order, and the count's position), but v1 is introduced by this PR and has never shipped, so this defines it rather than changes it. If any of this has been released somewhere I have missed, that reasoning fails and it needs a bump — say so.

**Frontend:** both new fields are `X | None = None`, so they emit as `anyOf [type, null]` with no default and `openapi-typescript` marks them **optional** — no fixture needed changing, unlike #128's defaulted ints. Ran the full chain on Node 22: typecheck 0 errors, 742 tests passing.

Gates: ruff, ruff format, mypy, lint-imports clean; 2584 passed / 0 failed (2621 after merging #134 up).

Left for you to route: `GET /runs/{id}` still carries neither field, and the GUI and #140's bundle do not either — the finding asked for the result JSON specifically.

christo-xorciseai:
Verified at `20c7d4c` (`fee7258`): all six items fixed, and the constant-digest regression you found is confirmed (pre-fix every report printed `xorcise-evidence…`), fixed and pinned at two levels. Timing on 57.6 MB: 1.75 s → 0.46 s, within 0.06 s of the select-plus-hash floor; ordering is in SQL with BINARY collation and matches Python's sort across a probe spanning U+007F to U+1F600; the guard turns a `database is locked` from a 500 into a 200 with "could not verify"; OpenAPI and TS zero drift; 13 of 16 new tests red on revert (the other three are negative guards).

On `_DIGEST_VERSION`: agreed, no bump. Nobody has run the dev builds against real data, so no v1 seal under the earlier construction exists. Please record that in the PR body in one line ("v1 is the construction as of merge; no earlier v1 seals exist") so the decision is on the record.

Two things to decide before merge:

**1. Where verification-on-read lives.** `/result` now re-hashes the run's evidence on every call, and three first-party consumers fetch `/result` once per run in a loop: `run list` (`_run_score`), the leaderboard roll-up, and the frontend's per-run results query. At about 8 ms/MB, a 124-run dataset at 3 MB/run adds about 3 s to every `run list`. I would keep the re-hash on `/report` and regrade and have `/result` carry the digest only (one row read), or put it behind `?verify=1`.

**2. `/result` collapses "unavailable" to `null/null`.** The view's `unavailable` flag is dropped at `routers/runs.py:758-759`, so the machine-readable surface has exactly the ambiguity the report no longer has: a sealed-but-unhashed run is indistinguishable from a pre-feature one. A status field (`verified | mismatch | unverifiable | unavailable | none`) or a bare boolean, plus the regen.

Follow-ups, none blocking:

- "A first grade does not pay for a second hash" is true only for drain > 0. With `XORCISE_TELEMETRY_DRAIN_SECONDS=0` (the whole test suite and the documented operator mode) `seal_terminal` seals synchronously, so `_grade_run` sees its own seal as pre-existing, hashes twice (measured), and can log "evidence no longer matches" on a first grade if a straggler lands in the race window. A module-level "sealed here" set consumed in `_grade_run` fixes it and keeps the boot-sweep case right.
- The sentinel reason is `str(exc)` verbatim: unbounded and multi-line (stored only). First line, 120 chars, or the exception class name.
- A seal-row read failure returns an empty view, which suppresses the row: silence, not "could not verify". A fourth state would make the two failure modes distinguishable.
- The unknown-scheme row is silent (the view drops the digest), so "recorded, but this build could not verify it" is reached only by a read-time re-hash failure, not by the sentinel as the reply says.
- #140 can read `evidence_seal_view` directly and should put `evidence_digest` + `evidence_verified` in its bundle JSON. `run status` already has both fields in hand and prints neither.

PR #140, opened 2026-09-16
feat(cli): export a set of runs in one command

### LINKED ISSUE #114: [Feature]: Bulk export (report + traces + events) for a mission / filter in one command
### Summary
`run report`, `run traces`, and `run events export` are per-run only. Exporting a set of runs (e.g. every run of one mission) requires scripting a loop over run IDs.

### Why it matters
Analysis and hand-off almost always operate on a *group* of runs (a mission, an agent, a date range). Today that's manual glue for every user.

### Suggested fix
Add a batch export, e.g.:
```
xorcise run export --mission <m> [--agent <a>] [--since <ts>] --out <dir>
```
that writes `report.(md|html)` + `traces.otlp.jsonl` + `events.jsonl` per run into a directory tree. Bonus: a `--genuine-only` filter once run health/`terminal_trigger` is surfaced (see the crashed-run issue) so infra-failed runs can be excluded from an export in one flag.

### Where to look
- `src/xorcise/core/cli/commands/` (run command group) — compose over the existing single-run export paths.

Good first issue.

---

guru-xorciseai:
Closes #114.

## The gap

`run report`, `run traces --export` and `run events export` are per-run, so exporting a mission's
worth of runs meant scripting a loop over run ids — manual glue every user had to write for
themselves. Analysis and hand-off almost always operate on a *group*: a mission, an agent, a date
range.

```
xorcise run export --mission sqli-login --since 2026-07-01T00:00:00+00:00 --out ./export
```

Each selected run becomes `<out>/<run-id8>/` holding `report.md` (or `.html`),
`traces.otlp.jsonl` and `events.jsonl` — fetched from the **same endpoints** the three single-run
commands use, so this adds no second export format to keep in sync with them.

## Decisions worth a look

**Selection is a pure function.** `select_runs_for_export(runs, …)` holds every filtering rule, so
they are tested without a disk or a server; the command is I/O around it. That is also what makes
the boundary cases below cheap to pin.

**`--since` is inclusive.** A half-open boundary silently drops the run created exactly at a
timestamp pasted from a previous export — precisely the run someone re-running a range is most
likely to want. There is a test named for that case rather than a comment hoping someone notices.

**`--genuine-only` reuses the leaderboard's trigger set** (`{"done", "completed"}`) instead of
defining "a real run" a second time. Two subtly different definitions of that is how two surfaces
start disagreeing about the same data — and the issue explicitly wants this to line up with run
health. Without the flag nothing is hidden: the default export is everything that finished.

**Only terminal runs, ever.** An active run has no report, no sealed trace and no final event
stream — exporting it would write three misleading files instead of failing honestly.

**One bad run does not end the batch.** A run whose files cannot be fetched is reported and
skipped, because not doing that defeats the point of batching. But if *nothing* lands the command
exits 1 — a wholly failed export must not read as a quiet success.

## Tests

Six, written first and watched fail: four on the pure selector (terminal-only, mission + agent
filters, the inclusive `--since` boundary, `--genuine-only` including its does-nothing-by-default
half) and two on the command itself, writing into a `tmp_path` — one asserting all three artefacts
land per run, one asserting a failing run is skipped while the good one still lands.

`--agent` accepts a **name or an id**: run rows carry `agent_id`, but a name is what people have.

`ruff`, `ruff format --check`, `mypy` (496 files), `lint-imports` (4 contracts kept) — each checked
individually. Suite: **2488 passed, 2 skipped** across unit + adapters + topology + integration.
`xorcise run export --help` rendered and read through.

## Not done here

The issue's `--out` tree is flat per run (`<id8>/`). Grouping by mission or agent
(`<mission>/<id8>/`) would be easy to add but is a layout choice I would rather you make than
guess at — a tree shape is hard to change once people script against it.


christo-xorciseai:
**Bulk export has two reproducible correctness failures and an error-handling issue.**

1. **Naive `--since` values crash.** `datetime.fromisoformat("2026-07-01")` succeeds but produces a naive datetime; comparing it with a normal UTC `created_at` raises `TypeError: can't compare offset-naive and offset-aware datetimes`. The same occurs for a timestamp without an offset. Please define a timezone policy or reject these inputs with `BadParameter`, and cover it through the CLI parser.

2. **A still-grading response is exported as a successful report.** Terminal does not imply graded. `/runs/{id}/report` can return HTTP 202 with `{"status":"grading", ...}`; `get_text()` accepts it and the exporter writes it to `report.md`/`report.html`. Reproduced through `CliRunner` with the real `RestClient` response handling and a mocked HTTP 202: JSON was written to `report.md`, the other files were written, and exit status was 0. Please inspect status/content type and wait for or explicitly skip pending reports, as the single-run command does. Also document that this endpoint may start/retry missing grading through `ensure_graded_async`, potentially invoking the configured paid judge; export is not necessarily a passive download.

3. **The broad catch swallows the client's exit signal.** If the service goes down after `/runs` succeeds, `RestClient` prints its diagnostic and raises `typer.Exit(1)`. `except (OSError, Exception)` catches it, retries for every selected run, and reports `skipped <id>: 1`. Reproduced through the CLI. Please distinguish service-wide failures from run-specific failures and preserve a useful cause instead of converting the exit code into a skip reason.

Reviewed at [a469e77](https://github.com/xorcise-ai/xorcise/blob/a469e77741f954668ff347d77f28e97543201444/src/xorcise/core/cli/commands/run.py).

guru-xorciseai:
All three fixed, each reproduced first.

**1. Naive `--since`.** Confirmed the TypeError. `--since 2026-07-01` is the obvious thing to type, so it is now read as UTC rather than rejected — naive input on either side gets the same treatment, so the comparison is always aware-vs-aware. Covered through the selector for both a bare date and an offset-less timestamp.

**2. The 202 written as a report.** Confirmed — `get_text` hands back the grading envelope like any other body, and `{"status":"grading"}` landed in `report.md`. Those runs are now detected and held back.

One decision beyond what you asked: pending runs are tracked separately from failures, because the next action differs. An export where everything is still grading exits **3** — the in-progress code `run report` and `run status` already use — not 1. Telling a scripted caller the export broke, when it just needs retrying, is its own small lie.

On your documentation point: worth saying plainly that `/report` can start or retry grading through `ensure_graded_async`, so a bulk export is not necessarily a passive download and can invoke the configured judge. Since the exporter now skips pending runs rather than waiting, it does not drive that itself — but the endpoint's behaviour still belongs in the docs and I have not written it there yet.

**3. The swallowed exit.** `typer.Exit` is re-raised now, so a service-wide failure stops the export instead of becoming N identical `skipped: 1` lines with the real diagnostic lost.

Worth flagging: my first version of that test passed for the wrong reason — it exited 1 via the "nothing landed" path rather than by propagating. It now counts attempts and asserts the export stops at the FIRST run.

guru-xorciseai:

| PR | was | now |
|---|---|---|

christo-xorciseai:
*Edited: this originally overlapped with the comment above; trimmed to what is still open at 89a4b4e.*

Timezone, the grading envelope and the re-raised `typer.Exit` are in, and exit 3 for "everything still grading" is the right call. Still open:

**1. Zero matches still exits 3.** `run.py:636-640`. `_shared.py:61-64` documents `3 operation still in progress`, and you have just (correctly) used it for "all pending". A typo in `--mission`/`--agent` (`--agent alpha` vs registered `Alpha`; `--mission SQLi-Login` vs slug) now yields the same code as "come back shortly", so a script that treats 3 as retry loops on a typo. Exit 1 or 2 for a filter that matches nothing.

**2. `--agent`/`--mission` bypass the resolvers.** `run.py:627-630` is an exact case-sensitive name match with a silent fallback to "treat as id", and the selector (`:583`) matches only the slug/`mission_id`, not the display name `run create --mission` accepts and `run list` shows; the help does not say "slug". `resolve_agent_name` / `resolve_mission` (exact → case-insensitive → prefix → did-you-mean, fail loud) are already imported for `run create`; use them here too.

**3. Half-populated directories with no marker.** `target.mkdir` at `:659` runs after the report fetch but before traces and events; if either of those fails (a per-run exception → skipped, or a `typer.Exit` → abort), `report.md` is left on disk with no traces or events, and a consumer globbing `<out>/*/` sees something that looks like a run. Fetch all three, then write; or remove a directory this invocation created on failure.

**4. Prefix collision and path traversal.** `:649` `rid[:8]`: two runs sharing a prefix silently share a directory (second wins, `exported 2 run(s)`). Ids are `uuid4().hex`, so it is rare, but the failure is silent; fall back to the full id on a duplicate. A row with `run_id="../evil…"` writes outside `--out` (hostile/foreign server only); `re.fullmatch(r"[0-9a-f]{32}", rid)` is a one-line guard.

**5. `--genuine-only` is a copy, and its vocabulary is off.** `_GENUINE_TRIGGERS` (`:518-521`) duplicates `results._COMPLETED_TRIGGERS` with a "Mirrors" comment; two definitions is what the body says it avoids. The server writes `done`, `operator`, `timeout`, `deploy_failed` and `crashed`; `completed` is never written. The flag also drops `operator`-terminated and `crashed` runs, which the help (`:609`, "deploy failures, budget kills") does not say, and "budget kills" are recorded as `timeout`. Import the set (same layer) or lift it into `_ux` beside `run_state_label`, and fix the help.

**6. Cost with #139 as base.** `report_assembly.py` re-hashes the run's evidence on every `GET /report` (about 2 s for a 40k-row run), and `get_text` uses the 5 s default timeout. A bulk export is N serial re-hashes, and any run whose report or multi-MB `otlp.jsonl` takes over 5 s now aborts the whole export via the re-raised `typer.Exit`. Pass a larger per-call timeout as `run create` does (300 s).

Nits: an existing `<out>/<id8>/` is silently overwritten with stale extra files kept, and `--out` pointing at a file yields one `[Errno 20]` skip per run; an upfront check and a help sentence would be cheaper. `_is_grading_envelope` (`:524-531`) and the inline check in `run report` (`:722`) are the same test written twice; share the helper. The command tests still call `run_cmd.run_export(...)` directly, so exit codes and `--agent` resolution go untested.

Verified on the original head and unaffected by the fix: byte-equal `report.html`, `traces.otlp.jsonl` and `events.jsonl` against the three single-run commands; one `/runs` listing plus N×3 fetches (and `GET /runs` has no pagination cap). Stack: `src` applies cleanly onto main; only the test hunk depends on #128's context, and nothing from #139 is used, so this could sit on #128 rather than the whole chain.

guru-xorciseai:
All six items and the nits fixed at `42422f8`. Every finding reproduced first — eleven new tests failed for the right reason before any source changed.

**1. Zero matches exited 3.** Confirmed. Now exit 2 with "no finished runs matched the filters — nothing to export", pointing at `run list`. Exit 3 is left to the one genuinely retryable case, asserted separately so the two cannot collapse back together.

**2. Resolvers bypassed.** Worse than described: `--agent alpga` against a registered `Alpha` exited **3 with completely empty stderr**. A typo was indistinguishable from "come back shortly" *and* silent. Now uses `resolve_agent_name` and `resolve_mission` exactly as `run create` does, so a typo gets a did-you-mean and exit 1. Help says what each takes.

Behaviour change worth flagging: `--agent <id>` no longer works. It was an undocumented silent fallback, and keeping it would defeat the did-you-mean.

**3. Half-populated directories.** Confirmed — a failing `otlp.jsonl` left `report.md` alone on disk. All three documents are now fetched before `mkdir`, so nothing lands unless all three are in hand.

One honest limit: a *write* failure mid-directory (ENOSPC after the first file) can still leave a partial. You offered "fetch all three then write, **or** clean up on failure"; I took the first, which closes the reported failure completely but not that one.

**4. Prefix collision and traversal.** Traversal reproduced concretely — a row with `run_id="../outside/evil"` wrote outside `--out` and exited 0. Now `[0-9a-f]{32}` shape-checked, and a non-conforming id becomes a *skip* rather than a path, consistent with "one bad run must not end the batch". Colliding 8-char prefixes fall back to the full id.

**5. `--genuine-only` — you are right, and here is the evidence.** The column has exactly one write site, `runs/repository.py:334` in `mark_terminal`, reachable only from `reconcile.py:133` and `run_terminate.seal_terminal`. The complete set the server can write:

| trigger | written at |
|---|---|
| `done` | `routers/runcontrol.py:302` — the agent's own `/complete` |
| `operator` | `routers/runs.py:166` |
| `timeout` | `budget_watchdog.py:35`, `run_terminate.py:311` |
| `deploy_failed` | `run_readiness.py:50` |
| `crashed` | `reconcile.py:133` |

`completed`, `budget` and `error` are never written — they exist only in display maps. Help corrected to name what it actually drops; "budget kills" is gone, since those are recorded as `timeout`.

On sharing the set: importing from `results` was impossible — `results.py` already imports `judge_degraded` from `run.py`, so the reverse is circular. Lifted to `_ux.COMPLETED_TRIGGERS` beside `run_state_label`, which `lint-imports` accepts, and both import it. I **kept** `completed` in the set as a documented legacy synonym: `run_state_label` and the GUI both still accept it, and dropping it would silently change the leaderboard's `completed` column.

**6. Timeout.** 300 s per document, the value `run create` already posts with. Server-side re-hash is #139's and is fixed there — about 3.5× faster now.

**Nits.** `--out` pointing at a non-directory now fails upfront before a single fetch, with the overwrite semantics documented. `run report` calls the shared `_is_grading_envelope`. And the parser-level point: every export test now goes through `runner.invoke(app, [...])`, including the two round-one direct-call ones. The stub patches `RestClient.get`/`get_text` on the **class**, because the resolvers construct their own client — a module-level swap would have left `--agent` resolution talking to a real socket, which is exactly the kind of gap calling the function directly hides.

**Flagged, not changed:**
- `run traces --export` and `run report` still use the 5 s default for their own `get_text` calls. Same multi-MB argument applies, but they are pre-existing single-run commands outside this diff.
- `results._PARTIAL_TRIGGERS` still contains `budget`, which by the evidence above the server never writes. Not on your list, and changing it would alter leaderboard partial-rate semantics.

Gates: ruff, ruff format, mypy, lint-imports clean; 2589 passed / 0 failed (2642 after merging the stack up).

christo-xorciseai:
Verified at `b87686c` (`42422f8`): all six items and the nits fixed, reproduced pre-fix and gone on head, 14 tests through the parser, 10 red on revert. The run-id shape guard is safe: `uuid4().hex` is the only mint site, the create payload has no `run_id` field, and stub mode uses the same path. Agree on dropping `--agent <id>`, on exit 2 for zero matches, and on exit 3 only for all-pending.

One correction: no server has ever written `completed`. Every `terminal_trigger` write goes through `mark_terminal`, and the initial commit already had the same five values, so "legacy synonym" at `_ux.py:189` is unsupported by the history. Keeping it for parity with `run_state_label` and the GUI map is fine; say that instead.

Follow-ups, none blocking:

- The docstring's "one bad run never costs you the other ninety-nine" now holds only for non-HTTP errors: `RestClient._send` raises `typer.Exit(1)` for every error status (`rest_client.py:225-243`), including a per-run 404/500 from `/report`, `/otlp.jsonl` or `/events.jsonl`, and the export re-raises it, so one run whose report 500s aborts the whole batch. Pre-existing client behaviour, but the promise is this command's. A fetch variant that distinguishes per-request 4xx/5xx from transport failures would close it.
- `run.py:687` / `:699` gate on truthiness, so `--agent ""` skips resolution and exports everything; `is not None`.
- `resolve_agent_name` returns a name drawn from the same `/agents` list, so the `canonical` fallback at `:692-695` is dead and the second `/agents` GET is redundant.
- Partial success (some written, some skipped) exits 0 and is undocumented, so a script cannot detect a skipped run. Document it, or exit 1 when `skipped` is non-empty.
- Your flag on `results._PARTIAL_TRIGGERS` is understated: it contains `budget` (never written) and lacks `operator`, while the server's own partial flag is `recorded in ("timeout", "operator")` (`run_terminate.py:250`). The leaderboard's partial column disagrees with the server on operator kills. Worth its own issue.
- A previous export's `<id8>/` directory is orphaned when a later export hits a prefix collision (both runs then land under full ids); and the mid-write partial (ENOSPC on a multi-MB trace) is the realistic case the fix does not cover; temp dir + rename would take both.
- `run traces --export` and `run report` still fetch on the 5 s default; a shared document-fetch constant used by all three is the natural follow-up, since #139 made every report fetch slower.

